Issue #105, opened 2017-01-17
New Airport: Osaka Kansai International Airport


indianbhaji:
See #99

PR #116, opened 2017-01-19
feature/105

indianbhaji:
Kansai International Airport (関西国際空港 Kansai Kokusai Kūkō) (IATA: KIX, ICAO: RJBB) is an international airport located on an artificial island in the middle of Osaka Bay, 38 km (24 mi) southwest of Ōsaka Station, located within three municipalities, including Izumisano (north), Sennan (south), and Tajiri (central), in Osaka Prefecture, Japan. The airport is off the Honshu shore and was designed by Italian architect Renzo Piano. The airport serves as an international hub for All Nippon Airways, Japan Airlines, and Nippon Cargo Airlines, and also serves as a hub for Peach, the first international low-cost carrier in Japan.

It is colloquially known as Kankū (関空) in Japanese.

Kansai opened 4 September 1994 to relieve overcrowding at Osaka International Airport, which is closer to the city of Osaka and now handles only domestic flights. During the 2006 fiscal year, KIX had 116,475 aircraft movements, of which 73,860 were international (31 countries, 71 cities), and 42,615 were domestic (19 cities). The total number of passengers was 16,689,658 of which 11,229,444 were international, and 5,460,214 were domestic, sixth in Japan and second in Osaka area. However, in 2009, airport traffic has fallen by almost 20% in just two years to 13.4 million. In 2010 airport traffic had risen to over 14 million, with international passengers accounting for approximately 10.4 million and domestic passengers accounting for approximately 3.7 million. Freight volume was at 802,162 tonnes total, of which 757,414 t were international (18th in the world), and 44,748 t were domestic. The 4,000 m × 60 m (13,123 ft × 197 ft) second runway was opened on 2 August 2007. As of June 2014, Kansai Airport has become an Asian hub, with 780 weekly flights to Asia and Australasia (including freight 119), 59 weekly flights to Europe and the Middle East (freight 5), and 80 weekly flights to North America (freight 42).

Kansai International Airport was named the most Muslim friendly airport by Salam Standard alongside Kuala Lumpur International Airport (KLIA) and Singapore Changi Airport (SIN).

From [Wikipedia](https://en.wikipedia.org/wiki/Kansai_International_Airport)

Also known as zlsa/atc#696
Requested by @Exfil in zlsa/atc#683

Featured in #105 
Resolves openscope/openscope#105


assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+    "ctr_ceiling": 7000,
+    "initial_alt": 3000,
+    "position": ["N34d26m03.00", "N135d13m58.00"],
+    "rr_radius_nm": 5.0,
+    "rr_center": ["N34d26m03.00", "N135d13m58.00"],
+    "has_terrain": false,
+    "wind": {
+		"angle": 204,
+        "speed": 9
+    },
```
erikquinn: Be careful to not use "hard tabs" (the `\t` character). As shown in the github code preview thing here, `angle` and `speed are misaligned, as caused by there being two "hard tabs" before `angle`, while `speed` has eight spaces. The latter is preferred.

Please review the file and remove any tabs and replace with four spaces. I'm not sure how to do so in vscode, but sublime and atom both can search by regex, so if that's possible just find all `\t` and replace with four spaces.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+			["N34d39m14.00", "E136d05m25.00"],
+			["N34d16m09.00", "E136d09m43.00"],
+			["N33d35m04.00", "E134d59m50.00"]
+        ]
+    }],
+    "fixes": {
+		"AKASI": ["N34d35m00.30",	"E134d45m00.30"],
+		"ALISA": ["N34d30m42.60",	"E134d08m04.10"],
+		"ALLAN": ["N34d14m04.10",	"E134d57m28.30"],
+		"ASAMI": ["N34d07m57.90",	"E134d48m41.80"],
```
erikquinn: the fixes section is an example where hard tabs are present and should be converted.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+		"SALTY": ["N33d51m10.00",	"E132d55m31.00"],
+		"STORK": ["N32d56m48.00",	"E133d14m35.00"],
+		"SUC":   ["N32d45m21.00",	"E132d59m48.00"],
+		"TAPOP": ["N32d40m01.00",	"E136d07m57.00"]
+	},
+    "runways": [{
+        "name": ["06L", "24R"],
+        "name_offset": [
+            [0, 0], [0, 0]
+        ],
```
erikquinn: If no offset is being applied, should we omit this? It seems like it's often unecessary to offset them anyway, as the numbers are an aid, and not a necessity.

erikquinn: If no offset is being applied, should we omit this? It seems like it's often unnecessary to offset them anyway, as the numbers are an aid, and not a necessity.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+        "end": [
+            ["N34d25m42.86", "E135d12m22.33", "122.5ft"],
+            ["N34d27m04.58", "E135d14m24.08", "122.6ft"]
+        ],
+        "length": 4.000,
```
erikquinn: The `length` property has been deprecated, as it is calculated from the distance between the ends. Just `name`, `end`, and `ils` are the only properties that are needed.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+            "exitPoints": {
+                "MAIKO": [
+                    ["MAIKO", "A50+"]
+                ]
+            },
+            "rwy": {
+                "06L": ["KNE", ["MAIKO", "A50+"]],
+                "06R": ["KNE", ["KCE", "A25+"], ["MAIKO", "A50+"]],
+                "24L": [],
+                "24R": []
```
erikquinn: There will be some issues here:
 - from Runways 24L & 24R, this will be a "one-fix" procedure, which makes the game freak out.
 - the `MAIKO` fix is listed twice, and it should only be listed once. Else, the aircraft will have a waypoint list of `KNE, KCE, MAIKO, MAIKO` from `06R`.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+        }
+    },
+    "stars": {
+        "AKASI": {
+            "icao": "AKASI",
+            "name": "Akasi",
+            "entryPoints": {
+                "TOKUSHIMA": [
+					"TSC"
+				]
```
erikquinn: Due to the need to type the name of the entry point when assigning aircraft a new STAR, please use the proper ICAO fix/navaid identifier for all entry and exit points:
```
"entryPoints": {
    "TSC": [
        "TSC"
    ]
}
```

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+            },
+            "body": [
+                ["GATES", "A40"]
+            ],
+            "draw": [
+                ["TSC*", "GATES"]
+            ]
+        },
+        "DANDE_A": {
```
erikquinn: What's the deal with these variants? Provided that suffixes be completed, is that something that would cover this?

Regardless, due to the need for users to type the names of these procedures, please refrain from using special characters like underscores.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+            "icao": "TOMOH1",
+            "name": "Tomoh One",
+            "exitPoints": {
+                "KOCHI_S": [
+					"UMAJI", "KRE"
+				],
+                "KUSHIMOTO": [
+					["GOBOH", "A60+"], "KEC"
+				]
+            },
```
erikquinn: Due to the need to type the name of the exit point when assigning aircraft a new SID, please use the proper ICAO fix/navaid identifier for all entry and exit points:

```
"entryPoints": {
    "KRE": [
        "KRE"
    ],
    "KEC": [
        "KEC"
    ]
}
```

assets/airports/rjbb.json
```diff
@@ -0,0 +1,2732 @@
...
+            ["aal/long", 8],
+            ["aar", 1],
+            ["aua", 1],
+            ["ava/long", 1],
+            ["baw", 123],
+            ["bel", 1],
+            ["lzb", 1],
+            ["cpa", 1],
+            ["ces", 1],
+
```
erikquinn: Depending on the aircraft spawned in this pattern, many will not be able to keep up with this speed, which will result in compression, aircraft passing each other, etc. A few guidelines for speeds:

#### Known to be turbojet only:
 - if below 10k ft, set to `250`
 - if not important, set to `280`
 - never set higher than `320`

#### Known to contain turboprops
 - same as above, except never set higher than `280`

That should do a pretty good job at keeping those conflicts at bay.

erikquinn: Be sure to always leave a single blank line at the end of each file.

erikquinn: Are the `rate` values for the arrival spawnPatterns accurate? There are only 9.6 arrivals per hour, making it pretty slow. Is that expected?

indianbhaji:
I'll have a look at your points and fix anything where I can tomorrow - thanks!

The A - E variants are listed as separate arrivals in the AIP.

I'm unsure of how you want me to replace the underscores too, as spaces are prohibited (unless you want dashes?)

indianbhaji:
The frequency of aircraft is calculated through the total aircraft movements per year divided by hours in a year - this comes to something like 18 aircraft per hour (average)

assets/airports/rjbb.json
n8rzz: you could just remove `MAIKO` from the `rwy` section since it's already listed in `exitPoints`

erikquinn: >you could just remove MAIKO from the rwy section since it's already listed in exitPoints

This would be the solution to the second bullet. We also need something to resolve the first bullet. I would tend to think pulling up the chart for the procedure could point us in the right direction here.

indianbhaji:
Condensing the STARs to use the suffix feature wouldn't exactly work in our favour either - there are arrivals that span from A - E and we only have 4.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,1958 @@
...
+            "name": "Nanko One",
+            "rwy": {
+                "06L": ["B6L20", ["NANKO", "A80+"]],
+                "06R": ["B6R10", ["B6R11", "A25+"], "B6R13", ["NANKO", "A80+"]],
+                "24L": ["B4L10", ["NANKO", "A80+"]],
+                "24R": ["B4R20", ["NANKO", "A80+"]]
+            },
+            "exitPoints": {
+                "SIGAK": ["NANKO", "OGURA", "KYOTO", "SIGAK"],
+                "KCC": ["NANKO", "OGURA", "KYOTO", "SIGAK", "KCC"]
```
erikquinn: in cases like this, where there are fixes that are present in _all_ the `exitPoints`, or all `rwy`s, for reasons of readability, please move them to a `body` section:
```
"rwy": {
    "06L": ["B6L20"],
    "06R": ["B6R10", ["B6R11", "A25+"], "B6R13"]
    "24L": ["B4L10"],
    "24R": ["B4R20"]
}
"body": [["NANKO", "A80+"], "OGURA", "KYOTO"],
"exitPoints": {
    "SIGAK": ["SIGAK"],
    "KCC": ["SIGAK", "KCC"]
}
```

You also do not want to have fixes listed in multiple sections, for example with `NANKO` in the `rwy` and `exitPoints` sections above. That will lead to there being two waypoints at the same place, and it's possible that could cause issues.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,1958 @@
...
+        },
+        "DANDE": {
+            "icao": "DANDE",
+            "name": "Dande",
+            "suffix": {"06L":"B", "06R":"A", "24L":"C", "24R":"C"},
+            "entryPoints": {
+                "DANDE": ["DANDE"]
+            },
+            "body": [
+                ["JOLLY", "A80-|A60+"]
```
erikquinn: the game currently does not support an altitude range, only an altitude restriction. So only one of these will be honored. I recommend (since its a STAR) just including the `A80-` piece, as that will "force" the aircraft to descend to 8000 as it approaches `JOLLY`.

assets/airports/rjbb.json
```diff
@@ -0,0 +1,1958 @@
...
+                ["ups", 3]
+            ]
+        },
+        {
+            "origin": "",
+            "destination": "RJBB",
+            "category": "arrival",
+            "route": "SUC.BECKY.RJBB",
+            "altitude": 20000,
+            "speed": 430,
```
erikquinn: Please reduce this speed, such that it is no higher than `320`.

assets/airlines/jjp.json
```diff
@@ -0,0 +1,12 @@
...
+    "callsign": {
+        "name": "Orange Liner",
+        "length": 3
+    },
+    "fleets": {
+        "default": [
+            ["A320", 20]
+        ]
+    }
+}
```
erikquinn: Be sure to always include a single empty line at the end of a file.

assets/airlines/sfj.json
```diff
@@ -0,0 +1,12 @@
...
+    "callsign": {
+        "name": "Starflyer",
+        "length": 3
+    },
+    "fleets": {
+        "default": [
+            ["A320", 9]
+        ]
+    }
+}
```
erikquinn: Be sure to always include a single empty line at the end of a file.

Issue #109, opened 2017-01-17
"sid" / "star" commands need upper case procedures in readback


erikquinn:
_From @erikquinn on December 15, 2016 4:52_

See `AircraftInstanceModel.js:934`

_Copied from original issue: n8rzz/atc#204_

erikquinn:
_From @n8rzz on December 15, 2016 4:55_

@erikquinn do you mean they _should_ be lower case, or they are lower case and should not be?

erikquinn:
Sorry, yeah, they ARE being given in lower case, and it freaks me out to look at it, because they're never really shown in lower case. They should always be upper case.

And I should clarify one other thing as well, this is simply having to do with the readback issued by `runSID`. Below are two screenshots: the first is using the `caf` command, and the latter using `sid quiet7`.

![image](https://cloud.githubusercontent.com/assets/5103735/21212287/3a1d40be-c259-11e6-92f5-51d9edd29a72.png)

![image](https://cloud.githubusercontent.com/assets/5103735/21212353/cf758fa4-c259-11e6-9159-b3a6aeb8b052.png)

In both cases, `QUIET7` should be displayed, never `quiet7`.

erikquinn:
This is actually also the case with the `star` *and* `sid` commands.

PR #157, opened 2017-01-23
feature/109

Dabea:
This completes openscope/openscope#109  
This adds formatted variables so they are always uppercase when called

If we do not want the .toUppercase() in these functions option 2 would be to create methods in the flight management system. However changing case and formatting strings is normally the job of a model


src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1276,7 +1277,7 @@ export default class Aircraft {
         this.fms.followSID(route.toUpperCase());
 
         const readback = {
-            log: `cleared to destination via the ${sidId} departure, then as filed`,
+            log: `cleared to destination via the ${sidIdOutput} departure, then as filed`,
```
n8rzz: Is there a reason we would need it stored as a variable?  Believe it or not, I'm inclined to say we should just put this inline. Or, better yet, is this a value that could be obtained from the `StandardRouteCollection`?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -499,6 +499,7 @@ export default class Aircraft {
         }
 
         if (this.model.weightclass === 'U') {
+
```
n8rzz: could you remove this empty line?  I'm guessing this came over from a merge somewhere.

CHANGELOG.md
```diff
@@ -10,6 +10,16 @@
 
 
 
+### Minor
+- Makes sure the output for sid and start commands are always uppercase. [#109](https://github.com/openscope/openscope/issues/109)
```
n8rzz: star instead of start.  you have a `T` at the end of `star`

Issue #154, opened 2017-01-22
Remove github pages fluff


n8rzz:
Once the move to openscope has completed and openscope-prod is purring like a kitten:  
- [x] remove github pages remnants
- [x] verify master branch is top

---

after merge and before close:
- [ ] remove gh-pages branch

PR #190, opened 2017-01-29
feature/154

n8rzz:
completes #154 

removes github pages/jekyll files no longer needed with new deploy process


_config.yml
```diff
@@ -1,18 +0,0 @@
...
-# Enable Github flavored markdown
-kramdown:
-  input: GFM
-  hard_wrap: false
-
-# Set default values for documentation
-defaults:
-  -
-    scope:
-      path: "documentation"
```
n8rzz: I left the contents of the `documentation` folder intact, all of them.  We may use them at a later date.  We may even put together a real `/docs` endpoint.  Now that we have a big boy server, we can do those kinds of things 😉

Issue #145, opened 2017-01-19
Continued simplification and refactor of AircraftCommander class


erikquinn:
_From @n8rzz on January 15, 2017 4:54_

_note: This is a continuation of n8rzz/atc#181 which was merged under ( n8rzz/atc#236 )_

_Copied from original issue: n8rzz/atc#260_

n8rzz:
pulled in to and will be finished as a part of openscope/openscope#139

Issue #139, opened 2017-01-19
Refactor FMS


erikquinn:
#### story goals:
- accept a `routeString` or a collection of `FixModels` on instantiation
- add helper functions that can build `Leg`s and `Waypoint`s from a routeString
- make heading, speed altitude class properties _informed_ by fix restrictions
- consolidate `StandardRouteWaypointModel`, `Waypoint` and `FixModel`
    - this will likely necessitate a builder method within `FixModel` that can build a `Waypoint`
    - `StandardRouteWaypointModel` seems duplicative of `Waypoint`
- add method to `FixCollection` to generate new pooled `PositionModel` from coordinates that uses stored reference position

---
## ISSUES
- [ ] rr a plane repeatedly they will eventually just stop rr and just fly straight. but will say `rerouting to :XYZ`

## ADDITIONAL SPECS
Must also fulfill specifications from all of the below:
- [x] openscope/openscope#158 
- [x] openscope/openscope#129 
- [x] openscope/openscope#165 
- [x] openscope/openscope#114 
- [x] openscope/openscope#73 
- [x] openscope/openscope#87 
- [x] openscope/openscope#197 
- [x] openscope/openscope#277 
- [x] openscope/openscope#20 
- [x] openscope/openscope#153 
- [x] openscope/openscope#122

## DONE
- [x] parallel arrivals to 7r and 8 gives conflict notification even when both are established on localizer prior to descending on the glideslope
- [x] altitude readback and strip differ near lowest altitude (ex: at EDDM, assign `c 10`)
- [x] leader lines only show when aircraft is selected, and always point straight north
- [ ] ~~Add readbacks from aircraft that are asked to exceed their performance limitations (climb above ceiling, slow below `model.speed.min`, speed up above `model.speed.max`)~~
    - added [326](https://github.com/openscope/openscope/issues/326)
- [x] `rr FIX land` results in an undefined `destinationText` value
    - `AircraftStripView.updateViewForCruise()` has a switch on waypoint navmode, which is wrongfully assuming all aircraft in `CRUISE` mode are on `LNAV`. This needs to be reexamined, because rerouting and then clearing for approach throws here.
- [x] Airports with runways with elevations fail to load (ex: OTHH)
- [x] Video maps are not being drawn
- [x] You can taxi airborne aircraft to the runway
- [x] `[false, 'roger, reading back instructions']` boolean not causing red colored response when false
[...]

PR #193, opened 2017-01-30
feature/139

n8rzz:
completes openscope/openscope#139 

I have added you all to this one because there will be a lot of change involved with this PR and the more eyes on it the better.  Feel free to open up the review app to view current progress.

If you see something, say something!

depends on openscope/openscope#145 ( openscope/openscope#213 )


src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     *
+     * @for Fms
+     * @method _moveToNextLeg
+     */
+    _moveToNextLeg() {
+        this.currentLeg.destroy();
+        // this is mutable
+        this.legCollection.shift();
```
erikquinn: This idea of removing past navigation information is fine, but we do need to store somewhere the full route, for purposes such as `runSayRoute` or displaying the flightplan route in a flight strip. This could be a part of the aircraft instead of the fms if desired.

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -0,0 +1,44 @@
...
+        this.altitudeRestriction = -1;
+
+        // this.hold = {
+        //     dirTurns: null,
+        //     fixName: null,
+        //     fixPos: null,
+        //     inboundHd: null,
+        //     legLength: null,
+        //     timer: 0
+        // };
```
erikquinn: I assume you have better ideas of how to instate holds?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -229,6 +242,9 @@ export default class Aircraft {
         this.heading = _get(data, 'heading', this.heading);
         this.altitude = _get(data, 'altitude', this.altitude);
         this.speed = _get(data, 'speed', this.speed);
+
+        this._f = new Fms(data, this.initialRunwayAssignment, this._navigationLibrary);
```
n8rzz: very much so.  this is the _'I want to get it working without unhooking the previous implementation that has the name I was to use"_ trick

n8rzz: yeah, abbreviations === bad

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,180 @@
...
+        this.category = '';
+
+        this.init(aircraftInitProps);
+    }
+
+    /**
+     *
+     *
+     * @property currentWaypoint
+     * @return {FixModel|StandardRouteWaypointModel|undefined}
```
n8rzz: good catch here, an earlier implementation did not yet have the new `WaypointModel` so I was working with the classes listed there.

The short answer is no.  the `StandardRouteWaypointModel` give us functionality that the WaypointModel does not.  It has a very specific concern that I don't think we should generalize.  I _do_ think it could be simplified slightly, but I don't see a reason to get rid of it anytime soon.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
n8rzz: my thought with this was something along the lines of:
```javascript
get route() {
  const routeSegments = _map(this.legCollection, (legModel) => legModel.routeString);

  return routeSegments.join('..');
}
``` very crude, but hopefully you can see where I'm headed with that.

We may have to talk more about this one.  I don't see much need to say sections of a route that the aircraft is already past.  The exception is procedure routes, of course.  But with a procedure route the original routeString is stored as the route.  You might have to wait for the next commit to see that stuff though 😉

File: src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js (hunk 2 above)
Comment (n8rzz) [inline src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js]: ha!  well, not yet.  I'm not sure this is really a poor implementation, I mean it works doesn't it?  Once I get the main methods for the Fms and a basic Mcp up and running, holds are going to be the first thing I tackle.  So we'll know soon.

File: src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js [hunk 5]
```diff
@@ -0,0 +1,180 @@
...
+        this.legCollection = [];
+
+        /**
+         *
+         *
+         * @property category
+         * @type {string}
+         * @default ''
+         */
+        this.category = '';
```
n8rzz: the only time we need both `runway` and `category` is when we are building a `LegModel` for a `procedureRoute`.  Perhaps those can both go away?

I'm using category here the same way we use it with `SpawnPatternModels`.  Though true an aircraft could be all three of those, it is only one at any given time.  But, as I mentioned above, this property is only important when building a `procedureRoute`.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,124 @@
...
+        this._navigationLibrary = navigationLibrary;
+
+        /**
+         *
+         *
+         * @property _runway
+         * @type {string}
+         * @private
+         */
+        this._runway = initialRunwayAssignment;
```
n8rzz: see below

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: Well we can kinda deduce whether the aircraft is landing based on whether the last thing in their flightplan is the ICAO code of an airport, so the category isn't too necessary. The `runway` however will be needed to be known by the aircraft for a myriad of things.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: That would only give the portions of the route that haven't yet been completed. The entire route needs to be accessible, for example for displaying that route to the user in a flight strip.

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
erikquinn: Gotcha. If nothing else, one sucky thing about the way they worked was that the projection thing didn't know how it worked because it depended on game timers to time the legs.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
n8rzz: The exception is with a procedure route, where we _would_ have access to the original routeString in the LegModel until all the waypoints have been flown through.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,278 @@
...
+     * the active waypoint for that Leg.
+     *
+     * @for Fms
+     * @method skipToWaypoint
+     * @param waypointName {string}
+     */
+    skipToWaypoint(waypointName) {
+        const { legIndex, waypointIndex } = this._findLegAndWaypointIndexForWaypointName(waypointName);
+
+        this.legCollection = _drop(this.legCollection, legIndex);
```
n8rzz: adding a second list makes things doubly complex, because now you need to keep track of _two_ lists.  and you also need to managed how items move from one list to another.  Its an unnecessary complexity.  I'm more concerned with why you would need access to old portions of a route.  what are some examples of why we would need to store an old part of a route?

n8rzz: oh, and to answer your question: we drop them because then moving through legs and waypoints is super clean.  the first item in waypoints and the first item in legs is assumed to _always_ be active.  we can always add a waypoint to the left or right of the list if we need to (not fully implemented yet).  doing it this way means we don't need to keep track of two sets of indicies; one for legs and one for waypoints.  Doing that requires an entire interface to manage moving them, resetting them, etc.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,278 @@
...
+
+        this.currentLeg.skipToWaypointAtIndex(waypointIndex);
+    }
```
n8rzz: that is my plan, though not implemented yet.  we can't skip to a fix if it doesn't exist in the flight plan.

I haven't quite made it to the error handling sections yet, but these are good ideas for test cases ;-)

src/assets/scripts/client/aircraft/ModeControl/modeControlConstants.js
```diff
@@ -0,0 +1,49 @@
...
+    HOLD: 'HOLD',
+    LNAV: 'LNAV',
+    OFF: 'OFF',
+    VOR_LOC: 'VOR_LOC'
+};
+
+export const MCP_AUTOPILOT_MODES = {
+    ON: 'ON',
+    OFF: 'OFF'
+};
```
n8rzz: we can control that pretty easily with a class method.  meaning, when we turn autopilot on we do those checks.

src/assets/scripts/client/aircraft/ModeControl/modeControlConstants.js
```diff
@@ -0,0 +1,49 @@
...
+    ALTITUDE: 'altitudeMode',
+    AUTOPILOT: 'autopilotMode',
+    HEADING: 'headingMode',
+    SPEED: 'speedMode'
+};
+
+export const MCP_PROPERTY_MAP = {
+    ALTITUDE: 'altitude',
+    AUTOPILOT: 'autopilot',
+    HEADING: 'heading',
```
n8rzz: the term _fields_, in the front end space, is more often a reference to forms and inputs.  although it may be an acceptable aviation term, it is not a good name here because the values are not fields.  

The name actually leads you to its use.  `property` is a programming term.  This object is a map of mcp properties, where we map a, object key to a _property_ value.  that means we can do things like `this._modeController[MCP_PROPERTY_MAP.ALTITUDE]`.

The stuff in our object are actually _properties_ of the class.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -39,6 +41,51 @@ const ALTITUDE_RESTRICTION_PREFIX = 'A';
...
+ * Symbol denoting a less than restriction
+ *
+ * @property ABOVE_SYMBOL
+ * @type {string}
+ * @final
+ */
+const BELOW_SYMBOL = '-';
+
+/**
+ * Number to used to cnovert a FL altitude to an altitude in thousands
```
n8rzz: done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -39,6 +41,51 @@ const ALTITUDE_RESTRICTION_PREFIX = 'A';
...
+ *
+ * @property ABOVE_SYMBOL
+ * @type {string}
+ * @final
+ */
+const FL_TO_THOUSANDS_MULTIPLIER = 100;
```
n8rzz: good callout, I believe we have a helper function to do this logic already

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -39,6 +41,51 @@ const ALTITUDE_RESTRICTION_PREFIX = 'A';
...
+
+/**
+ * Enemuration for an invalid index number.
```
n8rzz: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: >I'm more concerned with why you would need access to old portions of a route. what are some examples of why we would need to store an old part of a route?

Because you just _do_. From the air traffic control perspective, having passed a fix changes nothing about your route. If I ask the system what an aircraft's route is, I will get the whole thing, not just the remaining portion. From the aircraft's perspective, if it's in the past, it's gone.

Any time we expose the route to the controller, it should be shown in full, not in part. If there's no way to achieve this, we can break from reality here to avoid unnecessary effort, but the expectation would be that the controller would see the entire flightplan route. Other than visually comparing with where they see the aircraft on the screen, they wouldn't even have the ability to get the "remaining portion" of the route, so it would just feel really weird to see it that way. Especially if they arrive at the end of the STAR, and receive vectors. Their "route" shown to the controller would be... empty.

src/assets/scripts/client/aircraft/ModeControl/modeControlConstants.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const MCP_MODES = {
+    ALTITUDE: MCP_ALTITUDE_MODES,
+    AUTOPILOT: MCP_AUTOPILOT_MODES,
+    HEADING: MCP_HEADING_MODES,
+    SPEED: MCP_SPEED_MODES
+};
+
+export const MCP_MODE_NAME = {
```
erikquinn: Then keeping with _that_ pattern, should the others that are plural be changed to `MCP_HEADING_MODE.HOLD`, etc?

src/assets/scripts/client/aircraft/ModeControl/modeControlConstants.js
erikquinn: Yeah, I understand. Though it is technically helpful in terms of indicating the structure, I don't find it intuitive or explanatory in terms of what it does. Just in the sense that it would be harder to explain to aviation-background newcomers what it does. But since I don't have any good alternatives, I guess that works.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,481 @@
...
+        /**
+         * Name of the initial runway assigned for takeoff/landing.
+         *
+         * This value is likely to change as an aircraft moves through the airspace.
+         *
+         * @property _runway
+         * @type {string}
+         * @private
+         */
+        this._runwayName = initialRunwayAssignment;
```
erikquinn: property name doesn't match docblock.

We could store runway assignments in the fms altogether I suppose. They shouldn't exist _both_ in the fms and the aircraft, so we could move all of that to here to centralize things a bit.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -0,0 +1,481 @@
...
+        /**
+         * Current flight phase of an aircraft
+         *
+         * Currently only supports `arrival` and `departure`
+         *
+         * @property currentPhase
+         * @type {string}
+         * @default ''
+         */
+        this.currentPhase = '';
```
erikquinn: This seems like something that may belong more to the aircraft than the fms. But if you don't agree, it's okay to stay here.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -131,7 +131,7 @@ export default class Aircraft {
          */
 
         // Initialize the FMS
-        this.fms = new AircraftFlightManagementSystem({
+        this.__fms__ = new AircraftFlightManagementSystem({
```
n8rzz: yes, easy to search and replace.  or just find in general.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -158,12 +158,24 @@ export default class Aircraft {
...
         this.updateStrip();
     }
 
     /**
+     * The name of the currently assigned runway
+     *
+     * @property initialRunwayAssignment
+     * @return {string}
+     */
+    get initialRunwayAssignment() {
```
n8rzz: done

src/assets/scripts/client/math/flightMath.js
```diff
@@ -154,6 +142,76 @@ export const calculateDistanceToBoundary = (airport, pos) => {
...
+    const nominalNewCourse = _calculateNominalNewCourse(
+        aircraft.fms.getNextWaypointPositionModel(),
+        currentWaypointPosition
+    );
+    const courseChange = _calculateCourseChangeInRadians(currentHeading, nominalNewCourse);
+    // meters, bank establishment in 1s
+    const turnInitiationDistance = calcTurnInitiationDistance(speed, bankAngle, courseChange);
+
+    // convert m to km
+    return turnInitiationDistance / 1000;
```
n8rzz: there is an enumeration for this

PR #214, opened 2017-02-05
feature/179

### LINKED ISSUE #179: Update WIP flags
There are many airports that do not have fully completed elements, or are unreliable for not yet determined reasons. These should all be marked as WIP (work in progress) airports until the time that they are completed.

I'd like to move more toward the _lack_ of a flag meaning that an airport is 100% filled out and realistic. And the rest are being worked on. But if there's no little flag, any issues are the fault of the sim, not the airport file. As such, we will need to keep all non-WIP airports fully compliant with new features as they are released.

erikquinn:
Updates `wip` flags for all airports.

Resolves #179.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Minor
 - Makes sure the output for sid and star commands are always uppercase. [#109](https://github.com/openscope/openscope/issues/109)
-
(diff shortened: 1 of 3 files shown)


erikquinn:
Yeah, me too. Fixed.

PR #215, opened 2017-02-06
bugfix/210

### LINKED ISSUE #210: Data block colored bar width inconsistent
image

The weird callsign business notwithstanding (see #164), the colored bar on the left of the data block is being scaled up and down based on the width of the data block. It should instead be a constant number of pixels (2 or 3 or so?).

erikquinn:
Make data block colored bar constant width.

Resolves #210.


erikquinn:
whoops. fixed both.

n8rzz:
:shipit:

PR #251, opened 2017-02-19
hotfix/249

### LINKED ISSUE #249: GA aircraft not spawning at EDDT
GA aircraft do not spawn, due to a casing issue: `n/fastga` instead of `n/fastGA`.

See PR: #251

erikquinn:
Restore spawning of GA aircraft at EDDT.

Resolves #249.

### DIFF
--- a/assets/airports/eddt.json
+++ b/assets/airports/eddt.json
@@ -313,7 +313,7 @@
           "method": "random",
           "rate": 7,
           "airlines": [
-              ["n/fastga", 5],
+              ["n/fastGA", 5],
               ["ber", 20],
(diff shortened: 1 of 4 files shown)


assets/airports/eddt.json
```diff
@@ -313,7 +313,7 @@
           "method": "random",
           "rate": 7,
           "airlines": [
-              ["n/fastga", 5],
+              ["n/fastGA", 5],
```
n8rzz: id be more inclined to use `fastga` here instead of `fastGA`.  but for the purposes of the hotfix, this should do it.

PR #260, opened 2017-02-20
bugfix/259

### LINKED ISSUE #259: A332 & A343 files mistakenly cleared out
Oh, Erik.... :disappointed: 

Fix your mess.

Sincerely,
Erik

---

See #260.

erikquinn:
Restore content for mistakenly cleared out aircraft files. This will fix the failing build on `develop`.

Resolves #259.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
 ### Bugfixes
 - Standardized indentation in all json files [#256](https://github.com/openscope/openscope/issues/256)
-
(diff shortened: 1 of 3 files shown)


erikquinn:
User error. They weren't removed. I had a macro set up on the keyboard and was rifling through files as they were being reformatted. There were occasional misfires as atom couldn't keep up with the blazing speed at which the key macros were executed, and so it would enter the remaining keystroke (a lowercase L) _in_ the file. Evidently I did not catch it when these two files exhibited it.

n8rzz:
:man_facepalming: 

yay automation!!

PR #261, opened 2017-02-20
feature/138

### LINKED ISSUE #138: Allow optional id prefixes in BaseModel and BaseCollection
_From @n8rzz on January 7, 2017 5:11_

Each class is an inheritable and uses `_uniqueId()` to generate a value for `_id`. This task would allow for an optional prefix to that property.

- [x] support a string prefix with output like `positionModel-42`
- [x] support no-prefix where default output is `baseModel-42`

_Copied from original issue: n8rzz/atc#255_

Dabea:
Resolves #138.

Add additional context to the  ._id property by accepting an optional parameter as a string
Also if given an object will be able to pull some information as well like airport name or fix.


src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +45,27 @@ export default class BaseModel {
...
     }
+
+    /**
+     * This will verify if the given argument is a string and will return a modified pre-value for the _id
+     *
+     * @for BaseModel
+     * @method veriftyModelName
+     * @param {string / Object}
+     */
+    verifyModelName(modelName) {
```
n8rzz: are we _verifying_ a model name here?  or are we _building_ something?

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,49 @@ import _isString from 'lodash/isString';
...
+});
+
+ava('instantiates with an _id with string property', t => {
+    const resultString = new BaseModel('string');
+
+    t.true(_isString(resultString._id));
+});
+
+ava('instantiates with an _id with an Object', t => {
+    const resultObj = new BaseModel({});
```
n8rzz: this should only ever accept a string, this test can be removed

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,49 @@ import _isString from 'lodash/isString';
...
+
+    t.true(_isString(resultObj._id));
+});
+
+ava('instantiates with an _id with an array', t => {
+    const resultArray = new BaseModel([]);
```
n8rzz: this should only ever accept a string, this test can be removed

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,49 @@ import _isString from 'lodash/isString';
...
+
+    t.true(_isString(resultArray._id));
+});
+
+ava('instantiates with an _id with an boolean', t => {
+    const resultBol = new BaseModel(false);
```
n8rzz: this should only ever accept a string, this test can be removed

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,49 @@ import _isString from 'lodash/isString';
...
+
+    t.true(_isString(resultBol._id));
+});
+
+ava('instantiates with an _id with an int', t => {
+    const resultInt = new BaseModel(42);
```
n8rzz: this should only ever accept a string, this test can be removed

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +45,27 @@ export default class BaseModel {
...
+        if (typeof modelName === 'string') {
```
n8rzz: there is a lodash helper for this `_isString()`

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -8,15 +8,15 @@ import _uniqueId from 'lodash/uniqueId';
...
+    constructor(modelName = 'BaseModel') {
         /**
          * Unigue string id that can be used to differentiate this model instance from another.
          *
          * @property _id
          * @type {string}
          * @private
          */
-        this._id = _uniqueId();
+        this._id = this.verifyModelName(modelName) + _uniqueId();
```
n8rzz: additionally, `_uniqueId` accepts a string that can be used as the id base.  so in our case: 
```
_uniqueId(`${modelName}-`)
```

test/base/BaseModel.spec.js
Dabea: There are cases in the code where it is currently being called by other parameters. I was confused on why that was the case though.

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +45,27 @@ export default class BaseModel {
...
+     * @method veriftyModelName
+     * @param {string / Object}
+     */
+    verifyModelName(modelName) {
+        if (typeof modelName === 'string') {
+            return `${modelName}-`;
+        }
+
+        // This will give more infomation on airpot names and runways to help identify if there is an issue.
+        if (modelName.name === undefined && typeof modelName === 'object') {
```
Dabea: This will show additional information on the object that is used to create it give you `_id`  values like 
`BaseModel-Caribbean Airlines-1087`  So you can see if it is only happening with certain Data, But I can just remove it.
I did an oversimplification when i was cleaning up the logic here
This would be a more accurate version
```
if (modelName.name === undefined) {
            modelName.name = '';
        }

        // This will give more information on airport names and runways to help identify if there is an issue.
        if (typeof modelName === 'object') {
            console.log(modelName);
            return `BaseModel-${modelName.name}-`;
        }
```

test/base/BaseModel.spec.js
n8rzz: hmm, the inheriting classes should not need to send their parameters to the parent with `super(value, value, value)` etc.

If you find any of those, go ahead and remove them (carefully).  Any class inheriting `BaseModel` should have an empty call to `super()`.

src/assets/scripts/client/base/BaseModel.js
n8rzz: I see what you're doing there.  That actually isn't the intent.  I may not have explained the terms here very well (and they do collide a little, which is confusing).

This class, and all classes inheriting it, has an `_id` property.  Without any input from us, this `_id` will be a string of numbers that increments every time this class is instantiated.  so you'll end up with `_id`s that look like: `this._id = 3432` and so on.  

My intent for this is for an inheriting class to have the ability to send an optional sting that can be added to the `_id` that helps differentiate each instance.  so say we have the `FixModel`, which inherits the `BaseModel`.  In the `FixModel` constructor, we would call super with that optional `_id` string : `super('FixModel')` which would then be added to that id.  this would give us an `_id` of `FixModel-3432`.

I think the _model_ you are thinking of here is actually the `AircraftTypeDefinitionModel` or the `AirlineModel`, those are different things entirely.  So this block can go away.  It was a good thought, but thats not what I'm looking for here.

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -47,23 +49,17 @@ export default class BaseModel {
...
      *
      * @for BaseModel
      * @method veriftyModelName
      * @param {string / Object}
      */
-    verifyModelName(modelName) {
-        if (typeof modelName === 'string') {
+    buildModelPrefix(modelName) {
+        if (_isString(modelName)) {
             return `${modelName}-`;
```
n8rzz: This if logic should be reversed.  We should be exclusive, not inclusive, with our logic blocks.
```javascript
// instead of (inclusive)
if (_isString(modelName)) { 
  return `${modelName}-`;
}

// we should do (exclusive)
if (!_isString(modelName)) {
  return `${modelName}-`;
}
``` 
a method should eject (return early) as soon as we find data we don't want.

Also, since `modelName` is defaulted to `BaseModel` in the constructor, we shouldn't need to return that string again below.  We can just use `modelName-`.

So with that in mind, does the value we return change?  Doesn't look like it.  But we do need to do something in the case we receive bad data, and handle that appropriately.  In this case, we should use `console.error` if we receive a param that is not a string, but still return our `modelName-` because that kind of error shouldn't prevent the model from instantiating.

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -8,15 +9,16 @@ import _uniqueId from 'lodash/uniqueId';
...
+    constructor(modelName = 'BaseModel') {
         /**
          * Unigue string id that can be used to differentiate this model instance from another.
          *
          * @property _id
          * @type {string}
          * @private
          */
-        this._id = _uniqueId();
+        const optionalModelPrefix = this.buildModelPrefix(modelName);
```
n8rzz: this line should be moved to just under the constructor declaration

CHANGELOG.md
```diff
@@ -13,8 +13,8 @@
 
 
 ### Features
-
-
+- Fixes an instance of two runways called "34R" in Shanghai Pudong [#149](https://github.com/openscope/openscope/issues/149)
```
Dabea: tried git pull origin develop then pushed the changes however this diff did not go away

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -8,15 +8,15 @@ import _uniqueId from 'lodash/uniqueId';
  * @class BaseModel
  */
 export default class BaseModel {
-    constructor() {
+    constructor(modelName = 'BaseModel') {
```
Dabea: Added tests to check the default param

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +49,22 @@ export default class BaseModel {
...
+     * This will verify if the given argument is a string otherwise it will return 'Base Model'
+     *
+     * @for BaseModel
+     * @method veriftyModelName
+     * @param {string / Object}
+     */
+    buildIdPrefix(modelName) {
+        if (!_isString(modelName)) {
+            console.error('BaseModel#constructor expects a string for its first paramert but a string was not given');
+            return 'BaseModel-';
```
n8rzz: the `console.error` should be a `throw` instead.

since `modelName` is defaulted to `BaseModel`, the return here is redundant and can be removed.

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +49,22 @@ export default class BaseModel {
...
     }
+
+    /**
+     * This will verify if the given argument is a string otherwise it will return 'Base Model'
+     *
+     * @for BaseModel
+     * @method veriftyModelName
+     * @param {string / Object}
+     */
+    buildIdPrefix(modelName) {
```
n8rzz: could we make this method private?

src/assets/scripts/client/base/BaseModel.js
Dabea: So I should change it to 
`return `${modelName}-` 

Or should the default be changed to 'BaseModel-'

n8rzz: `return ${modelName}`;

Dabea: Just to be clear is the desired out put to look like
BaseModel-1234   Or
BaseMode1234

n8rzz: correctL `BaseModel-1234` will be the end result.  and since `_uniqueId()` takes care of the `1234` part of it, we only need the `BaseModel-` part.

src/assets/scripts/client/base/BaseModel.js
Dabea: By private do you mean just giving it a naming scheme with an _ in front of the method name and adding @private in the doc blocks?

src/assets/scripts/client/base/BaseModel.js
Dabea: So then the default parameter should be  changed to `BaseModel-`? and just return ${modelName}?

src/assets/scripts/client/base/BaseModel.js
n8rzz: that is correct 👍

src/assets/scripts/client/base/BaseModel.js
n8rzz: no, the default should stay the same.  this is the correct way to do it, by adding the `-` in the return.

Dabea: Also if I do a `throw` then it will never hit the return statement and things will break? (not sure on the term obtrusive error handling?) Or should this be non-obtrusive and just give a default response?

n8rzz: Yep, we want them to break here because the developer is doing something they shouldn't be doing. 

See my first comment about the return, the throw should be the only thing in this block.

Dabea:
Earlier you had me change the error to obtrusive error handling however we previously built theses tests with Ava. Should I remove theses tests or just change them to fail on everything  but pass when a string or nothing is given?

```ava('sets the #_id when passed invalid parameters', (t) => {
    const modelojb = new BaseModel({});
    t.true(modelojb._id.indexOf('BaseModel') !== -1);

    const modelArray = new BaseModel([]);
    t.true(modelArray._id.indexOf('BaseModel') !== -1);

    const modelInt = new BaseModel(42);
    t.true(modelInt._id.indexOf('BaseModel') !== -1);

    const modelBool = new BaseModel(false);
    t.true(modelBool._id.indexOf('BaseModel') !== -1);
});

ava('makes sure the model in instantiates with any data type', t => {
    t.notThrows(() => new BaseModel());
    t.notThrows(() => new BaseModel([]));
    t.notThrows(() => new BaseModel({}));
    t.notThrows(() => new BaseModel(42));
    t.notThrows(() => new BaseModel('string'));
    t.notThrows(() => new BaseModel(false));
});
```

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -156,7 +157,7 @@ export default class StandardRouteModel extends BaseModel {
          */
         this._entryCollection = null;
 
-        return this._init(standardRoute);
```
n8rzz: why did you remove this return?

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
 import BaseModel from '../../src/assets/scripts/client/base/BaseModel';
 import ExtendedBaseModelFixture from './_fixtures/ExtendedBaseModelFixture';
 
+ava('sets the #_id when passed valid parameters', (t) => {
```
n8rzz: this test can be improved a bit.
1) if you're going to do multiple assertions inside of a test you should use a single var to hold the results:
```javascript
let model = new BaseModel();
// ...
model = new BaseModel('string');
```

2) use a more realistic example for the passed model name:
```javascript
// instead of
modelBool = new BaseModel('string');

// do
modelBool = new BaseModel('modelName');
```  This adds a little more context to the test and informs a future developer about the intent.

File: test/base/BaseModel.spec.js [hunk 17]
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
...
+    const modelInt = new BaseModel();
+    t.true(modelInt._id.indexOf('BaseModel') !== -1);
+
+    const modelBool = new BaseModel('string');
+    t.true(modelBool._id.indexOf('string') !== -1);
+    console.log(modelBool._id);
```
n8rzz: errant console.log

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
...
+});
+
+ava('makes sure the model fails to instantiate with any incorrect', t => {
```
n8rzz: take a look at some of the other test files and try to stay close to the convention established there with how this type of test is worded.  its usually along the lines of: `throws when passed invalid parameters` or something like that.

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
...
+    t.notThrows(() => new BaseModel());
+    t.throws(() => new BaseModel([]));
+    t.throws(() => new BaseModel({}));
+    t.throws(() => new BaseModel(42));
+    t.notThrows(() => new BaseModel('string'));
+    t.throws(() => new BaseModel(false));
```
n8rzz: the next test validates the `notThrows` cases, so they can be removed from this test.

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
...
+});
+
+ava('makes sure the model in instantiates with the correct data type', t => {
+    t.notThrows(() => new BaseModel());
+    t.notThrows(() => new BaseModel('string'));
+});
+
 ava('instantiates with an _id property', t => {
-    const result = new BaseModel();
+    const resultEmpty = new BaseModel();
```
n8rzz: be sure to use `model` as the const name.  this is just an [undocumented] convention we've established in our tests to make it more readable.

test/base/BaseModel.spec.js
```diff
@@ -4,10 +4,39 @@ import _isString from 'lodash/isString';
...
+
+    t.true(_isString(resultEmpty._id));
+});
+
+ava('instantiates with an _id with string property', t => {
```
n8rzz: this test is redundant and can be removed

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
Dabea: I was tracking down a bug,  Accidentally left it out after it was found. Adding it back in. I Should have commented it out and added the `this._init(standardRoute);` as its own line with a comment showing that though

test/base/BaseModel.spec.js
Dabea: Made the changes now uses a common variable name and the given string is now modelName

src/assets/scripts/client/base/BaseModel.js
```diff
@@ -45,4 +49,22 @@ export default class BaseModel {
...
+     * This will verify if the given argument is a string otherwise it will return 'Base Model'
+     *
+     * @for BaseModel
+     * @method veriftyModelName
+     * @param {string / Object}
+     * @private
+     */
+    _buildIdPrefix(modelName) {
+        if (!_isString(modelName)) {
+            throw new TypeError('BaseModel#constructor expects a string for its first paramert but a string was not given');
```
n8rzz: paramert -> parameter

CHANGELOG.md
```diff
@@ -68,8 +68,11 @@
 
 ### Features
 - Enumerate magic number in RunwayModel [#269](https://github.com/openscope/openscope/issues/269)
+
+- Adds more contex to the Model classes by adding an optional input paramater [#138](https://github.com/openscope/openscope/issues/138)
```
n8rzz: make sure this is in the correct spot in the current chagnelog.  this diff looks like its from pre v5.0.0

erikquinn: Fixed.

PR #262, opened 2017-02-20
feature/40

### LINKED ISSUE #40: Autocomplete callsign by flight number
Select aircraft by matching a portion of the flight number (which should be globally unique). For further details, see the original bug report.

Originally reported in zlsa/atc#720.

Dabea:
Resolves #40.

Adds another check to allow a match by just the aircraft call sign is given by using the `this.callsign`.


src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -463,17 +463,24 @@ export default class Aircraft {
...
-        }
-
-        return _isEqual(callsignToMatch.toUpperCase(), this.getCallsign());
-    }
+     matchCallsign(callsignToMatch) {
+         if (callsignToMatch === '*') {
+             return true;
+         }
+         // checks to see if the given call sign matches the airline Id + callsign format
+         if (_isEqual(callsignToMatch.toUpperCase(), this.getCallsign())){
```
n8rzz: I'd like to see the first if in this method removed.  What you could do instead is:
```javascript
const shouldMatchAnyCallsign = callsignToMatch === '*';
```

then in this if you can do:
```javascript
if (shouldMatchAnyCallsign || _isEqual(callsignToMatch.toUpperCase(), this.getCallsign())) { }
```

n8rzz:
@Dabea lets finish the other two merge requests you have open and circle back to this one.  That'll give us time to let the `develop` branch stabilize and we can safely verify these changes.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -371,16 +371,22 @@ export default class AircraftInstanceModel {
...
      * @for AircraftInstanceModel
      * @method matchCallsign
      * @param callsign {string}
      */
     matchCallsign(callsignToMatch) {
-        if (callsignToMatch === '*') {
+        const shouldMatchAnyCallsign = callsignToMatch === '*';
+
+        // checks to see if the given call sign matches the airline Id + flight number
+        if (shouldMatchAnyCallsign || _isEqual(callsignToMatch.toUpperCase(), this.callsign)) {
```
erikquinn: Couldn't we get away with simply `===` here? It would be more performant, and since we aren't comparing arrays, I don't think there's any real advantage to using lodash `_isEqual`. Same comment for the other line.

n8rzz: _(already called out on slack, copying here for visibility)_
As long as we are comparing strings, I agree with Mr Quinn.  Lodash does a lot under the hood for [_isEqual()](https://github.com/lodash/lodash/blob/4.17.4/lodash.js#L11545) that is overkill in this case.  Its a good thought here, but a simple strict equals (`===`) will definitely be more performant.

tools/tasks/jsonAssembler.js
```diff
@@ -32,6 +32,8 @@ function _buildResultList(source, outputFilename) {
             return;
         }
 
+        gutil.log(fileSource);
```
n8rzz: did you add this?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -377,6 +377,18 @@ export default class AircraftInstanceModel {
     }
 
     /**
+     * Is an aircraft's current heading within a specific range
+     *
+     * @for AircraftInstanceModel
+     * @method isHeadingInsideDepartureWindow
+     */
+    isHeadingInsideDepartureWindow() {
```
n8rzz: did you add this one too?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -407,6 +395,10 @@ export default class AircraftInstanceModel {
     }
 
     /**
+    * verifies if there is a matched callsign and if the  aircraft is visable.
```
n8rzz: there are some extra spaces here that should be removed

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -28,6 +29,7 @@ import { TIME } from '../constants/globalConstants';
 import { COLOR } from '../constants/colors/colors';
 import { THEME } from '../constants/colors/themes';
 
+
```
n8rzz: this extra line should be removed

tools/tasks/jsonAssembler.js
```diff
@@ -32,6 +32,7 @@ function _buildResultList(source, outputFilename) {
             return;
         }
 
+
```
n8rzz: this extra line should be removed

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -377,19 +377,39 @@ export default class AircraftInstanceModel {
...
     matchCallsign(callsignToMatch) {
-        if (callsignToMatch === '*') {
+        const shouldMatchAnyCallsign = callsignToMatch === '*';
+         // checks to see if the given call sign matches the airline Id + callsign format
+        if (shouldMatchAnyCallsign || (this.airlineId.toUpperCase() + callsignToMatch.toUpperCase() === this.callsign)) {
             return true;
         }
 
+        // Checks to see if the given callsign matches only the callsign since callsign numbers should be unique
         return _isEqual(callsignToMatch.toUpperCase(), this.callsign);
```
n8rzz: does this method work the same way when used with any of the following:
- `AAL123`
- `123`
- `AaL123`

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: and `visible`

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
Dabea: Yes it works the same no matter what casing is used.  is there a way to generate the aircraft info in the test file for aircraftInstanceModel?

PR #268, opened 2017-02-24
Enumerate magic number 3280 in RunwayModel

NathanDJohnson:
The conversion factor from kilometers to feet is ~3280.84.

gs_gradient also must be in radians.


n8rzz:
Thank you for your contribution!  A few things need to happen before we can merge this in:
- I have created a corresponding issue to go along with these changes ( #269 )
- @erikquinn will need to add you to the organization so you can make commits directly to the repo (as long as its not to `develop` or `release`)
- erik will also create a feature branch within this repo and pull your changes into that.  any feedback we have can then be addressed in that branch.  this way our ci and coverage reports can run on your code.

---

- be sure to verify tests are still passing and fix any that are not.
- please add a changelog entry for these changes.  you can follow the example from previous entries.

erikquinn:
Closing in favor of #274.

PR #297, opened 2017-04-03
feature/191

### LINKED ISSUE #191: Create isEmptyObject helper
```javascript
if (!_isObject() || _isEmpty()) { ... }
```
This is a very common pattern throughout the app.

- [ ] investigate if this is an existing method within lodash

if not:
- [ ] abstract to a helper function
- [ ] add tests
- [ ] update usages
- [ ] update existing tests to test only A vs B logic branches for new function

---

the following tests should all pass:
```javascript
ava('.isEmptyObject() returns false when passed a non-object', (t) => {
    t.false(isEmptyObject('threeve'));
    t.false(isEmptyObject(false));
    t.false(isEmptyObject(42));
});

ava('.isEmptyObject() returns false when passed an object with properties', (t) => {
    t.false(isEmptyObject({
        a: 'threeve',
        b: 42,
        c: false
    }));
});

ava('.isEmptyObject() returns true when passed an empty object', (t) => {
    t.true(isEmptyObject({}));
    t.true(isEmptyObject([]));
    t.true(isEmptyObject(null));
});
```

Dabea:
Resolves #191.

Addes a object helper class for object validation.


src/assets/scripts/client/utilities/validatorUtilities.js
```diff
@@ -0,0 +1,36 @@
...
+import _isObject from 'lodash/isObject';
+
+/**
+ * This will return true if it is not an object or it is empty
+ *
+ * @funtion isNotObjectOrIsEmpty
+ * @param input {object}
+ * @return {boolean}
+ */
+export const isObjectEmpty = (value) => {
```
n8rzz: this should be renamed to `isEmptyObject`.  be sure to update the methods below with this name, too.

src/assets/scripts/client/utilities/validatorUtilities.js
```diff
@@ -0,0 +1,36 @@
...
+    return !_isObject(value) || _isEmpty(value);
```
n8rzz: add some tests to make sure the logic is correct here.  we want this to return true only when an object is empty.

src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
```diff
@@ -411,15 +412,15 @@ export default class SpawnPatternModel extends BaseModel {
      * @param airportController {AirportController}
      */
     init(spawnPatternJson, navigationLibrary, airportController) {
-        if (!_isObject(spawnPatternJson) || _isEmpty(spawnPatternJson)) {
+        if (_isEmpty(spawnPatternJson)) {
```
n8rzz: Could you add a cplease moment here about why we are returning early

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -12,11 +12,14 @@ import RouteModel from '../navigationLibrary/Route/RouteModel';
...
 import { convertStaticPositionToDynamic } from '../base/staticPositionToDynamicPositionHelper';
 import { speech_say } from '../speech';
+import { tau } from '../math/circle';
 import { abs } from '../math/core';
 import { distance2d } from '../math/distance';
-import { vlen } from '../math/vector';
-import { km } from '../utilities/unitConverters';
+import { calcTurnInitiationDistance } from '../math/flightMath';
+import { vlen, vradial, vsub } from '../math/vector';
+import { km, kn_ms, radiansToDegrees, degreesToRadians } from '../utilities/unitConverters';
```
n8rzz: hmm, any idea on why all these extra imports show up here as additions?

Dabea: I looked this over and it looks like it happens after a merge I believe it was this merge d7fbc7bb94b9fc528512018f0d467c3181b0ef1b  and that reworked and changed some things in the aircraft controller

n8rzz: verify that these additional imports are no longer needed and remove them if you could.

src/assets/scripts/client/utilities/validatorUtilities.js
```diff
@@ -0,0 +1,25 @@
+import _isArray from 'lodash/isArray';
+import _isEmpty from 'lodash/isEmpty';
+import _isObject from 'lodash/isObject';
+
+/**
+ * This will return true if it is not an object or it is empty
+ *
+ * @funtion isNotObjectOrIsEmpty
+ * @param input {object}
```
n8rzz: this param name should be value  and the type should be listed as `{*}` because is accepts _any_ type.

src/assets/scripts/client/utilities/validatorUtilities.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+export const isEmptyObject = (value) => {
+    return _isObject(value) && _isEmpty(value);
+};
+
+/**
+ * This will return true if it is not an array or it is empty
+ *
+ * @funtion isNotObjectOrIsEmpty
+ * @param input {object}
```
n8rzz: this param name should be value  and the type should be listed as `{*}` because is accepts _any_ type.

test/utilities/validatorUtilities.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    t.true(isEmptyOrNotArray({}));
+});
+
+ava('.isEmptyOrNotArray returns true when passed an empty object', (t) => {
+    t.true(isEmptyOrNotArray({}));
+    t.true(isEmptyOrNotArray([]));
+    t.true(isEmptyOrNotArray(null));
+});
+
+ava('.isEmptyOrNotArray return false when passed an array with arguments', (t) => {
```
n8rzz: ..passed and array with values

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -15,9 +15,9 @@ import { speech_say } from '../speech';
...
 import { abs } from '../math/core';
 import { distance2d } from '../math/distance';
-import { calcTurnInitiationDistance } from '../math/flightMath';
-import { vlen, vradial, vsub } from '../math/vector';
-import { km, kn_ms, radiansToDegrees, degreesToRadians } from '../utilities/unitConverters';
+import { isEmptyOrNotArray } from '../utilities/validatorUtilities';
+import { vlen } from '../math/vector';
+import { km } from '../utilities/unitConverters';
 import { FLIGHT_CATEGORY } from '../constants/aircraftConstants';
 import { isEmptyOrNotArray } from '../utilities/validatorUtilities';
```
n8rzz: looks like there is a duplicate import in here, that is why your tests were failing on the CI.  You might want to check the other files too just to be sure, or take a look at the CI and see where its failing (you can click the link at the bottom of the page).

src/assets/scripts/client/utilities/validatorUtilities.js
```diff
@@ -0,0 +1,25 @@
+import _isArray from 'lodash/isArray';
+import _isEmpty from 'lodash/isEmpty';
+import _isObject from 'lodash/isObject';
+
+/**
+ * This will return true if it is not an object or it is empty
+ *
+ * @funtion isNotObjectOrIsEmpty
+ * @param input {*}
```
n8rzz: these two doc block `param`s should be `value` instead of `input`

erikquinn: Fixed.

PR #341, opened 2017-04-18
bugfix/340

### LINKED ISSUE #340: Aircraft not reaching targeted speed
Sometimes aircraft will get "stuck" at a speed that is less than 0.1 knots away from the target speed, and not be updated, due to the fact that we were explicitly allowing a small level of error in the code. Really though, that is not needed, and should be removed to prevent cases like this.

This is the reason on a rare occasion, aircraft will start their takeoff roll, _appear_ to reach their rotation speed, and then just never rotate. It's because they rotate at 130, and are remaining at 129.994303546375. If the aircraft happens to calculate that speed between physics updates, they would end up getting stuck there forever.

Need to change this so aircraft will insist upon reaching their exact targeted speed.

erikquinn:
Resolves #340.

Ensures aircraft reach targeted speeds.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
@@ -1560,7 +1560,6 @@ export default class AircraftInstanceModel {
     updateSpeedPhysics() {
         let speedChange = 0;
         const differenceBetweenPresentAndTargetSpeeds = this.speed - this.target.speed;
-        const decelerationFactorDueToGroundBraking = 3.5;
 
         if (differenceBetweenPresentAndTargetSpeeds === 0) {
             return;
@@ -1570,7 +1569,7 @@ export default class AircraftInstanceModel {
             speedChange = -this.model.rate.decelerate * window.gameController.game_delta() / 2;
 
             if (this.isOnGround()) {
-                speedChange *= decelerationFactorDueToGroundBraking;
+                speedChange *= PERFORMANCE.DECELERATION_FACTOR_DUE_TO_GROUND_BRAKING;
             }
         } else if (this.speed < this.target.speed) {
             speedChange  = this.model.rate.accelerate * window.gameController.game_delta() / 2;
--- a/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
@@ -1576,10 +1576,10 @@ export default class AircraftInstanceModel {
             speedChange *= extrapolate_range_clamp(0, this.speed, this.model.speed.min, 2, 1);
         }
 
+        this.speed += speedChange;
+
         if (abs(speedChange) > abs(differenceBetweenPresentAndTargetSpeeds)) {
             this.speed = this.target.speed;
-        } else {
-            this.speed += speedChange;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1558,28 +1558,29 @@ export default class AircraftInstanceModel {
      * @method updateWarning
      */
     updateSpeedPhysics() {
-        let difference = null;
+        let speedChange = 0;
+        const differenceBetweenPresentAndTargetSpeeds = this.speed - this.target.speed;
+        const decelerationFactorDueToGroundBraking = 3.5;
```
n8rzz: Should this list be in a constants file?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1558,28 +1558,29 @@ export default class AircraftInstanceModel {
...
 
-        if (this.target.speed < this.speed - 0.01) {
-            difference = -this.model.rate.decelerate * window.gameController.game_delta() / 2;
+        if (differenceBetweenPresentAndTargetSpeeds === 0) {
```
n8rzz: So much cleaner 👍

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1558,28 +1558,29 @@ export default class AircraftInstanceModel {
...
-                // What is 3.5 is this restiance/breaking power?
-                difference *= 3.5;
+                speedChange *= decelerationFactorDueToGroundBraking;
             }
-        } else if (this.target.speed > this.speed + 0.01) {
-            difference  = this.model.rate.accelerate * window.gameController.game_delta() / 2;
-            difference *= extrapolate_range_clamp(0, this.speed, this.model.speed.min, 2, 1);
+        } else if (this.speed < this.target.speed) {
+            speedChange  = this.model.rate.accelerate * window.gameController.game_delta() / 2;
+            speedChange *= extrapolate_range_clamp(0, this.speed, this.model.speed.min, 2, 1);
```
n8rzz: What does the 1 and 2 mean?

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1558,28 +1558,29 @@ export default class AircraftInstanceModel {
...
-
-            if (abs(offset) < abs(difference)) {
-                this.speed = this.target.speed;
-            } else {
-                this.speed += difference;
-            }
+        if (abs(speedChange) > abs(differenceBetweenPresentAndTargetSpeeds)) {
+            this.speed = this.target.speed;
+        } else {
+            this.speed += speedChange;
```
n8rzz: Could we remove the else by moving this assignment above the if and then overriding it with the existing if?

```
this.speed += speedChange;

If (abs(speedChange) > abs(differenceBetweenPresentAndTargetSpeeds)) {
    this.speed = this.target.speed;
}

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: Eh, yeah. I was being lazy. :smile: 

Fixed.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: Honestly couldn't really make sense of it. Seems backwards, but obviously it's right. Some convoluted solution that's above my head at the moment.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: Sure, I guess.

Fixed.

PR #368, opened 2017-04-24
bugfix/364

### LINKED ISSUE #364: System commands with no arguments not working
For example, the command `version` should print the simulator version to the console.

Within the CommandParser, this fails at `#_buildSystemCommandModel)` when called at line 150. It seems to expect _either_ a callsign _or_ an argument. For example, `timewarp` takes an argument and works just fine, but those that have neither aren't handled correctly.

Additionally, if those commands that _expect_ arguments (like timewarp) don't receive them, they may not be working properly. `timewarp` alone results in the canvas going blank, which could either be an issue with the parser freaking out, or it could be working properly, and the canvas can't figure out what to do when time stops (in which case, we should enforce a default argument of `1` in the parser to prevent this).

n8rzz:
Resolves #364.

Repairs system commands that were not being parsed correctly so they can be used in-game.

### DIFF
--- a/src/assets/scripts/client/commandParser/argumentParsers.js
+++ b/src/assets/scripts/client/commandParser/argumentParsers.js
@@ -173,3 +173,25 @@ export const holdParser = (args) => {
 
     return [turnDirection, legLength, fixName];
 };
+
+/**
+ * the `timewarp` command needs to be able to provide a default value,
+ * this parser allows us to do that.
+ *
+ * @method timewarpParser
+ * @param  {array|undefined} [args=[]]
+ * @return {array<number>}
+ */
+export const timewarpParser = (args = []) => {
+    const defaultTimewarpValue = 0;
+
+    if (args.length === 0) {
+        return [defaultTimewarpValue];
+    }
+
+    // calling method is expecting an array with values that will get spread later, thus we purposly
+    // return an array here
+    return [
+        convertStringToNumber(args[0])
+    ];
+};
--- a/src/assets/scripts/client/commandParser/argumentParsers.js
+++ b/src/assets/scripts/client/commandParser/argumentParsers.js
@@ -183,7 +183,7 @@ export const holdParser = (args) => {
  * @return {array<number>}
  */
 export const timewarpParser = (args = []) => {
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/commandParser/argumentParsers.js
```diff
@@ -173,3 +173,25 @@ export const holdParser = (args) => {
...
+/**
+ * the `timewarp` command needs to be able to provide a default value,
+ * this parser allows us to do that.
+ *
+ * @method timewarpParser
+ * @param  {array|undefined} [args=[]]
+ * @return {array<number>}
+ */
+export const timewarpParser = (args = []) => {
+    const defaultTimewarpValue = 0;
```
erikquinn: Should be `1`.

src/assets/scripts/client/commandParser/argumentParsers.js
```diff
@@ -173,3 +173,25 @@ export const holdParser = (args) => {
...
+
+    if (args.length === 0) {
+        return [defaultTimewarpValue];
+    }
+
+    // calling method is expecting an array with values that will get spread later, thus we purposly
+    // return an array here
+    return [
+        convertStringToNumber(args[0])
```
erikquinn: Is this better than plain old `parseInt()` or `parseFloat()`?

n8rzz: well, this actually _is_ `parseInt`.  I've just wrapped inside a function name that provides a little more context on the intent.  I did this because this little guy gets used a lot in the `CommandParser` files and I didn't want to have to keep writing it.

src/assets/scripts/client/commandParser/argumentParsers.js
n8rzz: :facepalm:

n8rzz: fixed

src/assets/scripts/client/commandParser/argumentParsers.js
erikquinn: Seems goofy, but whatever floats your boat... ⛵

PR #377, opened 2017-04-25
bugfix/299

### LINKED ISSUE #299: Issues with STARs and SIDs at GCRR
Hi, there.  Got a few issues with STAR and SID setup in GCRR.  Playing version 4.1.2 if it matters, but I don't think so.  

1. I  had two planes collide that were coming in on the same STAR, TERT6P.GCRR, RYR495 & AEA235.   See screen shot below.  Pretty sure it was because the AEA flight was an AT43 aircraft with a max speed of 245, and the STAR is set up so that they are supposed to speed up to speed 450 at the TERTO fix (even if they enter the airspace at speed 250), so the faster plane just caught up to and hit the slower one.  Obviously two planes on the same STAR should never collide, so I wasn't even watching that part of route (I had to zoom out to even see where it occurred).

2.  There is a potential for a collision between departing flights following the KORAL SID, and arriving aircraft on the TERT6P STAR.  Both are set up to cross the POKAB fix at FL 150, and due to timing I often get near misses and near collisions.   Work around is to just clear the KORAL departure to FL 40 until passing that fix, but ideally a SID and STAR should never have a conflict (right?)

3. As you can see in the screen shot, there are a lot of errors due to missing exit points for the SIDS.  Not sure if this is by design or a bug?

Thanks,
Scott

image

indianbhaji:
Resolves #299.

Updates to GCRR (Lanzarote).

- [x] Rework SIDs
- [x] Check arrivals
- [x] Add entryPoints/exitPoints
- [x] Improve spawnPatterns


indianbhaji:
@erikquinn 
Traffic in-game: 11.5 acph
Real world: 6 acph
Reference: http://www.aena.es/csee/ccurl/825/352/Estadisticas_2016.pdf

assets/airports/gcrr.json
```diff
@@ -23,297 +23,526 @@
...
-                ["N29.0011333", "W13.7064167"]
+                ["N27d46m00.00", "W015d33m00.00"],
+                ["N27d46m00.00", "W015d36m00.00"],
+                ["N27d43m00.00", "W015d36m00.00"],
+                ["N27d43m00.00", "W015d33m00.00"],
+                ["N27d46m00.00", "W015d33m00.00"]
             ]
         }
     ],
     "fixes": {
```
erikquinn: There's a few missing fixes, if you wouldn't mind adding those in to clear the warnings. :smiley: 

image

indianbhaji:
wilco, not completely done yet

erikquinn:
@indianbhaji roger, just re-request review when ready :smile:

indianbhaji:
Closing in favour of @willfrd's work in progress update of GCRR!

erikquinn:
@indianbhaji I can't find "@willfrd's work in progress update of GCRR". If it exists, we can delete this old branch though?

PR #384, opened 2017-04-28
feature/317

### LINKED ISSUE #317: Update color set to match openscope
With the move to openscope, we need to get a fresh set of paint on the scope colors. See the `openscope-legacy` repository and #294 to get an idea of what we're looking to do with the colors.

For now, we can simply replace the colors in use. In the future (or for super-mega bonus points), we can create different color profiles, able to be changed by the user through the settings menu.

erikquinn:
Resolves #317.

Map all colors to "themes" so they can easily be changed, multiple themes added, etc.


src/assets/scripts/client/constants/colors/terrainThemes.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * @property TERRAIN_COLOR_DEFAULT
+ * @type {Object}
+ * @final
+ */
+const TERRAIN_COLOR_DEFAULT = {
```
n8rzz: the properties here could probably just be defined in the main object

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -882,17 +831,23 @@ export default class ConvasController {
         const alerts = aircraft.hasAlerts();
 
         if (!aircraft.inside_ctr) {
-            cc.fillStyle = COLORS.LIGHT_SILVER_03;
```
n8rzz: I think these are/were for drawing a different color dot (aircraft) based on the various states.  I don't know if any of those are working, though.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1556,10 +1523,10 @@ k
 
         $.each(airport.terrain || [], (elevation, terrainLevel) => {
             max_elevation = Math.max(max_elevation, elevation);
-            const color = `rgba(${prop.ui.terrain.colors[elevation]}, `;
+            const color = `rgba(${prop.ui.terrain.COLOR[elevation]}, `;
```
n8rzz: why is the terrain color stuff coming from the `uiController`?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1663,7 +1630,8 @@ k
             cc.fillStyle = 'transparent';
             this.canvas_draw_poly(cc, area.coordinates);
 
-            cc.fillStyle = COLORS.RESTRICTED_AIRSPACE_FILL;
+            // FIXME: Is the restricted airspace EVER filled???
+            cc.fillStyle = this.theme.RESTRICTED_AIRSPACE_FILL;
```
n8rzz: not to my knowledge.  its just a line drawn for the restricted area poly

src/assets/scripts/client/UiController.js
```diff
@@ -76,11 +63,8 @@ export default class UiController {
         this.ui.scale_max = 80; // max scale
         this.ui.scale_min = 1; // min scale
         this.ui.scale = this.ui.scale_default;
-        this.ui.terrain = {
-            colors: TERRAIN_COLORS,
-            border_opacity: 1,
-            fill_opacity: 0.1
-        };
+        // TODO: This belongs in the CanvasController, not UiController
+        this.ui.terrain = THEME.DEFAULT.TERRAIN;
```
n8rzz: were you going to move this?  or save it for another issue to be done later?

src/assets/scripts/client/constants/colors/terrainThemes.js
erikquinn: Do you mean just the `BORDER_OPACITY` and `FILL_OPACITY`? the theme itself is probably wise to keep separate, to make it easy to add different terrain coloring themes and treat them as a unit that belongs to an overall "theme" used by the canvas.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: It's not my preference either, but it can be moved to the appropriate place at a later date.

src/assets/scripts/client/UiController.js
erikquinn: the latter, i'll add the issue now.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Upon further review, they are always overwritten by other assignments. Removed.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: It's not "filled", but apparently that's what `context.fillText()` uses for coloring. Restored, fixed bad reference. All working better now. :+1:

PR #369, opened 2017-04-24
feature/93

erikquinn:
Resolves #93.

Issue #93, opened 2017-01-17
RunwayModel cleaning and improvements


erikquinn:
_From @n8rzz on December 10, 2016 5:0_

- [x] Remove circular reference of `AirportModel.runway.airportModel`
- [ ] ~~add `_get` to `parse` method~~
- [x] investigate current implementations and optimize
- [x] investigate pros/cons of implementing a RunwayCollection architecture
  - see: `AircraftConflict.checkProximity() #runwayRelationship`
  - see: `AirportModel.rwy.metadata`
- [x] replace `airport` param with `airport.position`
- [x] get to 90%+ coverage
- [x] streamline queue methods ~~which may necessitate adding `_uniqueId()` to the `AircraftInstanceModel`~~
- [x] abstract:`Aircraft._calculateTargetedHeadingToInterceptCourse()` logic

_Copied from original issue: n8rzz/atc#182_

PR #386, opened 2017-04-28
feature/374

### LINKED ISSUE #374: Allow for specification of airport's default arrival and departure runway
In the airport file, add (and require) a key that specifies the default departure runway and default arrival runway. This would eventually become a bad idea (with wind shifts and all), but since the winds are static for each airport (ughhh...), we might as well allow for specification of runway configurations to this limited degree to increase the similarity to the real-world operation of each airport.

---

Related: #34.

erikquinn:
Resolves #374.

Allow for specification of airport's default arrival and departure runway.


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -447,7 +447,7 @@ export default class AircraftCommander {
 
         // Set the runway to taxi to
         if (!taxiDestination) {
-            taxiDestination = this._airportController.airport_get().runway;
+            taxiDestination = this._airportController.airport_get().departureRunway.name;
```
n8rzz: I would much rather see:
```javascript
const airport = this._airportController.airport_get();
taxiDestination = airport.departureRunway.name;
``` 
It makes the intent a little clearer and eventually we would just store the current airport in the constructor so we don't have to use `airport_get()` all the time.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -459,51 +473,6 @@ export default class AirportModel {
         return wind;
     }
 
-    /**
-     * @for AirportModel
-     * @method updateRunway
-     */
-    updateRunway(length = 0) {
```
n8rzz: we should keep this in, even if we're not using it now.  in the future this will allow us to change the current wind and update the default runway accordingly.  additionally, I think I'm doing some work with this in #93 ( #369 )

n8rzz:
take a look at the tests on this one, too.  they are very unhappy with this branch at the moment. ☹️

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: Fixed.

src/assets/scripts/client/airport/AirportModel.js
erikquinn: Fixed.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -459,49 +512,33 @@ export default class AirportModel {
         return wind;
     }
 
+    // TODO: Implement changing winds, then bring this method back to life
```
n8rzz: in #93 this will actually move to the `RunwayCollection`

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -459,49 +512,33 @@ export default class AirportModel {
...
-        this.timeout.runway = window.gameController.game_timeout(this.updateRunway, Math.random() * 30, this);
-    }
-
-    /**
-     * @for AirportModel
-     * @method selectRunway
-     */
-    selectRunway() {
-        return this.runway;
+        // // TODO: this method contains some ambiguous names. need better names.
```
n8rzz: same with this stuff too, all in the `RunwayCollection`

PR #398, opened 2017-04-30
bugfix/393

### LINKED ISSUE #393: EDDT: Arrivals climb without bound
Inbound aircraft climb to `Infinity`. Two problems with that:
- they shouldn't be trying
- they shouldn't be able (performance)

erikquinn:
Resolves #393.

Prevent Aircraft from setting mcp altitude to infinity on spawn.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
@@ -901,6 +901,18 @@ export default class AircraftInstanceModel {
         if (this.altitude < 10000) {
             this.target.speed = Math.min(this.target.speed, AIRPORT_CONSTANTS.MAX_SPEED_BELOW_10K_FEET);
         }
+
+        if (this.target.altitude > this.model.ceiling) {
+            this.target.altitude = this.model.ceiling;
+        }
+
+        if (this.target.speed > this.model.speed.max) {
+            this.target.speed = this.model.speed.max;
+        }
+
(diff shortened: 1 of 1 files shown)


src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -185,7 +185,11 @@ export default class AircraftInstanceModel {
 
         // TODO: There are better ways to ensure the autopilot is on for aircraft spawning inflight...
         if (options.category === FLIGHT_CATEGORY.ARRIVAL) {
-            const bottomAltitude = this.fms.getBottomAltitude();
+            let bottomAltitude = this.fms.getBottomAltitude();
+
+            if (bottomAltitude === Infinity) {
```
n8rzz: this logic should happen in `fms.getBottomAltitude()`

erikquinn: I agree, but we don't have access to the aircraft's altitude in that scope. Unless you think we should just pass that? :grimacing:

n8rzz: no, you're not thinking 4th dimensionally

erikquinn:
@n8rzz Right, but in the case that `minAltitudeFromLegs.length === 0`, we cannot return `0`, we must return the current altitude.

n8rzz:
@erikquinn ahhhh, I smell what you're stepping in there.  I'll allow it 😉

PR #409, opened 2017-05-01
bugfix/408

### LINKED ISSUE #408: No response to improper single-dot fix-to-fix reroute command
Entering a command of `rr FIXXA.FIXXB` instead of the proper `rr FIXXA..FIXXB` results in `Fms.isValidRoute()` throwing. This improper route should receive a reply that the route was not entered properly, and in red.

erikquinn:
Resolves #408.

Ensure red response is given to improper reroute command.

### DIFF
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -799,6 +799,10 @@ export default class Fms {
     isValidRoute(routeString, runway = '') {
         const routeSegments = routeStringFormatHelper(routeString);
 
+        if (!routeSegments) {
+            return false;
+        }
+
         for (let i = 0; i < routeSegments.length; i++) {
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -799,6 +799,10 @@ export default class Fms {
     isValidRoute(routeString, runway = '') {
         const routeSegments = routeStringFormatHelper(routeString);
 
+        if (!routeSegments) {
```
n8rzz: I'd love veto see a test added for this case

erikquinn: A fair point! Added :smile:

n8rzz: So pretty!  :shipit:

PR #422, opened 2017-05-05
feature/33

### LINKED ISSUE #33: Actually USE the suffixes of procedures
Given a procedure `ANKER`, users should be able to enter `SID ANKER1Q` and/or `STAR ANKER1Q`, depending on the type of procedure. Such was the intent of the `suffix` property, but it was never really fully hooked up, and airport creators have worked around it by creating individual procedures `PIK27X`, `PIK27D`, `PIK23X`, `PIK23D`, `PIK21X`, `PIK21D`, etc. Completing the functionality of the suffixes would eliminate the need for them to do this.

Originally reported in zlsa/atc#655.

n8rzz:
Resolves #33.

Adds support for procedure suffixes.

pay close attention to the tests here.  It'll make reasoning about the code a lot easier because most of the changes here are a few layers removed from where the magic starts.


test/navigationLibrary/StandardRoute/StandardRouteModel.spec.js
```diff
@@ -16,6 +16,7 @@ import { FIX_LIST_MOCK } from '../Fix/_mocks/fixMocks';
 
 import {
     STAR_LIST_MOCK,
+    STAR_WITH_SIFFIX,
```
ij1: Is this use of S_I_FFIX intentional?

Looks genuine typo to me but of course I'm not native (nor that familiar with aviation) so it might be something I just don't know.

n8rzz: LOL.  no, thats definitely a type.  good catch!

n8rzz: done

n8rzz:
**BUGS** 
- [ ] When entering `sid OBAXA5S`on EDMM I get >SID name not understood
- [ ] So if you do `sid GIVMI1N` at Munich it displays GIVMI1N.GIVMI  but if you do `caf` it just shows GIVMI.GIVMI?

**TODO**
- [ ] update runway selection and verify route

panther2:
Concerning

> When entering sid OBAXA5Son EDMM I get >SID name not understood

This occurred because I entered `sid OBAX5S` as I am used to a maximum of 4 letters + suffix - that is how the AIRAC datasets are structured. So I never had the idea to enter 5 letters plus suffix.

n8rzz:
@erikquinn @indianbhaji @Fechulo please review, break and provide feedback when you can.  this is ready to go (changelog coming later)

n8rzz:
@erikquinn bump

n8rzz:
@erikquinn bump

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -369,16 +376,67 @@ export default class AircraftCommander {
...
-        const { name: airportName } = this._airportController.airport_get();
+        // TODO: why are we passing this if we already have it?
+        const runwayModel = aircraft.fms.arrivalRunway;
+        const airportModel = this._airportController.airport_get();
+
+        if (this._navigationLibrary.isSuffixRoute(routeString, PROCEDURE_TYPE.STAR)) {
+            return this._runSTARforSuffix(aircraft, airportModel, routeString);
+        }
+
+        return aircraft.pilot.applyArrivalProcedure(routeString, runwayModel, airportModel.name);
```
erikquinn: So if the string _includes_ a suffix, the appropriate procedure will be flown. If a suffix is _not specified_, the procedure will be run for the runway designated as active.

EXACTLY what I was looking for. And it worked perfectly in my trials. :+1:

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -116,21 +116,11 @@ export default class Fms {
...
-        // TODO: Use this!
-        /**
-         * Airport the aircraft arrives to
-         *
-         * @for Fms
-         * @property arrivalAirport
-         * @type {AirportModel}
-         */
-        this.arrivalAirport = null;
-
```
n8rzz: Including the airport model, up to two times, for each aircraft makes me very nervous. I would prefer to keep the fms as slim as practical. How do you plan to use them?

I worry about circular references and just unnecessary code.

erikquinn: Basically to remove all `airport_get()`, and the need for an aircraft or fms to have a `this._airportController` (if they even do). It also serves to separate the arrival/departure fields from the rest of the route, so we can allow/disallow players to change the destination airport of aircraft. Basically the same way we use the `arrivalRunway` and `departureRunway` we would do for the airports (especially useful when this is plural :wink:). And I agree on the circularity; I suppose the runway queue would be an example (though now that is `_id` only :+1:) of how we could end up with a problem there.

n8rzz: in this case here you have identified a problem; we are currently storing `RunwayModels`, and you would like to store `AirportModels` as well.  The airports themselves contain the `RunwayModels`.  

I agree in that we should try to remove the reliance on `airport_get()`, but I think there is a better way.  Storing the `AirportModel` (which is huge) times two for arrival and departure seems very heavy handed to me.  I would really like to see what you have planned with this, perhaps there is a better way?  Perhaps we add a facade to the `NavigationLibrary` or something?

erikquinn: I was under the impression it wouldn't be "storing" the airport multiple times, but only pointers. But perhaps putting the airports in the navigation library as you suggest makes more sense, then the fms can access the information it will need.

PR #426, opened 2017-05-06
bugfix/424

### LINKED ISSUE #424: altitude aa leads to NaN
Issuing "a aa" to an aircraft results in "maintain NaN" and NaN to the flight's strip.

ij1:
Resolves #424.

Fixes NaN being passed on from command parser for invalid altitudes.

### DIFF
--- a/src/assets/scripts/client/commandParser/argumentValidators.js
+++ b/src/assets/scripts/client/commandParser/argumentValidators.js
@@ -104,6 +104,12 @@ export const altitudeValidator = (args = []) => {
         return hasLengthError;
     }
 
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/commandParser/argumentValidators.js
```diff
@@ -104,6 +104,12 @@ export const altitudeValidator = (args = []) => {
         return hasLengthError;
     }
 
+    let numberFromString = convertStringToNumber(args[0]);
```
n8rzz: This can actually be a `const` instead of a `let`

PR #428, opened 2017-05-06
feature/372

### LINKED ISSUE #372: Add squawk command
add new command `squawk` that instructs an aircraft to set their transponder code to a specific numeric value.

this feels like an easy task to knock out, we're going to need these in the future anyway.  this will be a good opportunity for somebody else to add a new command.

---

acceptance criteria:
- [ ] `squawk` and `sq` command definitions
- [ ] accepts a single arg, as a number with exactly 4 digits
- [ ] sets `AircraftInstanceModel#transponderCode` property
- [ ] full test coverage of `CommandParser` additions

ij1:
Resolves #372.

Add squawk command.

### DIFF
--- a/src/assets/scripts/client/commandParser/argumentValidators.js
+++ b/src/assets/scripts/client/commandParser/argumentValidators.js
@@ -248,7 +248,7 @@ export const squawkValidator = (args = []) => {
         return ERROR_MESSAGE.SINGLE_ARG_LENGTH;
     }
 
-    if (REGEX.TRANSPONDER_CODE.exec(args[0]) == null) {
+    if (!REGEX.TRANSPONDER_CODE.test(args[0])) {
         return ERROR_MESSAGE.INVALID_SQUAWK;
     }
 };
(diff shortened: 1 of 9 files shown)


src/assets/scripts/client/commandParser/argumentValidators.js
```diff
@@ -228,3 +229,20 @@ export const holdValidator = (args = []) => {
...
+ * @function squawkValidator
+ * @param args {array}
+ * @return {string|undefined}
+ */
+export const squawkValidator = (args = []) => {
+    if (args.length !== 1) {
+        return ERROR_MESSAGE.SINGLE_ARG_LENGTH;
+    }
+
+    if (REGEX.TRANSPONDER_CODE.exec(args[0]) == null) {
```
n8rzz: Two things here:
- why not use `.test()` instead of `exec()`?
- please use strict equals `===` for comparisons

ij1: Those are just because of my general unfamiliarity with JS quirks/libs (which is btw good to keep an eye on if you review any JS from me :-)). I'll fix those.

PR #432, opened 2017-05-08
feature/19

### LINKED ISSUE #19: Fly-over vs fly-by fixes
Allow for differentiation between fixes that must be flown over before beginning turn or not, as there are some instances where it can make a difference.

image

Originally reported in zlsa/atc#427.

erikquinn:
Resolves #19.

Add capability for fly-over fixes in route strings.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftInstanceModel.js
@@ -1314,11 +1314,11 @@ export default class AircraftInstanceModel {
         const isTimeToStartTurning = distanceToWaypoint < nm(calculateTurnInitiaionDistance(this, waypointPosition));
         const closeToBeingOverFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_PASS_WAYPOINT_NM;
         const closeEnoughToFlyByFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_FLY_BY_WAYPOINT_NM;
-        const wantToFlyByFix = closeEnoughToFlyByFix && isTimeToStartTurning;
+        const shouldFlyByFix = closeEnoughToFlyByFix && isTimeToStartTurning;
         let shouldMoveToNextFix = closeToBeingOverFix;
 
         if (!this.fms.currentWaypoint.isFlyOverWaypoint) {
-            shouldMoveToNextFix = closeToBeingOverFix || wantToFlyByFix;
+            shouldMoveToNextFix = closeToBeingOverFix || shouldFlyByFix;
         }
 
         if (shouldMoveToNextFix) {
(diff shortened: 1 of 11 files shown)


src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -1,25 +1,13 @@
 import _get from 'lodash/get';
-import { VECTOR_SEGMENT_SYMBOL } from '../../constants/navigation/routeConstants';
+import {
+    RNAV_WAYPOINT_DISPLAY_NAME,
+    RNAV_WAYPOINT_PREFIX,
```
n8rzz: Mmm, constants where constants belong. Very nice!

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -134,6 +133,8 @@ export default class WaypointModel {
          */
         this.isHold = false;
 
+        // TODO: This should be moved to the StandardRouteWaypointModel
```
n8rzz: is this something that should be done in this PR?  Or should there be another issue for this 😉

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
```diff
@@ -1314,8 +1314,12 @@ export default class AircraftInstanceModel {
         const isTimeToStartTurning = distanceToWaypoint < nm(calculateTurnInitiaionDistance(this, waypointPosition));
         const closeToBeingOverFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_PASS_WAYPOINT_NM;
         const closeEnoughToFlyByFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_FLY_BY_WAYPOINT_NM;
-        // TODO: abstract this logic to helper method
-        const shouldMoveToNextFix = closeToBeingOverFix || (closeEnoughToFlyByFix && isTimeToStartTurning);
+        const wantToFlyByFix = closeEnoughToFlyByFix && isTimeToStartTurning;
```
n8rzz: this should be renamed to `shouldFlyByFix`

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
erikquinn: I thought the WaypointModel _extended_ the StandardRouteWaypointModel, since apparently the latter is created first, from which the former is generated... but that's by passing it's own properties to the constructor (which seems fishy?). Thus, this todo is invalid, as the property _is_ in fact needed in both objects. TODO removed.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: :thinking: Now that you've got me thinking of it, that's not ideal either, because whether or not the aircraft "_should_" fly by the fix is actually determined by whether or not the fix is a fly-by fix (whether they should is essentially the question of the if statement).

What about `isTimeToFlyByFix`?

n8rzz: well, in this sense, _should_ is used as in 'this should happen now'.  So yeah, I still prefer `shouldFlyByFix`.  

but you're so much better with words than I am 😉

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
n8rzz: yeah, I'm not a huge fan of that pattern.  it works as a way to get the same object from two different classes.  I'm not sold on the idea that inheritance is the correct pattern here, though, either.

src/assets/scripts/client/aircraft/AircraftInstanceModel.js
erikquinn: Changed to `shouldFlyByFix`.

Issue #286, opened 2017-03-07
ModelSourcePool.releaseReusable needs an early return


n8rzz:
```javascript
releaseReusable(constructorName, ...args) {
    let model = this._findModelByConstructorName(constructorName);

    if (!model) {
        model = new CLASS_MAP[constructorName](...args);

        // should be
        // return new CLASS_MAP[constructorName](...args);
    }

    model.init(...args);

    return model;
}
```

n8rzz:
not needed. closing

PR #436, opened 2017-05-08
bugfix/286

Dabea:
Completes #286 

Adds an early return for ModelSourcePool.releaseReusable.

I see why one might not have been added here since you need to call `model.init(...args);` either way. But it dose give the advantage of spiking of creating a new constructor object.

### DIFF
--- a/src/assets/scripts/client/base/ModelSource/ModelSourcePool.js
+++ b/src/assets/scripts/client/base/ModelSource/ModelSourcePool.js
@@ -91,9 +91,13 @@ class ModelSourcePool extends BaseCollection {
     releaseReusable(constructorName, ...args) {
         let model = this._findModelByConstructorName(constructorName);
 
-        if (!model) {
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/base/ModelSource/ModelSourcePool.js
```diff
@@ -91,9 +91,13 @@ class ModelSourcePool extends BaseCollection {
     releaseReusable(constructorName, ...args) {
         let model = this._findModelByConstructorName(constructorName);
 
-        if (!model) {
-            model = new CLASS_MAP[constructorName](...args);
+        if(model){
+            model.init(...args);
+            return model;
```
n8rzz: I'd like to see the logic reversed here.  The code inside the if block should happen only when `!model`.  pay close attention to spacing and line breaks too.  There should be a space after the word `if` and an empty line above `return`.

n8rzz:
closing as it is not needed

Issue #434, opened 2017-05-08
add support for non-procedural departures


n8rzz:
support departures for aircraft with non-procedural routeStrings (`FIXXA..FIXXB..FIXXC`)

this may already exist and needs only some tweaking.  this may not exist at all and may require a few days to get connected.

PR #445, opened 2017-05-09
Check waipoint for null.

calin-pere:


n8rzz:
@calin-pere I appreciate the attempt here, however, I believe that the `null` waypoint is indicative of a larger issue (mostly #434).

1) Firstly, if you would like to contribute, we can add you to the organization. that way you can push directly to the repo *_and_* our CI can run on your code.
1) pull requests should be titled after the branch name and our branches are named after issues (bugfix/XXX or feature/XXX).  Every PR should have a corresponding issue associated with it.
1) The `currentWaypoint` property is a getter in the `LegModel`, so if we want to default it to `null` thats where that change should live (and there should be new tests added for that case)
1) Our first assumption is that an aircraft will almost always have at least one waypoint or is initially assigned two or more waypoints.
1) In the cases where it does not have a waypoint, it is assigned the last heading it was on as it past the last waypoint.

Seeing `currentWaypoint` come up as `null` is cause for concern that there are other underlying issues causing that to happen.  

I'm going to close this but I do want to keep the conversation open.  So feel free to ask questions or call me names 😉 

If you would like to make an attempt at solving this, take a look at #434 (and ask more questions).  You can also join the conversation on our [slack channel](http://slack.openscope.co), where there is usually someone available to field questions and help with stuff.

calin-pere:
So, I did some debugging yesterday, and it seems to be an issue with the airport file. I was looking at the egnm.json and it has this `"route": "EGNM.NELSA.NELSA"` spawning pattern but it doesn't have that sid for rwy 14. So maybe extra verification when loading airports ?

Also , I've discovered another issue on that airport file , the POL sid : the POL waypoint is reached before airspace exit is checked so the mode control is set to HOLD determining a bad exit from airspace.

Issue #325, opened 2017-04-16
Review airports for mistaken "N" / "E" / "S" / "W" in coordinates


erikquinn:
OTHH is breaking because Runway 34R has a position of `["N25d15m19.65", "N051d37m36.41", "13.10ft"]`. Mistakes like these may have failed silently before, but will now throw errors and crash the game (this is by design, because we can see and fix the issues rather than stumble across issues that are hard to trace back to this simple error).

All airports should be briefly reviewed (loading, and a quick look through the fixes) to see if there are any obvious "I put `S` but really meant `N`" situations.

erikquinn:
Resolved by #450. Closing.

Issue #323, opened 2017-04-15
Singapore: Incorrect coordinates for MASBO fix


indianbhaji:
Should be N for the first value

erikquinn:
This seems to have been fixed. Closing as no longer valid.

Issue #246, opened 2017-02-19
Airport Revival: Abu Dhabi International Airport


indianbhaji:
Abu Dhabi International Airport (Arabic: مطار أبو ظبي الدولي‎‎) (IATA: AUH, ICAO: OMAA) is an international airport in the Emirate of Abu Dhabi, the capital of the United Arab Emirates.

The airport, located 16.5 nautical miles (30.6 km; 19.0 mi) east of Abu Dhabi city, is the second largest in the UAE, serving around 20 million passengers in 2014. It has three operational passenger terminals—Terminal 1 (divided into Terminals 1A and 1B), Terminal 2, Terminal 3. Abu Dhabi International Airport is spread over an area of 3,400 hectares (8,500 acres). Its terminal spaces are dominated by Etihad Airways, which is the United Arab Emirates' second largest air carrier after Emirates. More than 30 airlines offered service to over 120 destinations in more than 60 countries with non-stop services to all inhabited continents.

This airport was omitted when spawnPatterns were introduced.

PR #450, opened 2017-05-09
bugfix/325

ij1:
Should resolve all cases for #325 except what is mentioned below.

I ran quite many kind of sed tricks to detect various coordinate issues. Besides what is included here I came across these:
- OMAA (should get covered by #246).
- MASBO (#323, I thought there was PR for this already but I cannot find it now. If need be, I can either include it here or do another PR for it?).
- EKCH that is covered by #448.

For the rest of the airports, the lat and long amount now agree perfectly. However, this kind of checking only captures gross errors and correctness is not ensured for those airports that cross 0 degrees (mainly concern with EGxx airports) as there could still be invalid coordinates that would belong to the opposite hemisphere but there was too much noise to find them.


assets/airports/osdi.json
```diff
@@ -51,7 +51,7 @@
         {
             "name": ["05R", "23L"],
             "name_offset": [[0, 0], [0, 0]],
-            "end": [["N33d23m06.00", "E36d29m46.00"], ["N33d24m22.00", "N36d31m29.00"]],
+            "end": [["N33d23m06.00", "E36d29m46.00"], ["N33d24m22.00", "E36d31m29.00"]],
             "length": 3.600,
```
indianbhaji: Whilst you're at it, could you maybe remove `length`?

assets/airports/othh.json
```diff
@@ -172,23 +172,23 @@
...
+        "__34L":  ["N25d15m19.32", "E051d36m19.56"]
     },
 
     "runways": [
         {
             "name": ["16L", "34R"],
             "name_offset": [[0, 0], [0, 0]],
-            "end": [["N25d17m45.97", "E051d36m31.96", "13.21ft"], ["N25d15m19.65", "N051d37m36.41", "13.10ft"]],
+            "end": [["N25d17m45.97", "E051d36m31.96", "13.21ft"], ["N25d15m19.65", "E051d37m36.41", "13.10ft"]],
             "length": 4.850,
```
indianbhaji: Whilst you're at it, could you maybe remove `length`?

assets/airports/othh.json
```diff
@@ -172,23 +172,23 @@
...
             "ils": [true, true]
         },
         {
             "name": ["16R", "34L"],
             "name_offset": [[0, 0], [0, 0]],
-            "end": [["N25d17m27.52", "E051d35m23.07", "13.13ft"], ["N25d15m19.32", "N051d36m19.56", "13.22ft"]],
+            "end": [["N25d17m27.52", "E051d35m23.07", "13.13ft"], ["N25d15m19.32", "E051d36m19.56", "13.22ft"]],
             "length": 4.250,
```
indianbhaji: Whilst you're at it, could you maybe remove `length`?

PR #468, opened 2017-05-18
bugfix/467

### LINKED ISSUE #467: StandardRouteModel throws due to vector waypoints lack of a positionModel
`StandardRouteModel._updateWaypointsWithPreviousWaypointData()` crashes when trying to calculate distance between waypoint lists that include vector waypoints (eg `"#320"`) because those waypoints do not have a position.

erikquinn:
Resolves #467.

Prevent attempts to access vector waypoint positions.

### DIFF
--- a/src/assets/scripts/client/math/flightMath.js
+++ b/src/assets/scripts/client/math/flightMath.js
@@ -203,7 +203,7 @@ export const calculateTurnInitiaionDistance = (aircraft, currentWaypointPosition
     const nominalBankAngleDegrees = 25;
     const speed = kn_ms(aircraft.speed);
     const bankAngle = degreesToRadians(nominalBankAngleDegrees);
-    const nextWaypointModel = aircraft.fms.currentLeg.nextWaypoint;
+    const nextWaypointModel = aircraft.fms.getNextWaypointModel();
 
     if (!aircraft.fms.hasNextWaypoint() || nextWaypointModel.isVector) {
         return 0;
--- a/src/assets/scripts/client/math/flightMath.js
+++ b/src/assets/scripts/client/math/flightMath.js
@@ -203,8 +203,9 @@ export const calculateTurnInitiaionDistance = (aircraft, currentWaypointPosition
     const nominalBankAngleDegrees = 25;
     const speed = kn_ms(aircraft.speed);
     const bankAngle = degreesToRadians(nominalBankAngleDegrees);
+    const nextWaypointModel = aircraft.fms.currentLeg.nextWaypoint;
 
-    if (!aircraft.fms.hasNextWaypoint()) {
+    if (!aircraft.fms.hasNextWaypoint() || nextWaypointModel.isVector) {
         return 0;
     }
 
@@ -213,7 +214,7 @@ export const calculateTurnInitiaionDistance = (aircraft, currentWaypointPosition
     }
 
     const nominalNewCourse = _calculateNominalNewCourse(
-        aircraft.fms.getNextWaypointPositionModel(),
+        nextWaypointModel.positionModel,
         currentWaypointPosition
     );
     const courseChange = _calculateCourseChangeInRadians(currentHeading, nominalNewCourse);
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
@@ -537,15 +537,13 @@ export default class StandardRouteModel extends BaseModel {
      *
      * @for StandardRouteModel
      * @method _updateWaypointsWithPreviousWaypointData
-     * @parma waypointModelList {array<StandardRouteWaypointModel>}
+     * @param waypointModelList {array<StandardRouteWaypointModel>}
      * @private
      */
     _updateWaypointsWithPreviousWaypointData(waypointModelList) {
-        _forEach(waypointModelList, (waypoint, i) => {
-            let previousWaypoint = waypointModelList[i - 1];
-            if (i === 0) {
-                previousWaypoint = waypoint;
-            }
+        for (let i = 1; i < waypointModelList.length; i++) {
+            const waypoint = waypointModelList[i];
+            const previousWaypoint = waypointModelList[i - 1];
 
             let distance = 0;
             if (!waypoint.isVector && !previousWaypoint.isVector) {
@@ -557,6 +555,6 @@ export default class StandardRouteModel extends BaseModel {
 
             waypoint.distanceFromPreviousWaypoint = distance;
             waypoint.previousStandardWaypointName = previousWaypoint.name;
-        });
+        }
     }
 }
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
@@ -547,10 +547,14 @@ export default class StandardRouteModel extends BaseModel {
                 previousWaypoint = waypoint;
             }
 
-            const distance = this.calculateDistanceBetweenWaypoints(
-                waypoint.relativePosition,
-                previousWaypoint.relativePosition
-            );
+            let distance = 0;
+            if (!waypoint.isVector && !previousWaypoint.isVector) {
+                distance = this.calculateDistanceBetweenWaypoints(
+                    waypoint.relativePosition,
+                    previousWaypoint.relativePosition
+                );
+            }
+
             waypoint.distanceFromPreviousWaypoint = distance;
             waypoint.previousStandardWaypointName = previousWaypoint.name;
         });
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
@@ -541,11 +541,13 @@ export default class StandardRouteModel extends BaseModel {
      * @private
      */
     _updateWaypointsWithPreviousWaypointData(waypointModelList) {
+        // Starting iterations at i=1 because we need two waypoints in order to
+        // ensure that there is a 'previous' waypoint from which we can measure.
         for (let i = 1; i < waypointModelList.length; i++) {
             const waypoint = waypointModelList[i];
             const previousWaypoint = waypointModelList[i - 1];
-
             let distance = 0;
+
             if (!waypoint.isVector && !previousWaypoint.isVector) {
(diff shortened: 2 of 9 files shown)


src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -94,8 +94,13 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
 
     for (let i = 0; i < waypointModelList.length; i++) {
         const waypoint = waypointModelList[i];
-        const waypointPosition = waypoint.relativePosition;
         let previousWaypoint = waypoint;
+
+        if (waypoint.isVector || previousWaypoint.isVector) {
```
ij1: I don't know if this PR is "ready enough" for this comment to be of any use but isn't the waypoint and previousWaypoint same here?

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -94,8 +94,13 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
...
+            continue;
+        }
+
+        const waypointPosition = waypoint.relativePosition;
         let previousPosition = waypoint.relativePosition;
 
         if (i > 0) {
```
ij1: There's also an access to [i-1].relativePosition in this block and no check that is going to prevent its access.

n8rzz:
Also, changelog 😉

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
erikquinn: Yeah, that does seem fishy. Cleaned up the method a bit (though it could still use some attention. Added a `TODO` for the use of `continue`).

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
erikquinn: Fixed with the rearrangement.

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -92,25 +92,31 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
...
-        let previousWaypoint = waypoint;
-        let previousPosition = waypoint.relativePosition;
+        const previousWaypoint = waypointModelList[i - 1];
 
-        if (i > 0) {
-            previousWaypoint = waypointModelList[i - 1];
-            previousPosition = previousWaypoint.relativePosition;
+        // TODO: Refactor to avoid continue statement
+        if (waypoint.isVector || previousWaypoint.isVector) {
+            continue;
```
n8rzz: @erikquinn I'm really not a fan of the `no-continue` rule.  In fact, I think I changed it to a warning in one of my branches 🤔 

I understand the sprit of the rule.  My thoughts are more like this, though:
```javascript
// rule suggests
for (let i = 0; i< length; i++) {
  if (condition) {
    // do stuff
  }
}

// I much prefer
for (let i = 0; i< length; i++) {
  if (!condition) {
      continue;
  }

  // do stuff
}

```

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -92,25 +92,31 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
...
-        if (isWithinAirspace(airport, waypointPosition) && i > 0) {
+        const waypointPosition = waypoint.relativePosition;
+        const previousPosition = waypoint.relativePosition;
+
+        if (isWithinAirspace(airport, waypointPosition)) {
             distanceFromClosestFixToAirspaceBoundary = nm(calculateDistanceToBoundary(airport, previousPosition));
+            totalDistance += distanceFromClosestFixToAirspaceBoundary;
 
-            continue;
+            break;
```
n8rzz: do you mean to `break` here?  the previous code would simply continue the loop, this is going to stop the loop and continue down the line.

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
erikquinn: @n8rzz You know me, I'll do what I'm told. :laughing:

Linter said no, but if you want yes, feel free to open up the lint rc and change the rule. You'll hear no complaints from this guy, especially given the example above. :+1:

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
erikquinn: I did. No idea why we would continue. We don't want any of the distances _within_ the airspace border, since we don't spawn there. So we should break, but we _do_ want to add this value to the total, so I added an adder within the breaker block. (<-- now that's a cool sounding sentence)

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
n8rzz: yeah, leave the `continue`.  I'll address that `no-continue` rule before the end of the sprint

src/assets/scripts/client/math/flightMath.js
```diff
@@ -203,8 +203,9 @@ export const calculateTurnInitiaionDistance = (aircraft, currentWaypointPosition
     const nominalBankAngleDegrees = 25;
     const speed = kn_ms(aircraft.speed);
     const bankAngle = degreesToRadians(nominalBankAngleDegrees);
+    const nextWaypointModel = aircraft.fms.currentLeg.nextWaypoint;
```
n8rzz: does this still work when the `currentWaypoint` _is_ the last waypoint in a leg?  

`aircraft.fms.getNextWaypointPositionModel()` 
- will check for that
- skip to the next leg and check for a waypoint 
- and only return the position.

src/assets/scripts/client/math/flightMath.js
```diff
@@ -213,7 +214,7 @@ export const calculateTurnInitiaionDistance = (aircraft, currentWaypointPosition
     }
 
     const nominalNewCourse = _calculateNominalNewCourse(
-        aircraft.fms.getNextWaypointPositionModel(),
+        nextWaypointModel.positionModel,
```
n8rzz: see above.  I question the removal of `.getNextWaypointPositionModel()`

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -537,22 +537,24 @@ export default class StandardRouteModel extends BaseModel {
...
-     * @parma waypointModelList {array<StandardRouteWaypointModel>}
+     * @param waypointModelList {array<StandardRouteWaypointModel>}
      * @private
      */
     _updateWaypointsWithPreviousWaypointData(waypointModelList) {
-        _forEach(waypointModelList, (waypoint, i) => {
-            let previousWaypoint = waypointModelList[i - 1];
-            if (i === 0) {
-                previousWaypoint = waypoint;
+        for (let i = 1; i < waypointModelList.length; i++) {
```
n8rzz: - why are we stating at 1 here?
- why did you remove the `_forEach()`?

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -537,22 +537,24 @@ export default class StandardRouteModel extends BaseModel {
...
+            const waypoint = waypointModelList[i];
+            const previousWaypoint = waypointModelList[i - 1];
+
+            let distance = 0;
```
n8rzz: this should move up a line and be under `previousWaypoint`.  then there should only be an empty line between `let distance` and the if block below.

src/assets/scripts/client/math/flightMath.js
erikquinn: Yeah that second condition won't be covered by this. :(

Really what I should have done was make a new `getNextWaypointModel`. Because we _do not_ want only the position model, due to our attempt to check `model.isVector` below, which can't be done with the position alone.

n8rzz: That should be easy enough. You can just modify the return of that method and update the tests.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: We can only do useful math when we have two waypoints (hence why the previous had `if (i === 0) { nextWaypoint = waypoint; }`). So we start at the second waypoint where we have our first opportunity to calculate a distance. Removal of the lodash was in favor of the native for loop for the ability to start at i=1.

n8rzz: That all makes sense.  Could you be include a quick comment on the why here?

src/assets/scripts/client/math/flightMath.js
erikquinn: done

src/assets/scripts/client/math/flightMath.js
erikquinn: `nextWaypointModel` will now be reliable, so this should remain the same?

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: done

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -391,6 +391,26 @@ ava('.skipToWaypoint() skips to a waypoint in a different leg', (t) => {
     t.true(fms.currentWaypoint.name === waypointNameMock);
 });
 
+ava('.getNextWaypointModel() returns the `WaypointModel` for the next Waypoint in the collection', (t) => {
```
n8rzz: Can we add one more case here where the `currentWaypoint` is the last one in the leg but not the last one in the flightPlan?

erikquinn: sure! done. :smiley:

Issue #469, opened 2017-05-18
Add support for an initial runway assignment in SpawnPatternModel


n8rzz:
This is needed in `SpawnPatternModel._generateWaypointListForRoute()` which is currently only using the `entry` and `body` segments or a route.  Sometimes this is not sufficient.  We need to be able to grab the `exit` segment as well.  

We have access to the `airportController` which can provide an initial runway.  This should be all we need to get the exit `segment`.

PR #472, opened 2017-05-19
feature/469

n8rzz:
completes #469 

Updates SpawnPatternModel to use the AirportModel.arrivalRunway property when gathering waypoint models needed to calculate initial aircraft heading

### DIFF
--- a/src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
+++ b/src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
@@ -973,17 +973,18 @@ export default class SpawnPatternModel extends BaseModel {
         this.heading = heading;
     }
 
+    // TODO: add support for `FIXXA..FIXXB.PRCDR1.FIXXC` route strings
     /**
-     *
+     * Given a `routeString`, find the `FixModel`s or `WaypointModel`s associated with that route.
      *
      * @for SpawnPatternModel
      * @method _generateWaypointListForRoute
-     * @param route {string}
+     * @param routeString {string}
      * @param navigationLibrary {NavigationLibrary}
      * @return {array<FixModel>|array<StandardWaypointModel>}
      */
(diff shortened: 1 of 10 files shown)


src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
```diff
@@ -995,19 +996,30 @@ export default class SpawnPatternModel extends BaseModel {
...
         return waypointModelList;
     }
 
+    /**
+     *
+     * @for SpawnPatternModel
+     * @method _generateSelfReferencedAirportPositionModel
+     * @return {StaticPositionModel}
+     */
     _generateSelfReferencedAirportPositionModel() {
```
n8rzz: @erikquinn I'd like to add a description here, but I'm not sure what this was for or why we need it.  Do you remember?

n8rzz:
@erikquinn bump

src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
erikquinn: Airports are the center of the world. So they don't have a "reference position" (which is basically used for the rotation from magvar).

test/airport/_mocks/airportJsonMock.js
```diff
@@ -305,6 +305,25 @@ export const AIRPORT_JSON_KLAS_MOCK = {
                 "25R": []
             }
         },
+        // not a real route. used here to test `SpawnPatternModel._generateWaypointListForRoute()`
+        "GRNPA9": {
+            "icao": "GRNPA9",
+            "name": "Grandpa One",
```
erikquinn: Not that it matters, but... `1 !== 9`

n8rzz: no, it doesn't matter here.  but you're right one does not equal nine.

PR #497, opened 2017-05-28
feature/451

### LINKED ISSUE #451: Vectors are treated as fixes
Airports that use the new # vector method throw warnings since it thinks they're fixes. There should probably be an exception for 'fixes' beginning with # when checking if they exist

Aircraft who are on a heading specified in the airport file show #[hdg] on their tags, when they should just show the heading without the `#`

---

Related: #467.

erikquinn:
Resolves #451.

Clears warnings shown for vector and hold fixes not being defined in the `fixes` of the airport file.

### DIFF
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
@@ -296,23 +296,31 @@ export default class StandardRouteWaypointModel extends BaseModel {
      * @for StandardRouteWaypointModel
      * @method clonePositionFromFix
      * @param fixCollection {FixCollection}
+     * @return {StandardRouteWaypointModel}
      * @private
+     * @chainable
      */
     clonePositionFromFix() {
-        if (this.name[0] === VECTOR_WAYPOINT_PREFIX) {
+        const isFlyOverWaypoint = this.name.indexOf(FLY_OVER_WAYPOINT_PREFIX) !== -1;
+        const isHoldWaypoint = this.name.indexOf(HOLD_WAYPOINT_PREFIX) !== -1;
+        const isVectorWaypoint = this.name.indexOf(VECTOR_WAYPOINT_PREFIX) !== -1;
+
+        if (isVectorWaypoint) {
             return;
         }
 
         let name = this.name;
 
-        if (this.name[0] === HOLD_WAYPOINT_PREFIX
-            || this.name[0] === FLY_OVER_WAYPOINT_PREFIX) {
+        if (isFlyOverWaypoint || isHoldWaypoint) {
+            // remove prefixing character (`@BIKKR` --> `BIKKR`)
             name = this.name.substr(1);
         }
 
         const fixModel = FixCollection.findFixByName(name);
 
         if (!fixModel) {
+            // TODO: This console warn should be done elsewehere, so a single console
+            // message can inform the airport designer of all missing fixes at once.
             console.warn(`The following fix was not found in the list of fixes for this Airport: ${this.name}`);
 
             return this;
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
@@ -1,6 +1,7 @@
 import _isNil from 'lodash/isNil';
 import BaseModel from '../../base/BaseModel';
 import FixCollection from '../Fix/FixCollection';
+import RouteModel from '../Route/RouteModel';
 import WaypointModel from '../../aircraft/FlightManagementSystem/WaypointModel';
 import { REGEX } from '../../constants/globalConstants';
 import {
@@ -301,9 +302,9 @@ export default class StandardRouteWaypointModel extends BaseModel {
      * @chainable
      */
     clonePositionFromFix() {
-        const isFlyOverWaypoint = this.name.indexOf(FLY_OVER_WAYPOINT_PREFIX) !== -1;
-        const isHoldWaypoint = this.name.indexOf(HOLD_WAYPOINT_PREFIX) !== -1;
-        const isVectorWaypoint = this.name.indexOf(VECTOR_WAYPOINT_PREFIX) !== -1;
+        const isFlyOverWaypoint = RouteModel.isFlyOverRouteString(this.name);
+        const isHoldWaypoint = RouteModel.isHoldRouteString(this.name);
+        const isVectorWaypoint = RouteModel.isVectorRouteString(this.name);
 
         if (isVectorWaypoint) {
             return;
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
@@ -168,7 +168,7 @@ export default class StandardRouteWaypointModel extends BaseModel {
         this.previousStandardWaypointName = '';
 
         return this._init(routeWaypoint)
-                   .clonePoisitonFromFix();
+                   .clonePositionFromFix();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -280,12 +281,23 @@ export default class StandardRouteWaypointModel extends BaseModel {
...
-        const fixModel = FixCollection.findFixByName(this.name);
+    clonePositionFromFix() {
+        if (this.name[0] === VECTOR_WAYPOINT_PREFIX) {
+            return;
+        }
+
+        let name = this.name;
+
+        if (this.name[0] === HOLD_WAYPOINT_PREFIX
+            || this.name[0] === FLY_OVER_WAYPOINT_PREFIX) {
```
n8rzz: This `||` should be on the previous line

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -280,12 +281,23 @@ export default class StandardRouteWaypointModel extends BaseModel {
...
      * @for StandardRouteWaypointModel
-     * @method _clonePoisitonFromFix
+     * @method clonePositionFromFix
      * @param fixCollection {FixCollection}
      * @private
      */
-    clonePoisitonFromFix() {
-        const fixModel = FixCollection.findFixByName(this.name);
+    clonePositionFromFix() {
+        if (this.name[0] === VECTOR_WAYPOINT_PREFIX) {
```
n8rzz: Why are we using `this.name[0]`?  Name should be a string not an array.

(But I'm on my phone and can't check easily)

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -280,12 +281,23 @@ export default class StandardRouteWaypointModel extends BaseModel {
...
+            return;
+        }
+
+        let name = this.name;
+
+        if (this.name[0] === HOLD_WAYPOINT_PREFIX
```
n8rzz: We should have some additional Iona's tests for the new branching logic here

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: Since a string is array-like, I'm grabbing the first character of the string to see if it is the character used as a vector waypoint name prefix.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: As in all on one line, or as in "when you break it up, leave the `||` on the previous line"?

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
n8rzz: ahh.  well, you are correct there.  However, when checking for an individual character in a string, one typically uses [indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf).  Otherwise another developer might make the same assumption I did.

```javascript
// instead of
if (this.name[0] === VECTOR_WAYPOINT_PREFIX) { }

// preferred
if (this.name.indexOf(VECTOR_WAYPOINT_PREFIX) !== -1) { }
```

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
n8rzz: the second one
```javascript
// instead of 
if (this.name[0] === HOLD_WAYPOINT_PREFIX
    || this.name[0] === FLY_OVER_WAYPOINT_PREFIX) { }

// preferred
if (this.name[0] === HOLD_WAYPOINT_PREFIX || 
    this.name[0] === FLY_OVER_WAYPOINT_PREFIX) { }
```

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: roger!

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: changed in favor of `if (isFlyOverWaypoint || isHoldWaypoint)` with previously declared variables for clarity anyway. Noted though, thank you :smiley:

erikquinn:
Marking as blocked to ensure I remove the added holds in KSEA before merging.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -283,23 +283,31 @@ export default class StandardRouteWaypointModel extends BaseModel {
...
      * @param fixCollection {FixCollection}
+     * @return {StandardRouteWaypointModel}
      * @private
+     * @chainable
      */
     clonePositionFromFix() {
-        if (this.name[0] === VECTOR_WAYPOINT_PREFIX) {
+        const isFlyOverWaypoint = this.name.indexOf(FLY_OVER_WAYPOINT_PREFIX) !== -1;
+        const isHoldWaypoint = this.name.indexOf(HOLD_WAYPOINT_PREFIX) !== -1;
+        const isVectorWaypoint = this.name.indexOf(VECTOR_WAYPOINT_PREFIX) !== -1;
```
n8rzz: This looks good!  As I'm reading this, though, don't we already have helpers for these in the (poorly named) `RouteModel` as static methods?

erikquinn: Hm, you have a point. Although actually no, it was the `WaypointModel` that has the handy methods. All the standard route stuff used to kinda make sense to me, though with time it's feeling more and more like a pretty confusing maze of things. I think you're right, we could use better names, and I _feel_ like we may be able to simplify. But you have spent much more time in that area than me, so perhaps not. Regardless, I can access `this._isVector` and `this._isFlyOverWaypoint`, but I don't have anything else like that for holds. I could delve into figuring it out, but at this point it doesn't seem like that'd quite be worth it in my opinion.

n8rzz: you're not thinking 4th dimensionally.  I'm talking about the `RouteModel` (which should actually be called the `RouteStringModel`.  We have static methods on that class to help in these types of circumstances:

```javascript
RouteModel.isVectorRouteString(this.name)
```

erikquinn: :bulb:

erikquinn: I see now. done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -288,9 +289,9 @@ export default class StandardRouteWaypointModel extends BaseModel {
      * @chainable
      */
     clonePositionFromFix() {
-        const isFlyOverWaypoint = this.name.indexOf(FLY_OVER_WAYPOINT_PREFIX) !== -1;
-        const isHoldWaypoint = this.name.indexOf(HOLD_WAYPOINT_PREFIX) !== -1;
-        const isVectorWaypoint = this.name.indexOf(VECTOR_WAYPOINT_PREFIX) !== -1;
+        const isFlyOverWaypoint = RouteModel.isFlyOverRouteString(this.name);
+        const isHoldWaypoint = RouteModel.isHoldRouteString(this.name);
+        const isVectorWaypoint = RouteModel.isVectorRouteString(this.name);
```
n8rzz: mmm, so smooth, Nice work!

erikquinn:
Changes to seattle reverted. Proceeding with merge!

Issue #285, opened 2017-03-07
Extract AircraftInstanceModel view logic to new class


n8rzz:
There is a ton of view logic in this class that needs to be moved to a more appropriate place.  This will likely require a fair amount of untangling in order to remove it successfully.

This new class will need to be plugged into the update cycle in order to stay up to date.

Specifications:
- The `AircraftInstanceModel` should hold the data.
- The `AircraftView` should take the data and display it.
- Remove number transformations in AircraftStripView.updateAircraftTelemetryText() such as `Math.ceil()`
- Must fulfill specifications of #168

n8rzz:
depends on openscope/openscope#457

erikquinn:
Resolved by #406. Closing.

PR #505, opened 2017-05-31
feature/434

### LINKED ISSUE #434: add support for non-procedural departures

n8rzz:
Resolves #434.

Adds tests for non-procedural departures and arrivals.  The support for this feature was already there, but masked by errors in the old `AircraftStripView` class ~~(soon-to-be-replaced by #285)~~.


src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -667,6 +667,7 @@ export default class Pilot {
             this._mcp.setHeadingFieldValue(runway.angle);
         }
 
+        // FIXME: should this be LNAV or HOLD?
         if (this._mcp.headingMode === MCP_MODE.HEADING.OFF) {
             this._mcp.setHeadingLnav();
         }
```
erikquinn: LNAV. They won't take off without being cleared on their route, and if they haven't received any specific heading instructions, the expectation should be that they do whatever is on their flightplan. So LNAV is :+1:

assets/airports/klas.json
```diff
@@ -392,212 +392,11 @@
             "destination": "",
             "origin": "KLAS",
             "category": "departure",
-            "route": "KLAS.BOACH6.HEC",
+            "route": "BESSY..BOACH..HEC",
             "altitude": null,
             "method": "random",
             "rate": 5,
             "speed": null
-        },
```
n8rzz: yikes!  this file shouldn't have been committed.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
n8rzz: `FIXME` removed

Issue #423, opened 2017-05-05
Ability to change length of PTL from within game


MisterHighQ:
Modify `PTL_LENGTH: 1` in `src/assets/scripts/client/constants/canvasConstants.js` to a variable PTL_LENGTH that can be changed from within the settings of the game.

Edit: Presently, it's difficult to focus on the direction and current position of the aircraft with the 1 minute PTL lengths that are too long and distracting.

n8rzz:
@erikquinn hmm, I think we actually have an issue for this already?  Maybe buried in the requirements of some User Settings issue somewhere?

erikquinn:
Yes, #115.

In the meantime, it may be worth adding an "ugly solution" such as `scope {ptl_length: 0.5}` where users can change a few basic things without an actual attractive UI. Because changing the settings is the easy part it's creating the buttons that will take longer to get around to.

MisterHighQ:
That alone would be fantastic until something more rigid and permanent can be developed!

Issue #14, opened 2017-01-02
Scope Commands


erikquinn:
There are various commands that the *CONTROLLER* themselves actually will enter into their scope. Includes things like:
 - datablock offset direction and length
 - handoffs to other controllers (see zlsa/atc#358)
 - scratchpad entries

We need the ability to "toggle" between two different command bars. One that controls the simulation (airplanes), and one that allows us to manipulate our scope as a controller would. I'm thinking tab to switch between them.

Some commands will be simply typed, and others will have Q-messages (or similar) accessible through hotkeys.

---

Planned key mappings:
```
F1:  INIT TRK (Initiate Track)
F2:  VCT DEC (Vector Decrease)
F3:  VCT INC (Vector Increase)
F4:  DROP TRK (Drop Track)
F5:  INT (Interim Altitude)
F6:  RTE (Route Display)
F7:  HALO (Halo Display)
F8:  ALT (Altitude)
F9:  FR (Flightplan Request)
F10: PVD (Push Data Block)
```

PR #507, opened 2017-06-01
feature/423

n8rzz:
completes openscope/openscope#423 

Adds PTL length game option to the settings menu


src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -34,6 +35,16 @@ export const GAME_OPTION_VALUES = [
         ]
     },
     {
+        name: GAME_OPTION_NAMES.PTL_LENGTH,
+        defaultValue: '1',
+        description: 'Projected Track Line (PTL) Length',
```
erikquinn: This wraps because it is a bit too long. Since the whole `Projected Track Line` won't fit, I think we might as well just settle for the abbreviation, despite the downside that many people will not know what a "PTL" is. Though I think they'd probably change the value of the setting and figure it out. Maybe we can do `PTL Length (minutes)` here.

image

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -34,6 +35,16 @@ export const GAME_OPTION_VALUES = [
...
+        type: 'select',
+        data: [
+            ['1', '1'],
+            ['0.5', '0.5']
```
erikquinn: Let's add an `off` option, as well as `2`, `3`, and `5`.

src/assets/scripts/client/game/GameController.js
```diff
@@ -402,4 +403,14 @@ export default class GameController {
     getGameOption(optionName) {
         return this.game.option.get(optionName);
     }
+
+    /**
+     *
+     * @method getPtlLengthMultiplier
+     */
+    getPtlLengthMultiplier() {
```
erikquinn: I don't like this name here; we are never applying a multiplicative factor to the length of the PTL, we're setting the length to a given value. So `getPtlLength` would be more appropriate here. It's not really as if the default is 1 and then we scale that up and down, we just change the value to whatever is specified. It's sort of a semantical thing I suppose, but still.

src/assets/scripts/client/constants/gameOptionConstants.js
n8rzz: This can be fixed with css and we can keep the title.  there is no reason the select boxes should take up that much space, so I'm going to fix that 😉

src/assets/scripts/client/constants/gameOptionConstants.js
n8rzz: on `ksfo`, a length of `2` will cause the PTL from an arrival in the southeast to extend clear across _the entire_ airspace.

src/assets/scripts/client/game/GameController.js
n8rzz: agreed

n8rzz:
@erikquinn keep in mind this PR is very much in progress.  But I made some good headway on it last night so I figured I'd open this up so you could poke around (and I'm glad you did).  I won't typically do the changelog until I'm done to avoid merge conflicts.

src/assets/scripts/client/constants/gameOptionConstants.js
erikquinn: Definitely a good point, that'll work then. I do think we should indicate that the unit is _minutes_ because people will tend to assume it's a length of distance not time. That could be done in the display values also.

src/assets/scripts/client/constants/gameOptionConstants.js
erikquinn: :confused: That can't possibly be right. That'd be WAY longer than 2 minutes? Something smells fishy here.

erikquinn: Yeah, I'm not able to reproduce a 2min PTL going that far. Looks like it takes more than a 10 minute to stretch that far in most cases. Anyway, I'm not suggesting that anyone would leave it on such an absurd setting, but giving it as an option... we might as well. It would be a pain to change back and forth from settings like this, but the plan is for vector lines to be mapped to a hotkey (decrease/increase with F2/F3; see #14), and that's when the longer lengths really become useful: when you can run them out real quick to show a projection between converging traffic, then a few taps to shrink them back down to normal size.

n8rzz: there may have been an operator error on that one, cause I can't recreate it anymore either.

easy options to add.  I may actually take the time to re-work these `gameOptions` a little bit because its a little clunky.

n8rzz: yeah, reworking will let me use option displays like `30sec` and `1min` but use values of `0.5` and `1` respectively.  This also means I can get rid of the `_startCase()` stuff in there.

n8rzz:
@erikquinn bump

Issue #72, opened 2017-01-17
Convert appropriate classes to static classes


erikquinn:
_From @n8rzz on November 15, 2016 14:41_

Are there reasons we shouldn't do this?  There are several classes that are essentially static, but are being used as instances

- [ ] GameController
- [ ] CanvasController
- [ ] AirportController
- [ ] AircraftController
- [ ] AirlineController

---

There may be more than what is listed above.  It might not make sense to convert some to static.  It really only makes sense with a class like GameController and CanvasController where other classes use their methods and internal properties.

Static classes may not be the best solution.  More research and discussion is needed in order to flesh this out.

_Copied from original issue: n8rzz/atc#99_

n8rzz:
This makes me puke:

```javascript
this.gameController = new GameController(this.getDeltaTime);
window.gameController = this.gameController;

this.airportController = new AirportController(initialAirportData, airportLoadList);
// TODO: Temporary
window.airportController = this.airportController;

this.navigationLibrary = new NavigationLibrary(initialAirportData);
this.airlineController = new AirlineController(airlineList);
this.aircraftController = new AircraftController(aircraftTypeDefinitionList, this.airlineController, this.navigationLibrary);
// TODO: Temporary
window.aircraftController = this.aircraftController;

this.spawnPatternCollection = new SpawnPatternCollection(initialAirportData, this.navigationLibrary, this.airportController);
this.spawnScheduler = new SpawnScheduler(this.spawnPatternCollection, this.aircraftController, this.gameController);
this.uiController = new UiController(this.$element, this.gameController, this.airportController);
this.canvasController = new CanvasController(this.$element, this.navigationLibrary, this.gameController);
this.tutorialView = new TutorialView(this.$element);
this.aircraftCommander = new AircraftCommander(this.airportController, this.navigationLibrary, this.gameController, this.uiController);
this.inputController = new InputController(this.$element, this.aircraftCommander, this.uiController, this.aircraftController);
this.gameClockView = new GameClockView(this.$element);
```

n8rzz:
blocked by #137

PR #519, opened 2017-06-14
feature/483

### LINKED ISSUE #483: Add unique TransponderCodes and CIDs
we need a way to:
-  keep track of issued codes (codes in use)
- generate new codes for spawning arrival aircraft
- warn controller when attempting to assign a code already in use

this need not be a big class.  we can follow a similar pattern that was used for managing flight numbers and use a recursive pattern for generating unique numbers.  this class would likely live in the `AircraftController` and would be consulted with when instantiating a new aircraft _or_ when a new code is about to be issued.

---
- [x] make stripView list closed onload
- [x] rotate stripView-trigger based on open state
- [x] add CID number generator
    - [x] unique
    - [x] non-sequential
    - [ ] support `001 - 999`
- [x] add transponderCode generator method
    - [ ] allow `N` numbers to receive `1200`
    - [x] disallow invalid transponder codes
- [x] hook into command lifecycle when transponder command is issued

n8rzz:
Resolves #483.

Add unique TransponderCodes and CIDs


src/assets/scripts/client/App.js
```diff
@@ -199,7 +199,7 @@ export default class App {
         this.uiController = new UiController(this.$element, this.gameController, this.airportController);
         this.canvasController = new CanvasController(this.$element, this.navigationLibrary, this.gameController);
         this.tutorialView = new TutorialView(this.$element);
-        this.aircraftCommander = new AircraftCommander(this.airportController, this.navigationLibrary, this.gameController, this.uiController);
+        this.aircraftCommander = new AircraftCommander(this.airportController, this.navigationLibrary, this.gameController, this.uiController, this.aircraftController.onRequestToChangeTransponderCode);
```
n8rzz: I really don't like this, but its a necessary evil at the moment.  Once #72 is in place, this win't look so messy.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -568,6 +641,80 @@ export default class AircraftController {
...
+     * soon-to-be-instantiated `AircraftModel`
+     *
+     * @for AircraftController
+     * @method _generateUniqueTransponderCode
+     * @return {number}
+     * @private
+     */
+    _generateUniqueTransponderCode() {
+        const minCodeValue = 1203;
+        const maxCodeValue = 7499;
```
erikquinn: `7477` (octal)

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -568,6 +641,80 @@ export default class AircraftController {
...
+     * Boolean helper used to determine if a given `transponderCode` is both
+     * not in use and not an invalid transponderCode.
+     *
+     * @for AircraftController
+     * @method _isValidTransponderCode
+     * @param transponderCode {number}
+     * @return {boolean}
+     */
+    _isValidTransponderCode(transponderCode) {
+        return !this._isTransponderCodeInUse(transponderCode) && INVALID_SQUAWK_CODES.indexOf(transponderCode) === -1;
```
erikquinn: More will have to be added here to prevent non-octal squawk codes. Only digits 0-7 are allowable.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -568,6 +641,80 @@ export default class AircraftController {
...
+     * @private
+     */
+    _generateUniqueTransponderCode() {
+        const minCodeValue = 1203;
+        const maxCodeValue = 7499;
+        const transponderCode = _random(minCodeValue, maxCodeValue);
+
+        if (!this._isValidTransponderCode(transponderCode)) {
+            // the value generated is already in use, recurse back through this method and try again
+            this._generateUniqueTransponderCode();
```
erikquinn: Is there not a better way to do this and avoid recursion? Like maybe something like a randomly sorted array of all available ones, then pulling from that list? Seems like a simpler guarantee of uniqueness than generating random ones and then checking to see if they're unique. Particularly with the CIDs (since there's only 999 of them), once you get over a hundred or so aircraft, the frequency of a need to re-generate a new one because the random selection is actually in use is quite a bit greater. Just curious if there's another approach that might be better.

src/assets/scripts/client/aircraft/StripView/StripViewController.js
```diff
@@ -257,7 +277,53 @@ export default class StripViewController {
...
+        const nextCid = _random(1, CID_UPPER_BOUND);
+        const nextCidLeft = nextCid - 1;
+        const nextCidRight = nextCid + 1;
+
+        if (
+            this._cidNumbersInUse.indexOf(nextCid) !== -1 ||
+            this._cidNumbersInUse.indexOf(nextCidLeft) !== -1 ||
+            this._cidNumbersInUse.indexOf(nextCidRight) !== -1
+        ) {
+            this._generateCidNumber();
```
erikquinn: I don't follow... Why are we requiring that any CID cannot be adjacent to another CID in use?

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: I think that is actually covered by the `commandParser`

src/assets/scripts/client/aircraft/StripView/StripViewController.js
n8rzz: you said non-sequential CID numbers.  This if block looks for our random number `n` and then additionally `n - 1` and `n + 1`.  If any of those three exist, we re-generate because we've found either a duplicate, or a set of sequential numbers.

erikquinn: :man_facepalming: I see what happened. Sorry, I was definitely not real clear. What I meant by non-sequential was not to prevent the _existence_ of "adjacent" numbers, but rather to ensure that they are _generated non-sequentially_. As in, the first 3 CIDs ought not to be `001, 002, 003`, but that's not to say that we couldn't have `[054, 126, 354, 355, 617]`. My B!

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: what are your concerns with recursion?  using recursion is actually very efficient in javascript, especially in this case.

It also keeps close to a pattern used in a few other places, so if we wanted to abstract the logic here we could.

As far as performance, I would argue that using a recursion strategy is slightly faster.  But overall, I wouldn't be surprised if either strategy worked out to about the same performance.  I just feel that this reads better and is easier to maintain (plus the aforementioned pattern)

src/assets/scripts/client/aircraft/StripView/StripViewController.js
n8rzz: ohhhhh, well that makes much more sense. let me remove that block and make that into a nice small if statement.

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: You would know. It just seemed to me like it could be dangerous if we start to fill it up (I mean, if we have like 500 planes in the simulation, we'll probably have much bigger issues anyway...) Having to repeat two or three or four times seemed to me like it would be something better avoided. But if it doesn't have an adverse effect on performance, frankly I couldn't care any less! lol

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: I seems like we are missing a check for numbers `> 7`.  If we find one, that is also invalid.  Am I understanding that correctly?

erikquinn: Correct. Valid transponder codes may only use digits 0 through 7. Though I would say the `INVALID_SQUAWK_CODES` should be renamed to `RESERVED_SQUAWK_CODES` for clarity, because they are not inherently "invalid", they just aren't what we can assign aircraft for purposes of tracking them: "discrete codes" (aka unique, non-reserved codes).

n8rzz: sounds good, I'll make those changes

n8rzz:
@erikquinn you're going to get a kick out of my solution here 😉

src/assets/scripts/client/math/core.js
```diff
@@ -217,3 +219,21 @@ export const extrapolate_range_clamp = (range1_min, target_val, range1_max, rang
 
     return clamp(extrapolation_result, range2_min, range2_max);
 };
+
+/**
+ * Generate a random number with each digit between 0-7
+ *
+ * @function generateRandomOctalWithLength
+ * @return {number}                         number with digits between 0-7
```
n8rzz: this is actually a string

src/assets/scripts/client/math/core.js
```diff
@@ -217,3 +219,21 @@ export const extrapolate_range_clamp = (range1_min, target_val, range1_max, rang
...
+ */
+export const generateRandomOctalWithLength = (length = 1) => {
+    const value = [];
+
+    for (let i = 0; i < length; i++) {
+        const randomOctal = _random(0, 7);
+
+        value.push(randomOctal);
```
erikquinn: :+1:

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -568,6 +644,91 @@ export default class AircraftController {
...
+     * not in use and not an invalid transponderCode.
+     *
+     * @for AircraftController
+     * @method _isValidTransponderCode
+     * @param transponderCode {number}
+     * @return {boolean}
+     */
+    _isValidTransponderCode(transponderCode) {
+        return RESERVED_SQUAWK_CODES.indexOf(transponderCode) === -1;
+    }
```
erikquinn: In order to be valid, we must be checking that it is 4 digits or less, all of which are octal. This is checking that it's not reserved, which is not related to it being "valid". The only two conditions to check here are that it is 4 digits or less and that it contains only octal numbers.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -568,6 +644,91 @@ export default class AircraftController {
...
+     * valid and not in use.
+     *
+     * @for AircraftController
+     * @method _isDiscreteTransponderCode
+     * @param transponderCode {number}
+     * @return {boolean}
+     */
+    _isDiscreteTransponderCode(transponderCode) {
+        return !this._isTransponderCodeInUse(transponderCode) && this._isValidTransponderCode(transponderCode);
+    }
```
erikquinn: In order to be "discrete", it must be valid and not one of the reserved codes. It being in use or not in use is not related to it being discrete.

`return this._isValidTransponderCode(transponderCode) && RESERVED_SQUAWK_CODES.indexOf(transponderCode) === -1;`

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: Done

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: Done

PR #523, opened 2017-06-15
feature/137

### LINKED ISSUE #137: Create an AppController
_From @n8rzz on December 31, 2016 18:10_

We need a class that can act as a conductor between the `App`, with the game loop, and everything else.  This class would be responsible for orchestrating initial data load, airport change as well as providing a launching point for additional view classes that we will need in the future.

Instead of `App` creating all the top level classes, `App` would instantiate this class and it would then create everything else.

- [x] accepts `$element`, `airportLoadList`, `initialAirportIcao` as parameters
- [x] provides facade methods that are run within the `App` game loop
- [x] provides callback method(s) that can be sent to various classes for handling events (ex: airportChange)

_Copied from original issue: n8rzz/atc#252_

n8rzz:
Resolves #137.

Adds AppController class
* refactors App.js to handle only the game loop and then call facade
* methods on the AppController


src/assets/scripts/client/App.js
```diff
@@ -66,27 +43,18 @@ export default class App {
...
         this.prop = prop;
         this.prop.complete = false;
-        this.prop.temp = 'nothing here';
         this.prop.version = VERSION;
         this.prop.time = {};
         this.prop.time.start = time();
         this.prop.time.frames = 0;
-        this.prop.time.frame = {};
+        // this.prop.time.frame = {};
         this.prop.time.frame.start = time();
```
erikquinn: By commenting this out above, we get errors when trying to reference any properties of `this.prop.time.frame` (obviously). Initial load subsequently fails.

n8rzz: yes you would.  this is very much a WIP, I just wanted to push it and open the PR to see where I'm at with develop.  and to let you take a peak, too.

n8rzz:
blocked by #483

src/assets/scripts/client/AppController.js
```diff
@@ -0,0 +1,249 @@
...
+        this.aircraftCommander = null;
+        this.inputController = null;
+        this.uiController = null;
+        this.canvasController = null;
+
+        return this._init()
+            .enable();
+    }
+
+    _init() {
```
n8rzz: 🤔 

Method comments will be included in #72 ( #524 )

PR #547, opened 2017-07-01
bugfix/546

### LINKED ISSUE #546: Fms.flightPlanAltitude always 0

erikquinn:
Resolves #546.

Make strips show appropriate altitude values

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -45,7 +45,8 @@ import {
 } from '../utilities/radioUtilities';
 import {
     degreesToRadians,
-    nm
+    nm,
+    UNIT_CONVERSION_CONSTANTS
 } from '../utilities/unitConverters';
 import {
     MCP_MODE,
@@ -589,15 +590,15 @@ export default class AircraftModel {
      * @return {object<string, string>}
      */
     getViewModel() {
-        let assignedAltitude = this.mcp.altitude;
-        let flightPlanAltitude = this.fms.flightPlanAltitude;
+        let assignedAltitude = '-';
+        let flightPlanAltitude = '-';
 
-        if (assignedAltitude === INVALID_NUMBER) {
-            assignedAltitude = '-';
+        if (this.mcp.altitude !== INVALID_NUMBER) {
+            assignedAltitude = this.mcp.altitude * UNIT_CONVERSION_CONSTANTS.FT_FL;
         }
 
-        if (flightPlanAltitude === INVALID_NUMBER) {
-            flightPlanAltitude = '-';
+        if (this.fms.flightPlanAltitude !== INVALID_NUMBER) {
+            flightPlanAltitude = this.fms.flightPlanAltitude * UNIT_CONVERSION_CONSTANTS.FT_FL;
         }
 
         return {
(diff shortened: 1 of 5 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -592,12 +593,17 @@ export default class AircraftModel {
         let assignedAltitude = this.mcp.altitude;
         let flightPlanAltitude = this.fms.flightPlanAltitude;
 
+
         if (assignedAltitude === INVALID_NUMBER) {
             assignedAltitude = '-';
+        } else {
+            assignedAltitude *= UNIT_CONVERSION_CONSTANTS.FT_FL;
```
n8rzz: this should be moved up to the original assignment on ln 593

```
let assignedAltitude = this.mcp.altitude * UNIT_CONVERSION_CONSTANTS.FT_FL;
```

same with `flightPlanAltitude`. That way we can get rid of the `else`s

src/assets/scripts/client/utilities/unitConverters.js
```diff
@@ -61,11 +61,19 @@ export const UNIT_CONVERSION_CONSTANTS = {
...
      *
-     * @property FL_TO_FT_MULTIPLIER
+     * @property FL_FT
      * @type {number}
      * @final
      */
-    FL_TO_FT_MULTIPLIER: 100
+    FL_FT: 100,
+    /**
+     * Number used to obtain flight level from altitud in feet
```
n8rzz: altitud -> altitude

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: Knew you'd say that. 😉 I had already made that change before I figured out it won't work out that way. If we start by setting it to `mcp.altitude * UCC.FT_FL`, we can't ask `if (assignedAltitude === INVALID_NUMER)`. We _could_ ask `if (assignedAltitude === INVALID_NUMBER * UCC.FT_FL)`, but that seemed rather.... odd. At least when compared to adding an else.

src/assets/scripts/client/utilities/unitConverters.js
erikquinn: No me gusta tu actitud

erikquinn: Pero voy a repararlo no obstante! 😉

erikquinn: done

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: 💩 

it seems you know me well, as well.  I see what you're saying, though.  I don't like to see `else`s.

what if, instead, we did this:
```
let assignedAltitude = '-'; 

if (this.mcp.altitude !== INVALID_NUMBER) {
  assignedAltitude = this.mcp.altitude * UNIT_CONVERSION_CONSTANTS.FT_FL;
}
```

I don't like this because we will hit the if more often than not, but its cleaner.

erikquinn: clever. i like it! :shipit:

PR #550, opened 2017-07-01
hotfix/549

### LINKED ISSUE #549: Direction is wrongfully enforced on `fh`
Aircraft should be taking the fastest direction of turn to reach the heading specified in a `fh` command, but instead they continue with the direction of the last turn they were given.

re-pro:
```
t l 250
(then, while still in the turn...)
fh 350
```

erikquinn:
Resolves #549

Ensure previously specified directions of turn are not preserved when a new heading instruction is given

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -229,11 +229,8 @@ export default class AircraftCommander {
         const heading = data[1];
         const incremental = data[2];
         const readback = aircraft.pilot.maintainHeading(aircraft.heading, heading, direction, incremental);
-        const airport = AirportController.airport_get();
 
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -229,11 +229,8 @@ export default class AircraftCommander {
         const heading = data[1];
         const incremental = data[2];
         const readback = aircraft.pilot.maintainHeading(aircraft.heading, heading, direction, incremental);
-        const airport = AirportController.airport_get();
 
-        if (readback[0] && direction) {
-            aircraft.target.turn = direction;
-        }
+        aircraft.target.turn = direction;
```
n8rzz: If `direction` is a string, it should be reset to an empty string here.

Issue #267, opened 2017-02-21
Freezes if assigning invalid runway for SID


BingShearer:
The game and time just stop when I try the following command. "***### caf cvs w 1R" at KSFO where the assigned SID is MOLEN7.

erikquinn:
This is a known issue (#27), and will hopefully be fixed in the current sprint (ending March 21st). Appreciate the report though!

Closing as duplicate.

PR #551, opened 2017-07-01
bugfix/383

### LINKED ISSUE #383: Incorrect runway alternative chosen for SIDs
It seems that the second aircrafts picks incorrect runway from the SID "rwy" alternatives.

E.g.:

1. Choose LOWW
2. Select an aircraft bound to OSPEN
3. Issue "taxi 11 caf cvs" and then "to" to it
4. Repeat for another aircraft bound to OSPEN but use runway 16 instead of 11
5. The second aircraft seems to head towards WW100 rather than WW269

erikquinn:
Resolves #383.

Recalculate SID/STAR legs when changing assigned runway


src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -699,11 +703,6 @@ export default class Fms {
      * @param runwayModel {RunwayModel}
      */
     replaceDepartureProcedure(routeString, runwayModel) {
-        // this is the same procedure that is already set, no need to continue
-        if (this.hasLegWithRouteString(routeString)) {
```
n8rzz: why not just move this `if` block below the runway check?  not doing this at all means we could, potentially, attempt to replace a given flightPlan with exactly the same thing.  this block gives us an early out in that case and saves some processing.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -737,11 +736,6 @@ export default class Fms {
      * @param runwayModel {RunwayModel}
      */
     replaceArrivalProcedure(routeString, runwayModel) {
-        // this is the same procedure that is already set, no need to continue
-        if (this.hasLegWithRouteString(routeString)) {
```
n8rzz: same here as with `replaceDepartureProcedure`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -699,11 +703,6 @@ export default class Fms {
      * @param runwayModel {RunwayModel}
      */
     replaceDepartureProcedure(routeString, runwayModel) {
-        // this is the same procedure that is already set, no need to continue
-        if (this.hasLegWithRouteString(routeString)) {
-            return;
-        }
-
         if (this.departureRunwayModel.name !== runwayModel.name) {
             this.setDepartureRunway(runwayModel);
```
n8rzz: are you intending to use recursion here?  when `fms.replaceDepartureProcedure()` is called with a different runway, we would get the following:
```
.replaceDepartureProcedure()
  - > .setDepartureRunway()
    -> . _regenerateSidLeg()
        -> .replaceDepartureProcedure()
[finish first call to .replaceDepartureProcedure()]
```

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -737,11 +736,6 @@ export default class Fms {
      * @param runwayModel {RunwayModel}
      */
     replaceArrivalProcedure(routeString, runwayModel) {
-        // this is the same procedure that is already set, no need to continue
-        if (this.hasLegWithRouteString(routeString)) {
-            return;
-        }
-
         if (this.arrivalRunwayModel.name !== runwayModel.name) {
             this.setArrivalRunway(runwayModel);
```
n8rzz: same here as with `replaceDepartureProcedure`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -1357,6 +1351,42 @@ export default class Fms {
...
+        const starLegIndex = this._findLegIndexForProcedureType(PROCEDURE_TYPE.STAR);
+
+        if (starLegIndex === INVALID_INDEX) {
+            return;
+        }
+
+        const starLeg = this.legCollection[starLegIndex];
+
+        this.replaceArrivalProcedure(starLeg.routeString, this.arrivalRunwayModel);
+    }
```
n8rzz: current review ends here. will continue later tonight.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: > we could, potentially, attempt to replace a given flightPlan with exactly the same thing

That's what I was actually trying to do. Perhaps there is a better way. But I was saying whenever you assign a new departure runway, regenerate FROM SCRATCH the SID leg. If reassigning the same runway, or a runway where the resulting waypoints are identical, it is obviously wasted energy, but if a single fix is different (or same fix with an altitude restriction that only applies from one runway), by regenerating fresh, I can be confident we will always have the correct waypoints in place. 

Thoughts?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: Not really. I couldn't think of a good way to get it done so I just left that bit as it was. We need to be able to:
- assign a new runway and have the Sid leg regenerated
- assign a new SID and have some way to prevent aircraft from attempting to use it from their current expected runway when it's invalid for that procedure
    - we could forcibly reassign a new runway, though how would we deal with aircraft already in the queue when they get the new Sid?
    - we could simply reject the Sid to the user with a red message, so they will have to change their departure runway before being able to accept the reroute (problematic because currently no `expect` command)

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
n8rzz: I smell what you're stepping in.  But, I need to ask; what if you have the same `routeString` _and_ the same `runwayModel`?  Meaning, the user has issued the command for their current route (whether on purpose or accident).  

Then what?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
n8rzz: I see what you mean there, we actually need to run through it twice.  Once for the runway change and once for a route change.  That makes sense.

In that case, you should add a comment in here about the why.  someone looking through the methods would be curious on why things are laid out that way.

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -153,7 +154,15 @@ ava('.setDepartureRunway() throws when passed a string instead of a RunwayModel'
...
 
-ava.todo('.setDepartureRunway() validates route with changed runway');
+ava('.setDepartureRunway() regenerates SID legs for new runway', (t) => {
+    const nextRunwayMock = airportModelFixture.getRunway('19R');
+    const fms = buildFmsMockForDeparture();
+    const regenerateSidLegSpy = sinon.spy(fms, '_regenerateSidLeg');
+
+    fms.setDepartureRunway(nextRunwayMock);
+
+    t.true(regenerateSidLegSpy.called);
```
n8rzz: sinon has methods `.calledWith()` and `.calledWithExactly()`.  I would encourage you to use one (ideally the latter) of those instead of `called`.

This allows us to not only test that a method is called, but also that it is called with the correct stuff.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: Then there's no need to regenerate. But since we don't really have the ability to ask `what expected runway were the waypoints for this model based on?`, we don't really have a way to know whether the runway being assigned is the same as what it was previously. Maybe? : thinking_face:

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -436,16 +455,6 @@ ava('.getNextWaypointPositionModel() returns the `StaticPositionModel` for the n
 
 ava.todo('.replaceDepartureProcedure() updates the current runway assignment');
 
-ava('.replaceDepartureProcedure() returns early if the nextRouteString matches the current route', (t) => {
```
n8rzz: be sure to add some test cases for when the passed runway is the same, or different, than the one that is stored

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
...
 
+ava('._regenerateSidLeg() creates a new SID leg of identical route string for the currently assigned departure runway', (t) => {
+    const initialRunwayMock = airportModelFixture.getRunway('19L');
+    const nextRunwayMock = airportModelFixture.getRunway('19R');
+    const fms = buildFmsMockForDeparture();
+    const replaceDepartureProcedureSpy = sinon.spy(fms, 'replaceDepartureProcedure');
+    const sidLegIndex = fms._findLegIndexForProcedureType(PROCEDURE_TYPE.SID);
+    const oldWaypoints = fms.legCollection[sidLegIndex].waypointCollection;
+
+    t.true(_isEqual(fms.departureRunwayModel, initialRunwayMock));
```
n8rzz: I'm not sure #831 - #835 are needed.  this looks like it's just verifying the current state of the fms, is that right?

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
     t.true(fms.legCollection[1].routeString === routeAmmendmentMock[1]);
 });
 
+ava('._regenerateSidLeg() creates a new SID leg of identical route string for the currently assigned departure runway', (t) => {
```
n8rzz: be careful with multiple matchers (`t.true`, `t.false`, etc) in a single test.  In the event one of them fails you will not get any indication on which one, specifically, failed.  

try to keep tests small as possible.  this is a big one, though I think it could be shorter.  see below

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
...
+    t.true(oldWaypoints.length === 7);
+
+    fms.departureRunwayModel = nextRunwayMock;
+    fms._regenerateSidLeg();
+
+    const newWaypoints = fms.legCollection[sidLegIndex].waypointCollection;
+
+    t.true(newWaypoints[0].name === 'jaker');
+    t.true(newWaypoints[1].name === 'roppr');
+    t.true(newWaypoints[1].altitudeMaximum === 7000);
```
n8rzz: if we have a waypoint `roppr` couldn't we also assume that we have the altitude restriction as well?  The thing we are testing here is that the flight plan has been changed, thats it.  Restrictions and such would likely be tested elsewhere.

I don't mind the verbosity here, but I think this test would be just as effective with:
```javascript
t.true(newWaypoints[0].name === 'jaker');
t.true(newWaypoints.length === 7);
t.true(replaceDepartureProcedureSpy.calledWithExactly(routeString, departureModel);
```

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
     t.true(fms.legCollection[1].routeString === routeAmmendmentMock[1]);
 });
 
+ava('._regenerateSidLeg() creates a new SID leg of identical route string for the currently assigned departure runway', (t) => {
+    const initialRunwayMock = airportModelFixture.getRunway('19L');
```
n8rzz: I believe this should be called a _fixture_ because this is an actual `RunwayModel` 

`mock`: [fake] data that makes up a specific object or item in an app
`fixture`: an instantiated class, with working methods, that can be used under test

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
...
+    const replaceDepartureProcedureSpy = sinon.spy(fms, 'replaceDepartureProcedure');
+
+    fms.departureRunwayModel = airportModelFixture.getRunway('19R');
+    fms._regenerateSidLeg();
+
+    t.true(replaceDepartureProcedureSpy.notCalled);
+});
+
+ava('._regenerateStarLeg() creates a new STAR leg of identical route string for the currently assigned arrival runway', (t) => {
+    const initialRunwayMock = airportModelFixture.getRunway('19L');
```
n8rzz: all items stated in `.regenerateSidLeg()` apply to this test as well

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: done

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -171,7 +180,17 @@ ava('.setArrivalRunway() throws when passed a string instead of a RunwayModel',
...
+ava('.setArrivalRunway() regenerates STAR legs for new runway', (t) => {
+    const nextRunwayMock = airportModelFixture.getRunway('19R');
+    const fms = buildFmsMock();
+    const regenerateStarLegSpy = sinon.spy(fms, '_regenerateStarLeg');
+
+    t.true(fms.arrivalRunwayModel.name === '19L');
+
+    fms.setArrivalRunway(nextRunwayMock);
+
+    t.true(regenerateStarLegSpy.called);
```
erikquinn: Replaced in favor of `.calledWithExactly()`

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
...
+    fms.departureRunwayModel = nextRunwayMock;
+    fms._regenerateSidLeg();
+
+    const newWaypoints = fms.legCollection[sidLegIndex].waypointCollection;
+
+    t.true(newWaypoints[0].name === 'jaker');
+    t.true(newWaypoints[1].name === 'roppr');
+    t.true(newWaypoints[1].altitudeMaximum === 7000);
+    t.true(newWaypoints.length === 7);
+    t.true(replaceDepartureProcedureSpy.called);
```
erikquinn: Replaced in favor of `.calledWithExactly()`

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -819,6 +820,74 @@ ava('._prependLegCollectionWithRouteAmendment() adds LegModels for each routeStr
...
+    t.true(oldWaypoints.length === 13);
+
+    fms.arrivalRunwayModel = nextRunwayMock;
+    fms._regenerateStarLeg();
+
+    const newWaypoints = fms.legCollection[starLegIndex].waypointCollection;
+
+    t.true(newWaypoints[12].name === 'right');
+    t.true(newWaypoints.length === 13);
+    t.true(replaceArrivalProcedureSpy.called);
```
erikquinn: Replaced in favor of `.calledWithExactly()`

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: roger that!

done

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: covered not here, but in freshly added test for early return from `.setArrivalRunway()` and `.setDepartureRunway()`.

done

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: I'm not seeing any instances of multiple matchers, but I wasn't inclined to avoid it since I didn't know. I'll keep that in mind going forward.

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: Correct, the idea being that the test only gives a good result if we know the first value of the `departureRunwayModel` is one value which is changed to the specified value. I could skip the verification on the first part, but kinda thought it provided protection in case the fixtures were ever changed. For example, in this case, if the `buildFmsMockForDeparture()` was changed to use `19R` by default, this test would pass, even if the method was broken. I was trying to forcibly show that the value was changed between two expected values. Easily could be seen as over-defensive though.

Think it's better to just check the "after" instead?

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: fair enough!

done

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: >But, I need to ask; what if you have the same routeString and the same runwayModel?

Actually, we can address this without messing any of that up by having `.setDepartureRunway()` return early if the runway being "set" is the same as the one already in there. Then, `.replaceDepartureProcedure()` won't be called. And we can continue with the conservative approach of having `.replaceDepartureProcedure()` freshly regenerating the waypoints. Sounds to me like the best solution.

n8rzz: agreed.  I just want to be sure we aren't calling methods unnecessarily.  Perhaps we could also add a call count assertion for the tests here.

erikquinn: done

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: reduced all this to only examining the first fix, but continuing with the before-and-after comparison to protect from any changes to the fixtures.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -699,12 +711,14 @@ export default class Fms {
...
-        }
-
         if (this.departureRunwayModel.name !== runwayModel.name) {
+            // This does result in needless recursion (since `setDepartureRunway()`
+            // calls `replaceDepartureProcedure()`, but it is necessary because we
+            // need to be able to both:
+            //   - assign a new runway and have the SID leg regenerated
+            //   - assign a new SID and have some way to prevent aircraft from
+            //     attempting to use it from their current expected runway when it's
+            //     invalid for that procedure
```
n8rzz: This is a fantastic comment! 🙏

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -1357,6 +1373,42 @@ export default class Fms {
...
+     * Regenerate SID leg based on current runway assignment
+     *
+     * @for Fms
+     * @method _regenerateSidLeg
+     */
+    _regenerateSidLeg() {
+        const sidLegIndex = this._findLegIndexForProcedureType(PROCEDURE_TYPE.SID);
+
+        if (sidLegIndex === INVALID_INDEX) {
+            return;
```
n8rzz: If we make it to this method and we can't find a sid leg, is that a problem?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -1357,6 +1373,42 @@ export default class Fms {
...
+     * Regenerate STAR leg based on current runway assignment
+     *
+     * @for Fms
+     * @method _regenerateStarLeg
+     */
+    _regenerateStarLeg() {
+        const starLegIndex = this._findLegIndexForProcedureType(PROCEDURE_TYPE.STAR);
+
+        if (starLegIndex === INVALID_INDEX) {
+            return;
```
n8rzz: Same as above

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -137,13 +138,25 @@ ava('.init() calls ._buildLegCollection()', (t) => {
...
-    const nextRunwayMock = airportModelFixture.getRunway('19R');
+ava('.setDepartureRunway() sets #departureRunwayModel to the specified runway model', (t) => {
+    const nextRunwayFixture = airportModelFixture.getRunway('19R');
     const fms = buildFmsMockForDeparture();
 
-    fms.setDepartureRunway(nextRunwayMock);
+    fms.setDepartureRunway(nextRunwayFixture);
 
-    t.true(_isEqual(fms.departureRunwayModel, nextRunwayMock));
+    t.true(_isEqual(fms.departureRunwayModel, nextRunwayFixture));
```
n8rzz: I've actually started using `t.deepEqual()` is instead of lodash. Not something you need to change but something to keep in mind.

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -477,15 +510,7 @@ ava('.replaceDepartureProcedure() replaces the currentLeg with the new route', (
...
-ava('.replaceArrivalProcedure() returns early if the nextRouteString matches the current route', (t) => {
-    const routeStringMock = ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK.route;
-    const fms = buildFmsMock();
-    const _findLegIndexForProcedureTypeSpy = sinon.spy(fms, '_findLegIndexForProcedureType');
-
-    fms.replaceArrivalProcedure(routeStringMock, runwayAssignmentMock);
-
-    t.false(_findLegIndexForProcedureTypeSpy.called);
-});
+ava.todo('.replaceArrivalProcedure() updates the current runway assignment');
```
n8rzz: Is this todo still valid?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: It would be a problem, because we're operating under the assumption that it already exists, and then we find it and regenerate it with the new runway.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: same answer here

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: easier anyway, fewer imports! :tada:

test/aircraft/FlightManagementSystem/Fms.spec.js
erikquinn: it would be still valid, a separate thing to be done, like with what we'd need to do to fix #267.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
n8rzz: If that's the case we should probably throw here instead of just returning.

erikquinn: @n8rzz what kind of error do you think?

erikquinn: i guess just a `TypeError`, since it's not of the expected type. I'll do that.

n8rzz: :shipit:

erikquinn: added comment on why we can't throw here.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: added comment on why we can't throw here.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -1357,6 +1373,46 @@ export default class Fms {
...
+     * @for Fms
+     * @method _regenerateSidLeg
+     */
+    _regenerateSidLeg() {
+        const sidLegIndex = this._findLegIndexForProcedureType(PROCEDURE_TYPE.SID);
+
+        if (sidLegIndex === INVALID_INDEX) {
+            // would be throwing here, except that causes the initial setting of the
+            // departure runway to throw. So instead will just return early.
+            return;
```
n8rzz: 👍

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -1357,6 +1373,46 @@ export default class Fms {
...
+     * @for Fms
+     * @method _regenerateStarLeg
+     */
+    _regenerateStarLeg() {
+        const starLegIndex = this._findLegIndexForProcedureType(PROCEDURE_TYPE.STAR);
+
+        if (starLegIndex === INVALID_INDEX) {
+            // would be throwing here, except that causes the initial setting of the
+            // arrival runway to throw. So instead will just return early.
+            return;
```
n8rzz: 👍

PR #553, opened 2017-07-04
refactor/480

### LINKED ISSUE #480: Consolidate warnings for missing fixes in airport files
The sim currently puts warnings into the console for any fixes that are being used but have not been defined in the `fixes` section of the airport json file. They show up as separate entries, and often multiple times. If they could instead be put into an array that clears duplicate elements and sorts, that would be a much easier thing to work with. For example, I copy and paste that list into skyvector (since its waypoint database seems to be the most comprehensive), use the nav log thing to generate a pdf including the coordinates, and then sublime to reformat them as needed.

image

Related: #94.

erikquinn:
Resolves #480.

Consolidate console warnings for missing fixes to single message

### DIFF
--- a/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
+++ b/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
@@ -1,3 +1,4 @@
+import _filter from 'lodash/filter';
 import _isNil from 'lodash/isNil';
 import _uniq from 'lodash/uniq';
 import StaticPositionModel from '../base/StaticPositionModel';
@@ -351,12 +352,16 @@ export default class NavigationLibrary {
      * @private
      */
     _getAllFixNamesInUse() {
-        const sidFixes = this.sidCollection.getAllFixNamesInUse();
-        const starFixes = this.starCollection.getAllFixNamesInUse();
-        const allFixNames = _uniq(sidFixes.concat(starFixes)).sort();
-        const allNonVectorFixes = allFixNames.filter((fix) => fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
+        const allFixNames = [
+            ...this.sidCollection.getAllFixNamesInUse(),
+            ...this.starCollection.getAllFixNamesInUse()
+        ];
+        const uniqueFixNames = _uniq(allFixNames);
+        const allNonVectorFixes = _filter(uniqueFixNames, (fix) => {
+            return fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX;
+        });
 
-        return allNonVectorFixes;
+        return allNonVectorFixes.sort();
     }
 
 }
--- a/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
+++ b/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
@@ -332,7 +332,7 @@ export default class NavigationLibrary {
      * @method showConsoleWarningForUndefinedFixes
      */
     showConsoleWarningForUndefinedFixes() {
-        const allFixNames = this._getAllFixNames();
+        const allFixNames = this._getAllFixNamesInUse();
         const missingFixes = allFixNames.filter((fix) => !FixCollection.findFixByName(fix));
 
         if (missingFixes.length < 1) {
@@ -346,13 +346,13 @@ export default class NavigationLibrary {
      * Gathers a unique, sorted list of all fixes used in all known procedures
      *
      * @for NavigationLibrary
-     * @method _getAllFixNames
+     * @method _getAllFixNamesInUse
      * @return {array<string>} ['fixxa', 'fixxb', 'fixxc', ...]
      * @private
      */
-    _getAllFixNames() {
-        const sidFixes = this.sidCollection.getAllFixNames();
-        const starFixes = this.starCollection.getAllFixNames();
+    _getAllFixNamesInUse() {
+        const sidFixes = this.sidCollection.getAllFixNamesInUse();
+        const starFixes = this.starCollection.getAllFixNamesInUse();
         const allFixNames = _uniq(sidFixes.concat(starFixes)).sort();
         const allNonVectorFixes = allFixNames.filter((fix) => fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
 
--- a/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
+++ b/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
@@ -296,12 +296,15 @@ export default class NavigationLibrary {
                 route = this._sidCollection.findRouteByIcao(routeString);
 
                 break;
-            case PROCEDURE_TYPE.STAR:
+
+            case PROCEDURE_TYPE.STAR: {
                 const { procedure } = new RouteModel(routeString);
 
                 route = this._starCollection.findRouteByIcao(procedure);
 
                 break;
+            }
+
             default:
                 return false;
         }
@@ -357,9 +360,8 @@ export default class NavigationLibrary {
             ...this.starCollection.getAllFixNamesInUse()
         ];
         const uniqueFixNames = _uniq(allFixNames);
-        const allNonVectorFixes = _filter(uniqueFixNames, (fix) => {
-            return fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX;
-        });
+        const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
+            fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
 
         return allNonVectorFixes.sort();
     }
--- a/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
+++ b/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
@@ -1,4 +1,5 @@
 import _isNil from 'lodash/isNil';
+import _uniq from 'lodash/uniq';
 import StaticPositionModel from '../base/StaticPositionModel';
 import RouteModel from './Route/RouteModel';
 import FixCollection from './Fix/FixCollection';
@@ -8,6 +9,8 @@ import {
     FLIGHT_PHASE,
     PROCEDURE_TYPE
 } from '../constants/aircraftConstants';
+import { VECTOR_WAYPOINT_PREFIX } from '../constants/navigation/routeConstants';
+import { INVALID_INDEX } from '../constants/globalConstants';
 
 /**
  *
@@ -106,8 +109,11 @@ export default class NavigationLibrary {
         );
 
         FixCollection.addItems(fixes, this._referencePosition);
+
         this._sidCollection = new StandardRouteCollection(sids, PROCEDURE_TYPE.SID);
         this._starCollection = new StandardRouteCollection(stars, PROCEDURE_TYPE.STAR);
+
+        this.showConsoleWarningForUndefinedFixes();
     }
 
     /**
@@ -316,4 +322,41 @@ export default class NavigationLibrary {
             flightPhase === FLIGHT_PHASE.TAXI ||
             flightPhase === FLIGHT_PHASE.WAITING;
     }
+
+    /**
+     * Check all fixes used in procedures, and gather a list of any fixes that are
+     * not defined in the `fixes` section of the airport file, then sort and print
+     * that list to the console.
+     *
+     * @for NavigationLibrary
+     * @method showConsoleWarningForUndefinedFixes
+     */
+    showConsoleWarningForUndefinedFixes() {
+        const allFixNames = this._getAllFixNames();
+        const missingFixes = allFixNames.filter((fix) => !FixCollection.findFixByName(fix));
+
+        if (missingFixes.length < 1) {
+            return;
+        }
+
+        console.warn(`The following fixes have yet to be defined in the "fixes" section: ${missingFixes}`);
+    }
+
+    /**
+     * Gathers a unique, sorted list of all fixes used in all known procedures
+     *
+     * @for NavigationLibrary
+     * @method _getAllFixNames
+     * @return {array<string>} ['fixxa', 'fixxb', 'fixxc', ...]
+     * @private
+     */
+    _getAllFixNames() {
+        const sidFixes = this.sidCollection.getAllFixNames();
+        const starFixes = this.starCollection.getAllFixNames();
+        const allFixNames = _uniq(sidFixes.concat(starFixes)).sort();
+        const allNonVectorFixes = allFixNames.filter((fix) => fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
+
+        return allNonVectorFixes;
+    }
+
 }
--- a/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
+++ b/src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
@@ -10,8 +10,6 @@ import {
     FLIGHT_PHASE,
     PROCEDURE_TYPE
 } from '../constants/aircraftConstants';
-import { VECTOR_WAYPOINT_PREFIX } from '../constants/navigation/routeConstants';
-import { INVALID_INDEX } from '../constants/globalConstants';
 
 /**
  *
@@ -296,7 +294,6 @@ export default class NavigationLibrary {
                 route = this._sidCollection.findRouteByIcao(routeString);
 
                 break;
-
             case PROCEDURE_TYPE.STAR: {
                 const { procedure } = new RouteModel(routeString);
 
@@ -304,7 +301,6 @@ export default class NavigationLibrary {
 
                 break;
             }
-
             default:
                 return false;
         }
@@ -360,8 +356,8 @@ export default class NavigationLibrary {
             ...this.starCollection.getAllFixNamesInUse()
         ];
         const uniqueFixNames = _uniq(allFixNames);
-        const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
-            fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
+        const allNonVectorFixes = _filter(uniqueFixNames, (fixName) =>
+            !RouteModel.isVectorRouteString(fixName));
 
         return allNonVectorFixes.sort();
     }
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
@@ -1,8 +1,8 @@
 import _find from 'lodash/find';
+import _flatten from 'lodash/flatten';
 import _forEach from 'lodash/forEach';
-import _isArray from 'lodash/isArray';
-import _isObject from 'lodash/isObject';
 import _map from 'lodash/map';
+import _uniq from 'lodash/uniq';
 import BaseCollection from '../../base/BaseCollection';
 import RouteSegmentModel from './RouteSegmentModel';
 import { isEmptyObject } from '../../utilities/validatorUtilities';
@@ -138,18 +138,15 @@ export default class RouteSegmentCollection extends BaseCollection {
     }
 
     /**
-     * Returns an array of fix names used by any segment in the collection
+     * Returns an array of unique fix names used by any segment in the collection
      *
      * @for RouteSegmentCollection
      * @method getAllFixNamesInUse
      * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
      */
     getAllFixNamesInUse() {
-        let fixNames = [];
-
-        this._items.forEach((routeSegmentModel) => {
-            fixNames = fixNames.concat(routeSegmentModel.getAllFixNamesInUse());
-        });
+        const fixGroups = this._items.map((routeSegmentModel) => routeSegmentModel.getAllFixNamesInUse());
+        const fixNames = _uniq(_flatten(fixGroups));
 
         return fixNames;
     }
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
@@ -123,6 +123,7 @@ export default class RouteSegmentCollection extends BaseCollection {
         return segment.findWaypointsForSegment();
     }
 
+    // TODO: This gathers names of exit points, not names of fixes. Should be renamed!
     /**
      * Return a list of fixNames for all of the `RouteSegmentModel`s in the collection
      *
@@ -134,6 +135,23 @@ export default class RouteSegmentCollection extends BaseCollection {
      */
     gatherFixNamesForCollection() {
         return _map(this._items, (item) => item.name);
+    }
+
+    /**
+     * Returns an array of fix names used by any segment in the collection
+     *
+     * @for RouteSegmentCollection
+     * @method getAllFixNames
+     * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
+     */
+    getAllFixNames() {
+        let fixNames = [];
+
+        this._items.forEach((routeSegmentModel) => {
+            fixNames = fixNames.concat(routeSegmentModel.getAllFixNames());
+        });
+
+        return fixNames;
     }
 
     /**
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
@@ -141,14 +141,14 @@ export default class RouteSegmentCollection extends BaseCollection {
      * Returns an array of fix names used by any segment in the collection
      *
      * @for RouteSegmentCollection
-     * @method getAllFixNames
+     * @method getAllFixNamesInUse
      * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
      */
-    getAllFixNames() {
+    getAllFixNamesInUse() {
         let fixNames = [];
 
         this._items.forEach((routeSegmentModel) => {
-            fixNames = fixNames.concat(routeSegmentModel.getAllFixNames());
+            fixNames = fixNames.concat(routeSegmentModel.getAllFixNamesInUse());
         });
 
         return fixNames;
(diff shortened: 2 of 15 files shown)


src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +322,21 @@ export default class NavigationLibrary {
             flightPhase === FLIGHT_PHASE.TAXI ||
             flightPhase === FLIGHT_PHASE.WAITING;
     }
+
+    showConsoleWarningForUndefinedFixes() {
+        const allFixNames = this._getAllFixNames();
+        const missingFixes = allFixNames.filter((fix) => !FixCollection.findFixByName(fix));
+
+        console.warn(`The following fixes have yet to be defined in the "fixes" section: ${missingFixes}`);
```
erikquinn: `if (missingFixes.length > 0)`

erikquinn: instead `if (missingFixes.length < 1) {return;}`.

done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +322,41 @@ export default class NavigationLibrary {
...
+
+    /**
+     * Gathers a unique, sorted list of all fixes used in all known procedures
+     *
+     * @for NavigationLibrary
+     * @method _getAllFixNames
+     * @return {array<string>} ['fixxa', 'fixxb', 'fixxc', ...]
+     * @private
+     */
+    _getAllFixNames() {
```
n8rzz: Could we call this something like `_getAllFixNamesInUse()`?  Or somehow make me t clear that these fixes are used?

src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
```diff
@@ -137,6 +138,23 @@ export default class RouteSegmentCollection extends BaseCollection {
...
+     * Returns an array of fix names used by any segment in the collection
+     *
+     * @for RouteSegmentCollection
+     * @method getAllFixNames
+     * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
+     */
+    getAllFixNames() {
+        let fixNames = [];
+
+        this._items.forEach((routeSegmentModel) => {
```
n8rzz: Why not use `_map()` here and set the result to `fixNames`

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteCollection.js
```diff
@@ -212,6 +212,23 @@ export default class StandardRouteCollection extends BaseCollection {
...
+     *
+     * @for StandardRouteCollection
+     * @method getAllFixNames
+     * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
+     */
+    getAllFixNames() {
+        let fixNames = [];
+
+        this._items.forEach((standardRouteModel) => {
+            fixNames = fixNames.concat(standardRouteModel.getAllFixNames());
```
n8rzz: Why not use `_map()` here and set the result to `fixNames`.  Then youcan use `const` as well.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -291,6 +291,24 @@ export default class StandardRouteModel extends BaseModel {
     }
 
     /**
+     * Returns an array of fix names used by any portion of the procedure
+     *
+     * @for StandardRouteModel
+     * @method getAllFixNames
+     * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
+     */
+    getAllFixNames() {
```
n8rzz: I have thoughts on this. It I need a computer to show you. In the mean time, why not `_flatten()` for the draw section?

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
```diff
@@ -53,13 +54,21 @@ ava('.findWaypointsForSegmentName() returns an array of fixes for a given segmen
...
+ava('.gatherFixNamesForCollection() returns a list of entryPoint/exitPoint names for each RouteSegmentModel in the collection', t => {
     const collection = new RouteSegmentCollection(ROUTE_SEGMENTS_MOCK);
     const result = collection.gatherFixNamesForCollection();
 
     t.true(result.length === 8);
 });
 
+ava('.getAllFixNames() returns an array of fix names used by any segment in the collection', (t) => {
+    const collection = new RouteSegmentCollection(ROUTE_SEGMENTS_MOCK);
+    const result = _uniq(collection.getAllFixNames().sort());
```
n8rzz: What's the need for _uniq here?

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +322,41 @@ export default class NavigationLibrary {
...
+        const sidFixes = this.sidCollection.getAllFixNames();
```
n8rzz: There are a few blocks below with this same structure, and my suggestion here will apply to those as well.

Instead of `_uniq(thing.concat(otherThing)).sort`, we can do this:
```javascript
const allFixNames = [
  ...this.sidCollection.getAllFixNames(),
  ...this.starCollection.getAllFixNames()
];
const uniqueFixNames = _uniq(allFixNames);
const allNonVectorFixes = _filter(uniqueFixNames, (fix) => fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);

return allNonVectorFixes.sort();
```
This does a few things for us:
- gets rid of `concat` which will mutate an existing array (not desirable).
- gives us an aptly named var `uniqueFixNames` that provides a little more context
- adds `_filter()` which, although there is a native version, we are using lodash all over the place and should use lodash methods when available.  Plus native `filter` isn't supported by some versions of IE (though this argument is not a valid today as it was a year or two ago)

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
n8rzz: see above massive comment that also applies here

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: yeah, that works.

done

src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
erikquinn: Because with map I'll end up with nested arrays rather than a big pile of fixes. I could `.map()` and then `_flatten`?

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteCollection.js
erikquinn: discussed above

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
erikquinn: Just so I could explicitly write out the full list of fixes collected at this point in the following line, and ensure that was the output. I could extend that `expectedResult` to contain all of the duplicated fixes, but that seemed clunky (not that this is a real great answer either if you ask me).

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: love it!

done

erikquinn: Other than `_flatten`, I can't figure out how to redo the other methods. They are different because the spread is not done within an arrow function (where they seem to be illegal)?

erikquinn: Oh wait, I think this can be done with reduce. :thinking:

erikquinn: Sort of. I could do the following, which maybe is better:
```
// BEFORE
getAllFixNamesInUse() {
    let fixNames = [];

    this._items.forEach((routeSegmentModel) => {
        fixNames = fixNames.concat(routeSegmentModel.getAllFixNamesInUse());
    });

    return fixNames;
}

// AFTER
getAllFixNamesInUse() {
    const allFixNames = this._items.map((routeSegmentModel) => routeSegmentModel.getAllFixNamesInUse())
        .reduce((fixList, fixGroup) => fixList.concat(fixGroup));

    return allFixNames;
}
```

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +323,45 @@ export default class NavigationLibrary {
...
+     * @private
+     */
+    _getAllFixNamesInUse() {
+        const allFixNames = [
+            ...this.sidCollection.getAllFixNamesInUse(),
+            ...this.starCollection.getAllFixNamesInUse()
+        ];
+        const uniqueFixNames = _uniq(allFixNames);
+        const allNonVectorFixes = _filter(uniqueFixNames, (fix) => {
+            return fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX;
```
n8rzz: The return is unnecessary here, you can use a direct return without curly braces

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
n8rzz: You're still working too hard here. 

```javascript
const fixNames = _map(this._items, (routeSegmentModel) => {
  return [...routeSegmentModel.getAllFixNamesInUse()];
});
```

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
n8rzz: I ask because you are checking the result based on the return of another function. In these instances, usually, I prefer to have an explicit list to compare against because that won't ever change.  When one of those methods changes and the test fails, you now have two vectors of failure instead of just one. 

I know where you're going here, and it's ok, I just want to call out that slight difference.

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: That still yields a nested array though. There's no difference in result between `routeSegmentModel.getAllFixNamesInUse()` and `[...routeSegmentModel.getAllFixNamesInUse()]`. :confused:

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: :+1: It was only to break the line for the line-length lint warning. Should I disable the warning for this line, or leave it as shown?

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
n8rzz: Hmm 🤔 

Not having a laptop is a real pain. Flatten may be the way to go then.

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
n8rzz: You can put the part after the arrow on the next line.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteCollection.js
erikquinn: `_uniq(_flatten(_map))`

done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: expanded into a new array with spread operators to combine everything, then uniq-ed it at the end.

done

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
erikquinn: After thinking about it more, the function itself should be returning a unique array anyway, so amended it as such, and removed the uniq from the test, while leaving the unique `expectedResults` (effectively verifying it the same way).

done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: Very good. Used flatten to get it working.

done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: Would it be A or B?
```javascript
// A
const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
    fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);

// B
const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
    fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX
);
```

n8rzz: @erikquinn `A`. option `B` has incorrect syntax.

I think you were getting at the semi=colon part of it, though.  there should be a semi-colon at the end (but also the missing parenthesis)

src/assets/scripts/client/navigationLibrary/StandardRoute/RouteSegmentCollection.js
erikquinn: done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: they're both valid, but yes, that was the question. I'll move the `);` back onto the end of the same line.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -291,6 +291,25 @@ export default class StandardRouteModel extends BaseModel {
...
 
     /**
+     * Returns an array of unique fix names used by any portion of the procedure
+     *
+     * @for StandardRouteModel
+     * @method getAllFixNamesInUse
+     * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
+     */
+    getAllFixNamesInUse() {
+        const allFixNames = [
```
n8rzz: what happens if any one of these collections is `undefined`?

I think this would only be a problem during development of a new airport, but this may be an edge case worth writing a test for?

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -324,10 +324,6 @@ export default class StandardRouteWaypointModel extends BaseModel {
         const fixModel = FixCollection.findFixByName(name);
 
         if (!fixModel) {
-            // TODO: This console warn should be done elsewehere, so a single console
-            // message can inform the airport designer of all missing fixes at once.
-            console.warn(`The following fix was not found in the list of fixes for this Airport: ${this.name}`);
-
             return this;
```
n8rzz: do we need to return `this` here or can this be a simple `return`?

test/navigationLibrary/StandardRoute/RouteSegmentModel.spec.js
```diff
@@ -1,6 +1,8 @@
 /* eslint-disable arrow-parens, import/no-extraneous-dependencies*/
 import ava from 'ava';
 import sinon from 'sinon';
+import _isEqual from 'lodash/isEqual';
+import _uniq from 'lodash/uniq';
```
n8rzz: is `_uniq()` still in use within this file?

test/navigationLibrary/StandardRoute/StandardRouteModel.spec.js
```diff
@@ -3,6 +3,7 @@ import sinon from 'sinon';
 import _isEqual from 'lodash/isEqual';
 import _keys from 'lodash/keys';
 import _map from 'lodash/map';
+import _uniq from 'lodash/uniq';
```
n8rzz: is `_uniq()` still in use within this file?

test/navigationLibrary/StandardRoute/StandardRouteModel.spec.js
```diff
@@ -3,6 +3,7 @@ import sinon from 'sinon';
...
 
 import StandardRouteModel from '../../../src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel';
```
n8rzz: as mentioned above, we should add a test for cases where one of the collections is `undefined`.

for setup, you could do:
```javascript
const standardRouteModel = new StandardRouteModel(STAR_LIST_MOCK.GRNPA1);
standardRouteModel._exitCollection = undefined;
```
then run `standardRouteModel.getAllFixNamesInUse()` and see if stuff blows up.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: well all of them will at least return an empty array if it all goes wrong, and they are allowed to have some of them empty anyway. So there should be no throw from trying to do `...undefined` or something because it will always at least be `...[]`, which resolves to `[]`.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: There is no chaining done on this method anyway, so I can just have it return nothing when it's done.

test/navigationLibrary/StandardRoute/RouteSegmentModel.spec.js
erikquinn: Yes, all imports are used.

test/navigationLibrary/StandardRoute/StandardRouteModel.spec.js
erikquinn: Yes, all imports are used.

test/navigationLibrary/StandardRoute/StandardRouteModel.spec.js
erikquinn: done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: added a test that ensures the correct content is returned without throwing even if any of the collections are empty.

done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -289,12 +296,15 @@ export default class NavigationLibrary {
                 route = this._sidCollection.findRouteByIcao(routeString);
 
                 break;
-            case PROCEDURE_TYPE.STAR:
+
```
n8rzz: The empty line is unnecessary

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -289,12 +296,15 @@ export default class NavigationLibrary {
...
+            case PROCEDURE_TYPE.STAR: {
                 const { procedure } = new RouteModel(routeString);
 
                 route = this._starCollection.findRouteByIcao(procedure);
 
                 break;
+            }
+
```
n8rzz: The empty line is unnecessary

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +326,44 @@ export default class NavigationLibrary {
...
+     * @private
+     */
+    _getAllFixNamesInUse() {
+        const allFixNames = [
+            ...this.sidCollection.getAllFixNamesInUse(),
+            ...this.starCollection.getAllFixNamesInUse()
+        ];
+        const uniqueFixNames = _uniq(allFixNames);
+        const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
+            fix.indexOf(VECTOR_WAYPOINT_PREFIX) === INVALID_INDEX);
```
n8rzz: Isn't there a method on the `RouteModel` that does this already?

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
```diff
@@ -316,4 +326,44 @@ export default class NavigationLibrary {
...
+     * @return {array<string>} ['fixxa', 'fixxb', 'fixxc', ...]
+     * @private
+     */
+    _getAllFixNamesInUse() {
+        const allFixNames = [
+            ...this.sidCollection.getAllFixNamesInUse(),
+            ...this.starCollection.getAllFixNamesInUse()
+        ];
+        const uniqueFixNames = _uniq(allFixNames);
+        const allNonVectorFixes = _filter(uniqueFixNames, (fix) =>
```
n8rzz: This is nit-picky,  it instead of `fix` can we use `fixName` here?  With simply `fix` someone could assume that is a reference to the `FixModel`.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -291,6 +291,25 @@ export default class StandardRouteModel extends BaseModel {
...
+            ...this._entryCollection.getAllFixNamesInUse(),
+            ...this._bodySegmentModel.getAllFixNamesInUse(),
+            ...this._exitCollection.getAllFixNamesInUse(),
+            ...this.draw.reduce((list, fix) => list.concat(fix)).map((fixName) => fixName.replace('*', ''))
```
n8rzz: There is a lot going on and n this one line.  This should be set to a var instead of inline like this.

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
```diff
@@ -294,16 +294,12 @@ export default class StandardRouteWaypointModel extends BaseModel {
     }
 
     // TODO: why do we need to clone?
-    // TODO ClonePoisitonFromFix -> clonePositionFromFix
```
n8rzz: It could, but a little extra verbosity never hurts.

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
```diff
@@ -53,13 +54,21 @@ ava('.findWaypointsForSegmentName() returns an array of fixes for a given segmen
...
 
     t.true(result.length === 8);
 });
 
+ava('.getAllFixNamesInUse() returns an array of fix names used by any segment in the collection', (t) => {
+    const collection = new RouteSegmentCollection(ROUTE_SEGMENTS_MOCK);
+    const result = collection.getAllFixNamesInUse().sort();
+    const expectedResult = ['BAKRR', 'BESSY', 'FIXIX', 'HITME', 'JAKER', 'JESJI', 'MDDOG', 'MINEY', 'PIRMD', 'RBELL', 'ROPPR', 'TARRK', 'WASTE'];
+
+    t.true(_isEqual(result, expectedResult));
```
n8rzz: I know we do this all over the place,  but remember you can use Ava native method `t.deepEqual()` here instead of `_isEqual()`

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: nah, that makes sense to me!

done

src/assets/scripts/client/navigationLibrary/NavigationLibrary.js
erikquinn: ah yes! those static methods. good call, i just forgot about them.

done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
erikquinn: changed to:
```
const drawFixNames = this.draw.reduce((list, fix) => list.concat(fix))
    .map((fixName) => fixName.replace('*', ''));

const allFixNames = [
    ...this._entryCollection.getAllFixNamesInUse(),
    ...this._bodySegmentModel.getAllFixNamesInUse(),
    ...this._exitCollection.getAllFixNamesInUse(),
    ...drawFixNames
];
```

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
erikquinn: I guess the point was that it sounds like the name of a method that would take an argument: `.clonePositionFromFix(fixName)`, when it actually does not. It just looks up and clones the position of the FixModel for which the StandardRouteWaypointModel is named. As I look again, it seems more okay now. Removed the note.

test/navigationLibrary/StandardRoute/RouteSegmentCollection.spec.js
erikquinn: :+1: done

src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteWaypointModel.js
n8rzz: 🤔 thats a really good point.  That is certainly a better way to do that because it is a little handily at the moment.  This method is doing too much in that respect, it would be much easier to ask the `FixCollection` for a cloned `PositionModel` instead of _'Give me the fix.  Ok now Mr. Fix, give me a clone of your `PositionModel`.

Maybe add a `TODO` here to update that.  Good callout. 👍

n8rzz:
@erikquinn be sure to fix that failing test before you merge, though 😉

erikquinn:
:+1: all done!

Issue #554, opened 2017-07-05
Issues with EIDW Under v5.2.1


scott-j-smith:
Hi folks,

Congrats on releasing v5.2.1.  I'm still getting used to the new look & feel of the flight strips.  I was playing EIDW (one of the 3 airports not in WIP), and I ran into a couple of issues.

1.  The departure exit point / arrival entry point names in bold cyan text are overlapping and impossible to read (see screen shot 1).  

2.  On departure procedures BAMLI and  NEVRI(AB) the departing aircraft do not clear FL40 before crossing into the restricted area to the north of the airfield, leading to a -200 point penalty.  Is that correct?  I'd expect the procedures to be designed to prevent issues, because of this statement on the Wiki command reference page under the "CVS" command:  _They will climb to their filed cruise altitude, whilst complying with all altitude and speed restrictions posted in the procedure_.    Example:  Flights STK383 and STK935

3. Departing aircraft penalized for leaving the control area without clearance (-25 point penalty).  Is this a new feature that controllers need to be aware of?  Are the changes documented somewhere?  I'm not seeing any new commands in the Wiki page, just one deleted (Fix).   I assumed that once the flight was issued the "CAF" command, and then the "CVS" en route it would proceed on its merry way.  Example:  Flight N749NX (after running a few minutes past the point of the screen shot 1 below)

4. Arriving aircraft given the command to "DVS" get an assigned altitude of 0 in the flight strip (see screen shot 2).  Is this by design?  (Although I did find by accident, because I wasn't paying attention to all the planes and missed some arrivals, that they will follow the arrival procedure on their own without that command anyway.)  Example:  Flight BCY858

Thanks,
Scott

Screen Shot 1:
![image](https://user-images.githubusercontent.com/22403776/27845213-e546445a-60fa-11e7-8758-befce24f275b.png)

Screen Shot 2:
![image](https://user-images.githubusercontent.com/22403776/27845924-48f55b52-6101-11e7-8f71-22460f233d67.png)

Fechulo:
1.  I believe this has to do with the suffixes, which are now actually in use, but I guess they're all being drawn on top of each other because they're all defined as 1 procedure in the airport file.

2. The procedures don't have any altitude restrictions to avoid EID1. The issue is that EID1 is a danger area activated by NOTAM, so it's not normally active (In fact, I just noticed it should go up to 40,000ft not 4,000ft). This area should probably be removed to prevent problems like these. This might apply to some other areas at EIDW.

[...]

PR #561, opened 2017-07-05
bugfix/554

Fechulo:
This only addresses one of the points brought up in #554 but the other ones should probably have separate issues.


n8rzz:
@Fechulo could you create a new issue for this and reference the original?  We can mark that one as a `discussion` and create new issues for each item so we can keep things straight.

PR #573, opened 2017-07-15
feature/572

### LINKED ISSUE #572: Add SVG logo
At the moment, all we have are PNGs and ICOs and such. A good vector image needs to be created so all the various sized images generated from it are consistent and crisp.

erikquinn:
Resolves #572.

Add project emblem SVG


erikquinn:
@n8rzz Well my understanding is that a _logo_ generally has the organization's name incorporated into it, like you would put on letterhead. And what we're dealing with here is a graphic that is symbolic of the organization, which I've been told would be called an emblem.

That said, we should probably also generate proper logos and such, but that can be another project for another day. :wink:

n8rzz:
Typically, in FE, the logo is a symbol that identifies a website which may or may not include text. It's really just semantics, though. I'm fine either way.

PR #574, opened 2017-07-15
bugfix/570

### LINKED ISSUE #570: Aircraft not proceeding direct new fix after full reroute
Sometimes when I give a reroute command with a valid route, the planes to not immediately turn to the first fix in the route.  I have to give the direct (dct) command to make the start the route.  I've seen this 100% of the time in KBOS (command "rr noley..wyane..vocus..zumol").  I also see it playing VHHH (command "rr tamar..atena..hamok..sampu..plove..river..vh512..vh513), however only for arrivals on the STAR that comes in from MANGO to COLEY -- it works fine for arrivals on the other STARs.  I realize VHHH is still in WIP under v5.2.1 so I don't know if that matters or not...

erikquinn:
Resolves #570.

Set MCP heading mode LNAV after full reroute

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -496,8 +496,15 @@ export default class AircraftCommander {
     runReroute(aircraft, data) {
         // TODO: is this .toUpperCase() necessary??
         const routeString = data[0].toUpperCase();
+        const readback = aircraft.pilot.applyNewRoute(routeString, aircraft.initialRunwayAssignment);
 
-        return aircraft.pilot.applyNewRoute(routeString, aircraft.initialRunwayAssignment);
+        // Only change to LNAV mode if the route was applied successfully, else
+        // continue with the previous instructions (whether a heading, etc)
+        if (readback[0]) {
+            aircraft.mcp.setHeadingLnav();
+        }
+
+        return readback;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -496,8 +496,13 @@ export default class AircraftCommander {
     runReroute(aircraft, data) {
         // TODO: is this .toUpperCase() necessary??
         const routeString = data[0].toUpperCase();
+        const readback = aircraft.pilot.applyNewRoute(routeString, aircraft.initialRunwayAssignment);
 
-        return aircraft.pilot.applyNewRoute(routeString, aircraft.initialRunwayAssignment);
+        if (readback[0]) {
```
n8rzz: Let's be sure to add a comment on the _why_ here so we know next time we come to it

erikquinn: done

erikquinn:
@scott-j-smith @indianbhaji @Fechulo no objections or discoveries of situations where this will do something we don't want it to?

scott-j-smith:
No objections.  I even tested the case @n8rzz mentioned by issuing it to a
flight on the ground (KBOS):
caf w 22r
rr irsew..miltt..cyoti..woons
to c 30

When it left the airspace I got the 10 point credit an no errors!

Issue #236, opened 2017-02-16
canvas refactor


n8rzz:
- general canvas refactorings
- fix padding on scale indicator

tedrek:
I'm not an experienced Javascript/HTML5 canvas optimizer so bear that in mind.  I did a couple of branches testing performance, trying to make it multi canvas where the background didn't have to be redrawn every frame.  I couldn't get any significant performance benefits, and performance was significantly worse on my hardware without good 2d/3d accelerated hardware.  This was probably 2014/2015.  I tried both an approach where the canvases were layered and let the browser do the compositing, and a blit type approach where image data was copied from one canvas to another rather than using the drawing primitives.  I also tried an approach where the data blocks were persistent HTML elements, again the performance was worse than what is done currently.  It appeared that compositing the multiple items was really costly compared to just drawing them completely from scratch.  So my strong suggestion would be to do some cheap benchmarking before going to far on actual refactoring for multiple canvases.

n8rzz:
@tedrek are measuring performance of the initial draw?  Or the performance of re-drawing of each subsequent frame _after_ the initial draw?

I would expect the initial draw to be about the same performance as we see now.  However, I would expect the following renders (where we re-draw only the moving elements) to be significantly more performant.

>I'm not an experienced Javascript/HTML5 canvas optimizer so bear that in mind

^^ that part may be biting you a little bit here in that your 'terrain canvas' may actually be re-drawing each frame.

tedrek:
I was measuring sustained frame rate during play as the metric.

I too expected it to be better performing, and I think it may have been in terms of game code time.  However the cost of the browser compositing multiple layered canvases together via transparency was enough higher to outweigh the drawing time saved on my lower end hardware and break even if that on my higher end hardware.  Similiarly the approach I used to copy the already drawn elements from one canvas to another ended up costing more time copying data around than just drawing the elements from scratch in the one canvas.  I was surprised (and disappointed).

The background canvases were definitely not being modified every frame by the game code, only the aircraft/datablock canvas was being redrawn.  The basic algorithm was:
* Load airport
* Create background canvas
* Draw fixes, terrain, etc. on background canvas
* Create aircraft canvas
* Enter main loop
*  -> Clear and redraw aircraft canvas

[...]

PR #576, opened 2017-07-16
refactor/15

### LINKED ISSUE #15: Scope display profiles
Different profiles for scope colors, fonts, etc should be established in order to provide different themes that emulate the different radar systems used (STARS, ARTS, DSR/Host, ERAM, europe-things-i-don't-know-about).

Example items to be defined in a scope display profile (not all will be included in initial work):
 - [ ] brightness of aircraft targets, data blocks, control symbols
 - [ ] brightness of background color
 - [ ] brightness and radius of range rings
 - [ ] brightness of video map markings
 - [ ] center of range rings
 - [ ] default direction of data block leader lines
 - [ ] default length of data block leader lines
 - [ ] default length of PTLs

Related: zlsa/atc#371, zlsa/atc#395, n8rzz/atc#76.
Blocks: #115.
Originally reported in zlsa/atc#716.

erikquinn:
Resolves #15.

Implement scope themes, and changed default theme to blue-based

---

__Work still underway; do not merge.__ :smile:


src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -403,6 +404,11 @@ export default class ConvasController {
         return false;
     }
 
+    canvas_fill_background(cc) {
```
erikquinn: Add docblock

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -824,9 +830,9 @@ export default class ConvasController {
         cc.save();
 
         if (!aircraft.inside_ctr) {
-            cc.fillStyle = this.theme.RADAR_TARGET_OUTSIDE_RANGE;
+            cc.fillStyle = this.theme.HISTORY_DOT_OUTSIDE_RANGE;
```
n8rzz: I'd really love to see us get rid of some of these `else`es by doing something like:
```javascript
let fillStyle = this.theme.HISTORY_DOT_INSIDE_RANGE;

if (!aircraft.inside_ctr) {
  fillStyle = this.theme.HISTORY_DOT_OUTSIDE_RANGE;
}

cc.fillStyle = fillStyle;
```

src/assets/scripts/client/constants/themes/default/theme.js
```diff
@@ -0,0 +1,74 @@
...
+    VIDEO_MAP: COLOR.VIDEO_MAP_STROKE,
+    WIND_VANE: WIND_VANE_THEME.DEFAULT
+};
+
+
+// -----------------------------------------------------------------------------
+//                              NEW SCOPE THEMES
+// -----------------------------------------------------------------------------
+
+const COLOR2 = {
```
n8rzz: `COLOR2`?

erikquinn: temporary :wink: a little bit ago, i had the colors in this same file. Still fiddling with the optimal way to lay it all out. will continue on breaks at work today, leaving shortly.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/constants/themes/default/theme.js
erikquinn: this has since been changed

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: I'm trying to show some restraint in not mucking with everything I see on this PR. That said, I kept an eye out for this sort of thing and made a few small changes with things like this. But you'll see I left some so I could be sure to finish in a timely manner.

erikquinn:
@n8rzz Pausing this to get some feedback on the design of it all. Once we get a consensus, I'll slap together another theme in blue, and we'll be on our way! :wink:

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: 👍

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -404,13 +405,28 @@ export default class ConvasController {
...
+     * Fill entire scope with the background color specified in the current theme
+     *
+     * @for CanvasController
+     * @method canvas_fill_background
+     * @param cc {Object} HTML5 canvas context
+     */
+    canvas_fill_background(cc) {
+        const SCOPE_THEME = this.theme.SCOPE;
+
+        cc.fillStyle = SCOPE_THEME.BACKGROUND;
```
n8rzz: `SCOPE_THEME` is unnecessary here. You can simply use `this.theme.SCOPE.BACKGROUND`

n8rzz: I see this in a few places, same comment applies to those too

erikquinn: done :smiley:

erikquinn:
Updated to accommodate changes from #587 (feature #586).

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -404,6 +405,19 @@ export default class CanvasController {
...
+     * Fill entire scope with the background color specified in the current theme
+     *
+     * @for CanvasController
+     * @method canvas_fill_background
+     * @param cc {Object} HTML5 canvas context
+     */
+    canvas_fill_background(cc) {
+        cc.fillStyle = this.theme.SCOPE.BACKGROUND;
+
+        cc.fillRect(0, 0, this.canvas.size.width, this.canvas.size.height);
```
n8rzz: is it possible to move `this.canvas.size.width` to something like `this.size.width` or maybe even `this.width`?  I worry that these values are intertwined with other classes elsewhere in the project, though, so this may be a big ask.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -610,11 +628,7 @@ export default class CanvasController {
                 -round(UiController.km_to_px(fix.relativePosition[1])) + this.canvas.panY
```
n8rzz: (_I cant comment on the line I want, this is high up as I can go_)

can we refactor the `_forEach` to a native `for`?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -709,14 +723,18 @@ export default class CanvasController {
      * @param aircraft
      */
     canvas_draw_separation_indicator(cc, aircraft) {
-        if (!GameController.shouldUseTrailingSeparator(aircraft)) {
+        if (!GameController.shouldUseTrailingSeparationIndicator(aircraft)) {
             return;
         }
 
         const runway = aircraft.fms.currentRunway;
         const oppositeOfRunwayHeading = runway.angle + Math.PI;
```
n8rzz: is there a method on the `RunwayModel` that does this already?  if there isn't, there should be 😉

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1047,18 +1029,15 @@ k
...
-        // FIXME: How is this different from at line 793?
-        cc.fillStyle = COLOR.LIGHT_SILVER;
-        cc.strokeStyle = COLOR.LIGHT_SILVER;
-        cc.lineWidth = 2;
-
-        // console.time('canvas_draw_all_aircraft')
         for (let i = 0; i < prop.aircraft.list.length; i++) {
+            // color of position dot
+            cc.fillStyle = this.theme.RADAR_TARGET.RADAR_TARGET;
+            cc.lineWidth = 2;
```
n8rzz: should this value be in a constant somewhere?  perhaps also part of the theme?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1115,16 +1094,23 @@ k
             }
```
n8rzz: can you add some `TODO`s to the two `if` blocks above?  Those two need to be looked at and/or refactored to use methods in the input controller.

Actually, didn't we get rid of all the `prop` references?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1135,43 +1121,46 @@ k
...
 
-                cc.translate(
-                    ac_pos[0] + displacements[aircraft.datablockDir][0],
-                    ac_pos[1] + displacements[aircraft.datablockDir][1]
-                );
+            if (datablockDir === INVALID_NUMBER) {
+                datablockDir = this.theme.DATA_BLOCK.LEADER_DIRECTION;
             }
 
+            const leaderLength = this.theme.DATA_BLOCK.LEADER_LENGTH *
```
n8rzz: this smells like it should be abstracted to a method

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1832,4 +1826,25 @@ k
...
+     * This method must remain an arrow function in order to preserve the scope
+     * of `this`, since it is being invoked by an EventBus callback.
+     *
+     * @for CanvasController
+     * @method _setTheme
+     * @param themeName {string}
+     */
+    _setTheme = (themeName) => {
+        if (!_has(THEME, themeName)) {
+            return;
```
n8rzz: what kind of feedback should happen here and who is the target audience for this `return`?  Should this `throw` and break things hard or should this be a `console.error()` and let the developer know they aren't doing things correctly.

I feel there should be some sort of feedback here instead of just a return.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -26,10 +29,24 @@ export const GAME_OPTION_NAMES = {
  */
 export const GAME_OPTION_VALUES = [
     {
+        name: GAME_OPTION_NAMES.THEME,
+        defaultValue: 'default',
+        description: 'Scope Theme',
+        type: 'select',
+        onChangeEvent: EVENT.SET_THEME,
```
n8rzz: is this `onChangeEvent` key new?

**stopping point. will continue from here later tonight**

erikquinn: Yes, currently only used here.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Yeah, that terrifies me a little. And we have that canvas refactor issue already, so I figure it can be dealt with then. But I'll drop a todo at the very least.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: F'sho

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Anything for you!

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: :thinking_face: maybe!

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: I was wondering that myself. `prop` shows up as a lint error because it's not in the file. I figured it might still be used as some weird super-hyper-global sketchy variable business that we were working on trashing. But I haven't really looked into it. Will drop the requested todo.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Yeah, console error sounds good here.

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: thats what I was afraid of, I understand.  a `TODO` is a good step, thanks

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: if anything these can come from the `InputController`. but, yeah, **_all_** the `prop`s should have been removed some time ago.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -26,10 +29,24 @@ export const GAME_OPTION_NAMES = {
...
+                value: 'DEFAULT'
+            }
+        ]
+    },
+    {
         name: GAME_OPTION_NAMES.CONTROL_METHOD,
         defaultValue: 'classic',
         description: 'Control Method',
         type: 'select',
+        onChangeEvent: null,
```
n8rzz: I haven't made it to the implementation yet, but `_noop()` would be a much better choice than `null` here.  you can call `null` as a function but you can call `_noop()`.

src/assets/scripts/client/constants/themes.js
```diff
@@ -0,0 +1,5 @@
+import { THEME_DEFAULT } from './themes/default/theme';
```
n8rzz: I like what you've done here.  This makes it easy to add new themes, keeps the overall code clean, and gives us on single `THEME` export with everything we need.

well done 👍 👍

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
```diff
@@ -0,0 +1,168 @@
+import { COLOR } from '../colors';
+
+// TODO: Instead of individual colors, can we somehow specify an alpha value
```
n8rzz: thats a possibility. or you could define base `rgb` colors and create a function that returns an `rgba` color with a passed opacity value

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
```diff
@@ -0,0 +1,168 @@
...
+// for 'IN_RANGE', 'OUT_OF_RANGE' and 'SELECTED', to be applied to the entire
+// data block? Possibly `cc.globalAlpha`, to be only effective while drawing
+// the data block, then `cc.restore()` to return to normal opacity?
+export const DATA_BLOCK_THEME = {
```
n8rzz: don't forget the doc block here

src/assets/scripts/client/constants/themes/default/modules/scope.js
```diff
@@ -0,0 +1,137 @@
+import { COLOR } from '../colors';
```
n8rzz: just noticed this, we should stay consistent with export/filenames.  In this particular case, I would lean towards singular names. so:
```javascript
import { COLOR } from '../color'; 
```

Its very much a style thing here, but I find it reads better: `COLOR.RED_02` vs `COLORS.RED_02`.  we are only ever dealing with a _single_ value, so it makes sense for it's parent to also be singular.

src/assets/scripts/client/constants/themes/default/modules/scope.js
```diff
@@ -0,0 +1,137 @@
...
+     * Color of the heading text by the major hash marks of the relative compass
+     * headings displayed when an aircraft is selected
+     *
+     * @memberof SCOPE_THEME
+     * @property COMPASS_TEXT
+     */
+    COMPASS_TEXT: COLOR.WHITE,
+
+    /**
+     * Color of the crosshair symbol drawn by `CanvasController.canvas_draw_crosshairs()`
```
n8rzz: There is so much documentation here, I love it!!   This is looking really good!

src/assets/scripts/client/constants/themes/default/modules/scope.js
```diff
@@ -0,0 +1,137 @@
...
+    /**
+     * Color of the runways themselves
+     *
+     * @memberof SCOPE_THEME
+     * @property RUNWAY
+     */
+    RUNWAY: COLOR.WHITE_04,
+
+    /**
+     * Color of the lines showing where SIDs go
```
n8rzz: I think this could be worded a little better.  What does this actually mean?

src/assets/scripts/client/constants/themes/default/modules/scope.js
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @memberof SCOPE_THEME
+     * @property SID
+     */
+    SID: COLOR.BLUE_06,
+
+    /**
+     * Color of the lines showing where STARs go
```
n8rzz: same here, this could be worded better

src/assets/scripts/client/constants/themes/default/modules/terrain.js
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Opacity of the lines on outer edges of terrain-elevated areas
+     *
+     * @memberof TERRAIN_THEME
+     * @property BORDER_OPACITY
+     */
+    BORDER_OPACITY: 1,
+
+    /**
+     * Opaxity of the fill color within terrain-elevated areas
```
n8rzz: opaxity -> opacity

src/assets/scripts/client/constants/themes/default/modules/terrain.js
```diff
@@ -0,0 +1,32 @@
...
+     */
+    FILL_OPACITY: 0.1,
+
+    /**
+     * Colors for each elevation level (in feet), in RGB
+     *
+     * @memberof TERRAIN_THEME
+     * @property COLOR
+     */
+    COLOR: {
```
n8rzz: I feel the inline object here should be split out to its own definition somewhere else.

could you also add a `TODO` here to refactor the consumer(s) of this value to use actual `rgb` values? or at the very least accept `rgb` values and build the data it needs from those.  That way we can stick with the colors defined for a theme and the consumer can use those values.

src/assets/scripts/client/constants/themes/default/modules/windVane.js
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Color of the line indicating the wind direction in gusty winds
+     *
+     * @memberof WIND_VANE_THEME
+     * @property DIRECTION_LINE_GUSTY
+     */
+    DIRECTION_LINE_GUSTY: COLOR.RED_BRIGHT_07,
+
+    /**
+     * Color of the line indivating the wind direction
```
n8rzz: indivating?

src/assets/scripts/client/constants/themes/default/theme.js
```diff
@@ -0,0 +1,18 @@
...
+import { SCOPE_THEME } from './modules/scope';
+import { TERRAIN_THEME } from './modules/terrain';
+import { WIND_VANE_THEME } from './modules/windVane';
+
+/**
+ * @enum THEME_DEFAULT
+ * @type {Object}
+ * @final
+ */
+export const THEME_DEFAULT = {
```
n8rzz: this is so pretty 😌

src/assets/scripts/client/game/GameController.js
```diff
@@ -1,11 +1,14 @@
...
 import { round } from '../math/core';
+import { EVENT } from '../constants/eventNames';
 import { GAME_OPTION_NAMES } from '../constants/gameOptionConstants';
-import { SELECTORS } from '../constants/selectors';
 import { TIME } from '../constants/globalConstants';
+import { SELECTORS } from '../constants/selectors';
+import { THEME } from '../constants/themes';
 
 // Temporary const declaration here to attach to the window AND use as internal property
 const game = {};
```
n8rzz: if there are no more `prop.game` values throughout the app, this line can be removed along with the first line in the `constructor`.

src/assets/scripts/client/game/GameController.js
```diff
@@ -109,6 +115,8 @@ class GameController {
      * @method enable
      */
     enable() {
+        this._eventBus.on(EVENT.SET_THEME, this._setTheme);
```
n8rzz: if there isn't one already, be sure to add a `disable()` method that removes this event listener:
```javascript
disable() {
  this._eventBus.off(EVENT.SET_THEME, this._setTheme);

  return this;
}
```

src/assets/scripts/client/game/GameController.js
```diff
@@ -437,25 +446,59 @@ class GameController {
...
-     * @method shouldUseTrailingSeparator
-     * @param aircraft
+     * @method shouldUseTrailingSeparationIndicator
+     * @param aircraft {AircraftModel}
      * @return {boolean}
      */
-    shouldUseTrailingSeparator(aircraft) {
-        return aircraft.isDeparture() && this.getGameOption(GAME_OPTION_NAMES.DRAW_ILS_DISTANCE_SEPARATOR) === 'yes';
+    shouldUseTrailingSeparationIndicator(aircraft) {
+        const isArrival = aircraft.isArrival();
```
n8rzz: this `const` is unnecessary.  `aircraft.isArrival()` can simply be moved to the `return`

src/assets/scripts/client/game/GameController.js
```diff
@@ -437,25 +446,59 @@ class GameController {
...
+        let isIndicatorEnabled = userSettingsValue === 'yes';
+
+        if (userSettingsValue === 'from-theme') {
+            isIndicatorEnabled = this.theme.RADAR_TARGET.TRAILING_SEPARATION_INDICATOR_ENABLED;
+        }
+
+        return isArrival && isIndicatorEnabled;
     }
+
+    // TODO: Upon removal of `this.getPtlLength()`, this will no longer be needed
```
n8rzz: when is slated to happen?

src/assets/scripts/client/game/GameOptions.js
```diff
@@ -54,6 +56,8 @@ export default class GameOptions {
     }
 
     /**
+     * Gets the value of a given game option
+     *
      * @for GameOptions
      * @method get
```
n8rzz: this makes me cringe. `get` is a reserved word within es2015+.  I know this is legacy code, but we should really update this method name.

src/assets/scripts/client/game/GameOptions.js
```diff
@@ -63,15 +67,21 @@ export default class GameOptions {
...
+     * Sets a game option to a given value
+     *
      * @for GameOptions
      * @method set
-     * @param name {string}
-     * @param value
+     * @param name {string} name of the option to change
+     * @param value {string} value to set the option to
      */
     set(name, value) {
```
n8rzz: same cringe as above

src/assets/scripts/client/game/GameOptions.js
```diff
@@ -63,15 +67,21 @@ export default class GameOptions {
...
         localStorage[`zlsa.atc.option.${name}`] = value;
         this[name] = value;
 
+        if (this._options[name].onChangeEvent) {
```
n8rzz: ref `_noop()` comment from way up.

I would actually advocate for not including the `onChangeEvent` key at all if it isn't needed.  Simply check for the key and then create the handler only if the key is present.

I would also suggest a name change to `onChangeEventHandler`

src/assets/style/command.css
```diff
@@ -1,6 +1,5 @@
-
 #command {
-  background-color: rgba(0,0,0,0.7);
+  background-color: #141414;
```
n8rzz: you asked about vars in css a long time ago.  we could hook up `less` or `sass` and add some css theme vars

erikquinn: Also, I discovered today you can do `$('#elementId').css({ color: 'white', background-color: 'black'});` sort of thing. Though you want jQuery to die. Just wasn't sure if you knew that was a thing (you probably do)...

n8rzz: That is correct, you _can_ do that but that doesn't mean you _should_. These kinds of theme things are much easier to maintain by using CSS classes.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Oh, okay, so actually it's because `this.canvas` is an actual reference to the actual canvas object. I thought it was a dummy property that contained a bunch of things that were updated every cycle (which would be an icky idea for sure).

So are you saying that you _would_ want to add and maintain `CanvasController.width` and `CanvasController.height` from the canvas? That would seem odd to me, but will drop the todo nonetheless if you would like.

n8rzz: no need, I was under the same assumption.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done, with test

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: actually this is reference to an old stroke of the target which was removed, so the linewidth setting can also be removed.

done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: `prop` _is_ defined, and it is being used. It's being defined at `App.js:10`, which make sense as to why it would be globally accessible. Left a todo there. 

done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/constants/gameOptionConstants.js
erikquinn: done

src/assets/scripts/client/constants/themes.js
erikquinn: :tada:

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
erikquinn: we discussed this more as well. will deal with this in the future.

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
erikquinn: added to all theme modules

src/assets/scripts/client/constants/themes/default/modules/scope.js
erikquinn: done

src/assets/scripts/client/constants/themes/default/modules/scope.js
erikquinn: Amended to `Color of the lines drawn between fixes on SIDs`

done

src/assets/scripts/client/constants/themes/default/modules/scope.js
erikquinn: done

src/assets/scripts/client/constants/themes/default/modules/terrain.js
erikquinn: done

src/assets/scripts/client/constants/themes/default/modules/terrain.js
erikquinn: this has been changed around quite a bit since this point (I hadn't really gotten to the terrain yet). However, I did leave it in place as an inline object, as it seemed to make a lot of sense to me to keep it that way.

src/assets/scripts/client/constants/themes/default/modules/windVane.js
erikquinn: done

src/assets/scripts/client/game/GameController.js
erikquinn: there were not, so removed!

done

src/assets/scripts/client/game/GameController.js
erikquinn: done

src/assets/scripts/client/game/GameController.js
erikquinn: done

src/assets/scripts/client/game/GameController.js
erikquinn: not sure, but i'm too exhausted to rip the PTLs out of the game controller now, so sometime in the distant future I guess.

src/assets/scripts/client/game/GameOptions.js
erikquinn: changed to `.getOptionByName(name)`.

done

src/assets/scripts/client/game/GameOptions.js
erikquinn: changed to `.setOptionByName(name)`.

done

src/assets/scripts/client/game/GameOptions.js
erikquinn: discussed, and will leave in place with new name, and `_noop` when no operation is needed.

done

src/assets/scripts/client/airport/runway/RunwayModel.js
```diff
@@ -3,7 +3,10 @@ import BaseModel from '../../base/BaseModel';
 import StaticPositionModel from '../../base/StaticPositionModel';
 import { PERFORMANCE } from '../../constants/aircraftConstants';
 import { INVALID_NUMBER } from '../../constants/globalConstants';
-import { angle_offset } from '../../math/circle';
+import {
+    angle_offset,
+    radians_normalize
```
n8rzz: Not sure if I ever explained _why_ we multi-line these kinds of imports: it reduces git conflicts. If each import has it's own line, the diff is much easier to deal with.  So the rule is any more than one import, multi-line (just like here)

src/assets/scripts/client/airport/runway/RunwayModel.js
```diff
@@ -150,22 +153,33 @@ export default class RunwayModel extends BaseModel {
      *
      * @for RunwayModel
      * @property gps
-     * @return {array<number>} gps coordinates of the runway
+     * @type {array<number>} gps coordinates of the runway
```
n8rzz: 👍

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -603,21 +622,19 @@ export default class CanvasController {
         cc.lineJoin = 'round';
         cc.font = BASE_CANVAS_FONT;
 
-        _forEach(this._navigationLibrary.realFixes, (fix, i) => {
+        // for (const fix in this._navigationLibrary.realFixes) {
+        const fixes = this._navigationLibrary.realFixes;
+        for (let i = 0; i < fixes.length; i++) {
```
n8rzz: we can still use `fix` here, and we should be, but we have to do it manually:
```javascript
for (let i = 0; i < fixes.length; i++) {
  const fix = fixes[i];
}
```
this also gives us a `const` because each iteration through the loop is a new context (thus, a new `const`).

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: is this using that new method?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -797,26 +807,22 @@ export default class CanvasController {
...
         if (!aircraft.inside_ctr) {
-            cc.fillStyle = this.theme.RADAR_TARGET_OUTSIDE_RANGE;
+            cc.fillStyle = this.theme.RADAR_TARGET.HISTORY_DOT_OUTSIDE_RANGE;
         } else {
-            cc.fillStyle = this.theme.RADAR_TARGET_IN_RANGE;
+            cc.fillStyle = this.theme.RADAR_TARGET.HISTORY_DOT_INSIDE_RANGE;
         }
-
         const length = aircraft.relativePositionHistory.length;
         for (let i = 0; i < length; i++) {
```
n8rzz: we should re-assign `aircraft.relativePositionHistory[i]` to some other contextual var here (like `fix` in the other `for` loop above).  We should also give the `0` and `1` indexes an enumerated name, like `x` and `y`.  Otherwise it's a little hard to tell what those actually are.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1071,191 +1048,218 @@ k
...
-            const pi_slice = pi / 24;
-            let match = false;
-            let almost_match = false;
-
-            // Callsign Matching
-            if (prop.input.callsign.length > 1 && aircraft.matchCallsign(prop.input.callsign.substr(0, prop.input.callsign.length - 1))) {
-                almost_match = true;
-            }
+        const cs = aircraft.callsign;
+        const paddingLR = 5;
```
n8rzz: are the magic numbers in this method _themeable_?  or are they part of the default theme?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1071,191 +1048,218 @@ k
...
+        // width of datablock (scales to fit callsign)
+        const width = clamp(1, 5.8 * cs.length) + (paddingLR * 2);
+        const width2 = width / 2;
```
n8rzz: could we use `halfWidth` and `halfHeight` here?  the number in the var doesn't give us much clue about what it actually is.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1071,191 +1048,218 @@ k
...
+        // TODO: remove the if/else in favor of an initial assignment, and update with if condition
+        if (aircraft.inside_ctr) {
+            cc.fillStyle = this.theme.DATA_BLOCK.TEXT_IN_RANGE;
+        } else {
+            cc.fillStyle = this.theme.DATA_BLOCK.TEXT_OUT_OF_RANGE;
+        }
+
+        if (aircraft.trend === 0) {
+            // small dash (symbola font)
+            alt_trend_char = String.fromCodePoint(0x2011);
```
n8rzz: to the best of my knowledge, this is the only place we are using that font.  do you think we could remove this in favor of a smaller font?  Is that even possible with the characters we're using here?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1526,12 +1533,20 @@ k
         cc.save();
         cc.translate(this.canvas.panX, this.canvas.panY);
 
+        // TODO: Remove the jQuery in favor of _each()!
         $.each(airport.terrain || [], (elevation, terrainLevel) => {
```
n8rzz: `$.each()` 😱 

imagine that line reading line: _'Neuman'_

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1526,12 +1533,20 @@ k
...
+            if (elevation < 1000) {
+                console.warn(`${airport.icao}.geojson contains 'terrain' at or` +
+                    ' below sea level, which is not supported!');
+                // within `$.each()`, this return acts like `continue;`
+                return;
```
n8rzz: need an empty line above `return`

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1526,12 +1533,20 @@ k
...
+            }
+
             max_elevation = Math.max(max_elevation, elevation);
-            const color = `rgba(${UiController.terrain.COLOR[elevation]}, `;
+            const color = `hsla(${UiController.terrain.COLOR[elevation]}`;
```
n8rzz: why the move to `hsla`?  `hsla` tends to be a non-standard color format for the front-end.  it works, it's just not used very often in my experience.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -26,10 +29,28 @@ export const GAME_OPTION_NAMES = {
  */
 export const GAME_OPTION_VALUES = [
     {
+        name: GAME_OPTION_NAMES.THEME,
+        defaultValue: 'default',
+        description: 'Scope Theme',
+        type: 'select',
+        onChangeEventHandler: EVENT.SET_THEME,
```
n8rzz: 👍

src/assets/scripts/client/constants/themes/classic/modules/windVane.js
```diff
@@ -1,13 +1,9 @@
-import { COLOR } from './colors';
+import { COLOR } from '../color';
```
n8rzz: 👍

src/assets/scripts/client/game/GameController.js
```diff
@@ -117,6 +122,8 @@ class GameController {
      * @method disable
      */
     disable() {
+        this._eventBus.on(EVENT.SET_THEME, this._setTheme);
```
n8rzz: this should be `this._eventBus.off(EVENT.SET_THEME, this._setTheme);`

test/airport/runway/RunwayModel.spec.js
```diff
@@ -45,6 +45,14 @@ ava('#elevation returns #airportPositionModel.elevation if #_positionModel.eleva
     t.true(model.elevation === model.airportPositionModel.elevation);
 });
 
+ava.only('#oppositeAngle returns opposite of runway heading', (t) => {
```
n8rzz: be sure to remove the `ava.only()` here

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: :man_facepalming: it is now...

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: added `const positionHistory` to add some clarity. I agree we could use x and y properties rather than a 2 element array here, but the convention absolutely everywhere else is the 2 element array, and until they are all dealt with in #236, it would only serve to make the canvas controller more difficult to work with by fixing that in only one place.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: They are still hard coded. I pulled out the important things in a couple areas, but there are still some things that have not been theme-ified like this one.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Just went ahead and removed this code here, the font file and all use of the font (as well as from font.css). We don't need the arrows yet anyway. We'll find a better way to do it.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/game/GameController.js
erikquinn: done

test/airport/runway/RunwayModel.spec.js
erikquinn: oopsie! done

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -248,6 +249,8 @@ export default class CanvasController {
 
                 this.canvas_clear(cc);
 
+                this.canvas_fill_background(cc);
```
n8rzz: there should not be an empty line above this method call

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -603,21 +622,18 @@ export default class CanvasController {
         cc.lineJoin = 'round';
         cc.font = BASE_CANVAS_FONT;
 
-        _forEach(this._navigationLibrary.realFixes, (fix, i) => {
+        const fixes = this._navigationLibrary.realFixes;
```
n8rzz: this assignment is unnecessary.  we can simply use `this._navigationLibrary.realFixes` within the for declaration and when assigning `fix`.

If we _did_ need it, we should actually use destructuring here: 
```javascript
const { realFixes } = this._navigationLibrary;
```

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -603,21 +622,18 @@ export default class CanvasController {
...
+        for (let i = 0; i < fixes.length; i++) {
+            const fix = fixes[i];
```
n8rzz: 👍

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: ... from Seinfeld.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1526,12 +1526,21 @@ k
...
+                console.warn(`${airport.icao}.geojson contains 'terrain' at or` +
+                    ' below sea level, which is not supported!');
+
+                // within `$.each()`, this return acts like `continue;`
+                return;
+            }
+
             max_elevation = Math.max(max_elevation, elevation);
-            const color = `rgba(${UiController.terrain.COLOR[elevation]}, `;
+            const color = `hsla(${this.theme.TERRAIN.COLOR[elevation]}`;
```
n8rzz: could you add a comment here on _why_ we're using `hsla`?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1526,12 +1526,21 @@ k
...
 
-            cc.strokeStyle = `${color} ${UiController.terrain.BORDER_OPACITY})`;
-            cc.fillStyle = `${color} ${UiController.terrain.FILL_OPACITY})`;
+            cc.strokeStyle = `${color}, ${this.theme.TERRAIN.BORDER_OPACITY})`;
```
n8rzz: this (and the other uses) should be abstracted to a helper function in another file.

```javascript
// utilities/colorStyleUtilities.js
// or some better place
const generateRgbaStr = (hex, opacity) => `rgba(${color}, ${this.theme.TERRAIN.BORDER_OPACITY})`;
```

that then consolidates the logic for this in a single place so if we ever decide to change how this happens, its easier to update.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: the idea was that since the terrain is many layers deep, hsla would make it easier to create a progressive lightness shift or adjust the overall perceived opacity while maintaining the same hue. That could be done in rgb, but would require back and forth with a web color thing to make the adjustments, while tweaking the h/s/l meant a more convenient way to get the terrain looking the way you wanted.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn:
```
const generateRgbStr = (hsl, opacity) => `rgba(${rgb}, ${opacity})`;
```
^ We could do this? But that hardly seems worth the abstraction.

And as for the example you gave, that would only be valid for the drawing of the terrain, otherwise, we'd want to pass the opacity rather than directly referencing it in order for it to be more broadly applicable.

Overall, there are two instances of `rgba( ${color}...` and one instance of `hsla( ${color}...`. What would you like to do?

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: added comment (can remove later if you'd like to see it changed)

done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: @n8rzz bump ^

n8rzz: 🤔 

Hmm, I guess I thought this was in a few more places.  my bad.  I guess we can leave it as-is then.  I'll allow it 😉

erikquinn: okey doke, looks like all other topics have been addressed then. re-requesting! :smile:

erikquinn:
rodga dat! :tada:

PR #584, opened 2017-07-20
feature/536

### LINKED ISSUE #536: Make red guide marks optional
image

When on the ILS, aircraft have a trailing separation indicator, which should be made optional. Good for those unfamiliar with controlling, but distracting to others.

sam-irl:
Resolves #536.

Allows the distance separator behind planes on ILS to be toggled.

### DIFF
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -712,6 +712,7 @@ export default class CanvasController {
         if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
             return;
         }
+        
         if (GameController.game.option.get(GAME_OPTION_NAME.DRAW_ILS_DISTANCE_SEPARATOR) === 'no') {
             return;
         }
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -712,7 +712,7 @@ export default class CanvasController {
         if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
             return;
         }
-        if (GameController.game.option.get('drawIlsDistanceSeparator') === 'no') {
+        if (GameController.game.option.get(GAME_OPTION_NAME.DRAW_ILS_DISTANCE_SEPARATOR) === 'no') {
             return;
         }
 
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -712,6 +712,9 @@ export default class CanvasController {
         if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
             return;
         }
+        if (GameController.game.option.get('drawIlsDistanceSeparator') === 'no') {
+            return;
+        }
 
         const runway = aircraft.fms.currentRunway;
         const oppositeOfRunwayHeading = runway.angle + Math.PI;
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -709,7 +709,7 @@ export default class CanvasController {
      * @param aircraft
      */
     canvas_draw_separation_indicator(cc, aircraft) {
-        if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE || !GameController.shouldUseTrailingSeparator(aircraft)) {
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -712,6 +712,9 @@ export default class ConvasController {
         if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
             return;
         }
+        if (GameController.game.option.get('drawIlsDistanceSeparator') === 'no') {
```
n8rzz: This line should use the `GAME_OPTION_NAMES` constant here instead of it's literal string value.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -712,6 +712,9 @@ export default class ConvasController {
         if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
             return;
         }
```
n8rzz: There should be an empty line between if statements here. 

Actually, I think this would read better if there were a method that n the `GameController` that simply returned a Boolean. That way we could combine the ifs and use an `||` within a single if block with a small single return.

sam-irl: I'm not really following the logic, because as I understand it `aircraft.category` is in `AircraftController`. Anyway, I'll fix the newline now.

n8rzz: @erikquinn do you know what I'm talking about here?  I'm not near a computer and can't type out some examples.

erikquinn:
Please hold this for the time being, reference some potential conflicts with #15, because it partially addresses the issue and I want the two to complement each other correctly.

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: Something like `GameController.shouldUseTrailingSeparatorFor(aircraftModel)` I think is what he's getting at. Some way to combine the two if statements into a single check of a boolean method so it's more concise here.

sam-irl: Ah, I understand you now.

n8rzz: 👍 Hit the nail on the head!

src/assets/scripts/client/game/GameController.js
```diff
@@ -441,6 +441,27 @@ class GameController {
...
+     * Used by the `CanvasController` to determine whether or not to proceed with 
+     * `canvas_draw_separation_indicator`.
+     *
+     * @for GameController
+     * @method shouldUseTrailingSeparator
+     * @param aircraft
+     * @return {boolean}
+     */
+    shouldUseTrailingSeparator(aircraft) {
+        if (this.getGameOption(GAME_OPTION_NAMES.DRAW_ILS_DISTANCE_SEPARATOR) === 'yes') {
```
n8rzz: Without giving you much context, this is a very good first attempt. This is exactly what I was looking for, however, the logic can be simpler. 

Instead of using two ifs, we can simply return the following:
`return this.getGameOption(GAME_OPTION_NAMES.DRAW_ILS_DISTANCE_SEPARATOR) === 'yes'`

This will give us true when yes and false when not yes. Exactly what we need!

n8rzz:
@erikquinn works for me!

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -709,7 +709,7 @@ export default class ConvasController {
      * @param aircraft
      */
     canvas_draw_separation_indicator(cc, aircraft) {
-        if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE) {
+        if (aircraft.category === FLIGHT_CATEGORY.DEPARTURE || !GameController.shouldUseTrailingSeparator(aircraft)) {
```
n8rzz: Marvelous!

erikquinn: @n8rzz you wouldn't want to combine the category in the method as well?

n8rzz: Good point. One of those methods may already exist on the `AircraftModel`

erikquinn: yeah, `AircraftModel.isDeparture()` and `AircraftModel.isArrival()`. I'll just make that change and merge then. :shipit:

PR #585, opened 2017-07-21
documentation/146

### LINKED ISSUE #146: Glossary for aviation and ATC terminology
> @n8rzz:
>
>There is a lot of aviation-specific language used throughout the app.  It may be helpful to have a dictionary or glossary of those terms with quick definitions and links to more info for the aviation novice.
>
>Some words/terms that should be included:
>- preSpawn
>- spawnPattern
>- StandardRoute
>- RouteString
>- waypoint
>- fix 
>- ILS
>- airspace
>- runway naming patterns
>- flight level, msl, agl altitudes
>
>_there will be more (and I'll add them here as I run into them). these are the ones I keep running into that I feel need some documentation_

Originally reported in n8rzz/atc#247.

sam-irl:
Adds a glossary of aviation terms, resolves #146.

### DIFF
--- a/documentation/glossary.md
+++ b/documentation/glossary.md
@@ -1,0 +1,30 @@
+# Glossary of aviation terms
+*Openscope uses some terms specific to aviation, this glossary attempts to define them.*
+---
+## ILS
+*abbr. of* Instrument Landing System
+
+A radio antenna at the end of the runway, known as a ''localizer'', guides pilots down onto the runway for landing, via an instrument known as a *guide scope*.
+
+## Fix
+
+A pre-defined point that describes an aircraft's current position. A fix can be an airport (`EGKK`), a waypoint (`WILLO`), or other points, such as a VOR station (`GWC`).
+
+## Waypoint
+
+A fix that is used to aid in navigation, defined in a flight plan or route. Waypoints are assigned a 5-letter code, such as `WILLO` and `ABTUM`.
+
+## Flight Level
+*abbr.* FL
+
+A plane's altitude, in 100s of feet. For example, a plane at `FL350` is flying at 35,000ft.
+
+## SID
+*abbr. of* Standard Instrument Departure
+
+A route that an aircraft will follow to exit the airport's radar control area. Each SID begins at the airport and terminates at a fix, with waypoints in between.
+
+## STAR
+*abbr. of* Standard Terminal Arrival Route
+
+A route that an aircraft will follow to approach the airport, the inverse of a SID.
--- a/documentation/glossary.md
+++ b/documentation/glossary.md
@@ -28,3 +28,29 @@ A route that an aircraft will follow to exit the airport's radar control area. E
 *abbr. of* Standard Terminal Arrival Route
 
 A route that an aircraft will follow to approach the airport, the inverse of a SID.
+
+## Airspace
+
+A pre-defined three-dimensional area where certain restrictions occur. For example, in Class A airspace, VFR aircraft are not permitted.
+
+## VFR
+*abbr. of* Visual Flight Rules
+
+VFR aircraft do not have instruments enabling them to navigate and land under ILS. As such, they rely on visual clues and ATC instruction to navigate and land.
+
+## IFR
+*abbr. of* Instrument Flight Rules
+
+IFR aircraft **do** have instruments allowing them to navigate and land using ILS. IFR aircraft have flight plans that direct them to waypoints.
+
+## Route
+
+A set of waypoints that form a path to the destination. In openScope, the route is changed with the `route` and `reroute` commands.
+
+## Crosswind
+
+Wind blowing at an aircraft from the side. For example, for an aircraft flying north, crosswind is wind blowing west to east or vice versa. This can create dangerous landing conditions.
+
+## Tailwind
+
+Wind blowing at an aircraft from the tail. In a north-flying aircraft, this is wind from south to north.


documentation/glossary.md
```diff
@@ -0,0 +1,56 @@
...
+A pre-defined point that describes an aircraft's current position. A fix can be an airport (`EGKK`), a waypoint (`WILLO`), or other points, such as a VOR station (`GWC`).
+
+## Waypoint
+
+A fix that is used to aid in navigation, defined in a flight plan or route. Waypoints are assigned a 5-letter code, such as `WILLO` and `ABTUM`.
+
+## Flight Level
+*abbr.* FL
+
+A plane's altitude, in 100s of feet. For example, a plane at `FL350` is flying at 35,000ft.
```
n8rzz: This should include details on when this notation is used in regards to altitudes.

documentation/glossary.md
```diff
@@ -0,0 +1,56 @@
+# Glossary of aviation terms
+*Openscope uses some terms specific to aviation, this glossary attempts to define them.*
+---
+## ILS
+*abbr. of* Instrument Landing System
+
+A radio antenna at the end of the runway, known as a ''localizer'', guides pilots down onto the runway for landing, via an instrument known as a *guide scope*.
```
indianbhaji: This isn't exactly correct.

The localiser is used to line up with the runway for landing
The glideslope is used to descend an aircraft for landing

documentation/glossary.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Flight Level
+*abbr.* FL
+
+A plane's altitude, in 100s of feet. For example, a plane at `FL350` is flying at 35,000ft.
+
+## SID
+*abbr. of* Standard Instrument Departure
+
+A route that an aircraft will follow to exit the airport's radar control area. Each SID begins at the airport and terminates at a fix, with waypoints in between.
```
indianbhaji: Could also be vectors

documentation/glossary.md
```diff
@@ -0,0 +1,56 @@
...
+
+## STAR
+*abbr. of* Standard Terminal Arrival Route
+
+A route that an aircraft will follow to approach the airport, the inverse of a SID.
```
indianbhaji: @erikquinn normally this takes an aircraft to an IAF, should this be defined?

documentation/glossary.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Airspace
+
+A pre-defined three-dimensional area where certain restrictions occur. For example, in Class A airspace, VFR aircraft are not permitted.
+
+## VFR
+*abbr. of* Visual Flight Rules
+
+VFR aircraft do not have instruments enabling them to navigate and land under ILS. As such, they rely on visual clues and ATC instruction to navigate and land.
```
indianbhaji: I wouldn't exactly involve aircraft in this, but up to @erikquinn since they are a set of _rules_, instead I think this should be based of the rules, since VFR aircraft can have instruments allowing them to navigate

documentation/glossary.md
sam-irl: The glideslope is, the glide**scope** is the instrument.

documentation/glossary.md
sam-irl: I'm going to remove this, it was included because I think there is an issue to add VFR aircraft

documentation/glossary.md
indianbhaji: It really isn't. See https://en.m.wikipedia.org/wiki/Instrument_landing_system.

An aircraft approaching a runway is guided by the ILS receivers in the aircraft by performing modulation depth comparisons. Many aircraft can route signals into the autopilot to fly the approach automatically. An ILS consists of two independent sub-systems. **The localiser provides lateral guidance; the glide slope provides vertical guidance.**

erikquinn: @indianbhaji has it. :+1:

Maybe we can do:
>Runways may have an associated Instrument Landing System which provides arriving aircraft with approach alignment information. There are two independent components: the _localizer_ (which provides lateral guidance), and the _glide slope_ (which provides vertical guidance).

documentation/glossary.md
erikquinn: Let's go for something more generic here to cover all bases.

> An instrument procedure used to guide aircraft from a departure airport out of that airport's airspace.

And then add a term _instrument procedure_:

> A charted series of instructions guiding aircraft along a specific path. This could be a series of headings to fly until passing certain altitudes, or a series of fixes to follow. Altitude and speed restrictions may often apply at certain points along the procedure as well. These are used by air traffic control to organize traffic on a large scale, and reduce the load on controllers under normal operations.

documentation/glossary.md
erikquinn: Yessir! Let's do `IAF`, `FAF`, `MAP`, `IAP` as well to cover the approach plate department.

documentation/glossary.md
erikquinn: Nah, let's keep it. Might as well have information distinguising IFR/VFR. The VFR guys _may_ not have the fancy equipment, but just about all of them can shoot an ILS if they want to. It's as @indianbhaji says, just a set of rules they're following (VFR allows for more freedom and variation, provided they stay out of the way of stuff).

> A set of flight rules under which aircraft may fly visually, rather than depending completely on flight instruments. Because visibility is obviously required, VFR is only permitted when the cloud ceiling is at least 1,000 feet above ground level, and the visbility at least 3 statute miles, as well as remaining clear of clouds. Under VFR, aircraft do not have a flight plan filed with air traffic control, and therefore they may fly wherever they want, provided that they are talking to the right controllers when entering airspace of an elevated classification. They can therefore choose to spin in circles, fly along rivers and highways, fly direct to their destination, or follow a route just like an IFR aircraft would.

documentation/glossary.md
erikquinn: And to do so, we will need to add the term `Transition Altitude` if it doesn't already exist.

documentation/glossary.md
felixscheffer: done

documentation/glossary.md
felixscheffer: done

documentation/glossary.md
felixscheffer: done

PR #592, opened 2017-07-22
feature/564

### LINKED ISSUE #564: Right click and drag to move scope view point
Simple as it sounds. Right click and drag to move the center of the screen around, instead of left click and drag. This will prevent accidental movement of the screen when selecting aircraft, and is consistent with VATSIM ATC clients, so many users will be familiar with this.

sam-irl:
Issue #564  proposes to make the radar drag screen button right click. I propose making the button configurable in the options menu, with right being the default. Input is welcome. Resolves openscope/openscope#564

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Add `sa`, `saa`, `sh`, `sah`, `ss`, `sas` commands [#641](https://github.com/openscope/openscope/issues/641)
 - Add toggleable scope command bar, and lays foundation for the Scope, its commands, and its collections of radar targets. [#14](https://github.com/openscope/openscope/issues/14)
 - Checks if the airport in localStorage exists before loading it [#709](https://github.com/openscope/openscope/issues/709)
+- The mouse button to drag the radar screen is configurable, and defaults to 'right click' [#564](https://github.com/openscope/openscope/issues/564)
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Add `sa`, `saa`, `sh`, `sah`, `ss`, `sas` commands [#641](https://github.com/openscope/openscope/issues/641)
 - Add toggleable scope command bar, and lays foundation for the Scope, its commands, and its collections of radar targets. [#14](https://github.com/openscope/openscope/issues/14)
 - Checks if the airport in localStorage exists before loading it [#709](https://github.com/openscope/openscope/issues/709)
-- The mouse button to drag the radar screen is configurable, and defaults to 'right click' [#564](https://github.com/openscope/openscope/issues/564)
+- The mouse button to drag the radar screen is now right click [#564](https://github.com/openscope/openscope/issues/564)
 
 
 
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -37,6 +37,7 @@ export default class InputController {
      */
     constructor($element, aircraftCommander, aircraftController, scopeModel, tutorialView) {
         this.$element = $element;
+        this.$body = null;
         this.$window = null;
         this.$commandInput = null;
         this.$canvases = null;
@@ -63,9 +64,7 @@ export default class InputController {
         this.input.isMouseDown = false;
         this.commandBarContext = COMMAND_CONTEXT.AIRCRAFT;
 
-        // TODO: Do we want setupHandlers to be `enableHandlers`, and called from `.enable()`?
         this._init()
-            .setupHandlers()
             .enable();
     }
 
@@ -74,7 +73,7 @@ export default class InputController {
      * @method _init
      */
     _init() {
-        this.$body = $('body')[0];
+        this.$body = this.$element[0];
         this.$window = $(window);
         this.$commandInput = this.$element.find(SELECTORS.DOM_SELECTORS.COMMAND);
         this.$canvases = this.$element.find(SELECTORS.DOM_SELECTORS.CANVASES);
@@ -84,30 +83,8 @@ export default class InputController {
     }
 
     /**
-     * Enable handlers
+     * Enable all event handlers
      *
-     * @for InputController
-     * @method setupHandlers
-     */
-    setupHandlers() {
-        this.$body.addEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
-
-        return this;
-    }
-
-    /**
-     * Disable handlers
-     *
-     * @for InputController
-     * @method setupHandlers
-     */
-    disableHandlers() {
-        this.$body.removeEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
-
-        return this;
-    }
-
-    /**
      * @for InputController
      * @method enable
      */
@@ -122,6 +99,7 @@ export default class InputController {
         this.$canvases.on('mousemove', (event) => this.onMouseMoveHandler(event));
         this.$canvases.on('mouseup', (event) => this.onMouseUpHandler(event));
         this.$canvases.on('mousedown', (event) => this.onMouseDownHandler(event));
+        this.$body.addEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
 
         // TODO: Fix this
         this._eventBus.on(EVENT.STRIP_CLICK, this.selectAircraftByCallsign);
@@ -130,6 +108,8 @@ export default class InputController {
     }
 
     /**
+     * Disable all event handlers and destroy the instance
+     *
      * @for InputController
      * @method disable
      */
@@ -141,10 +121,9 @@ export default class InputController {
         this.$canvases.off('mousemove', (event) => this.onMouseMoveHandler(event));
         this.$canvases.off('mouseup', (event) => this.onMouseUpHandler(event));
         this.$canvases.off('mousedown', (event) => this.onMouseDownHandler(event));
+        this.$body.removeEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
 
         this._eventBus.off(EVENT.STRIP_CLICK, this.selectAircraftByCallsign);
-
-        this.disableHandlers();
 
         return this.destroy();
     }
@@ -262,7 +241,7 @@ export default class InputController {
     onMouseDownHandler(event) {
         event.preventDefault();
 
-        // TODO: this should use early returns instead of the else if
+        // TODO: This should use a switch on `event.which` instead of `if/else if`
         if (event.which === MOUSE_EVENT_CODE.MIDDLE_PRESS) {
             UiController.ui_zoom_reset();
         } else if (event.which === MOUSE_EVENT_CODE.RIGHT_PRESS) {
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -173,20 +173,6 @@ export default class InputController {
         this.input.mouseDown = [0, 0];
         this.input.isMouseDown = false;
     }
-    
-    /**
-     * This checks the Mouse click drag option, and returns the appropriate event.
-     * Used in onMouseDownHandler.
-     *
-     * @returns MOUSE_EVENT_CODE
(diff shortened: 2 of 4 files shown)


src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -11,7 +11,7 @@ export const GAME_OPTION_NAMES = {
     PTL_LENGTH: 'ptlLength',
     DRAW_PROJECTED_PATHS: 'drawProjectedPaths',
     SOFT_CEILING: 'softCeiling',
-    INCLUDE_WIP_AIRPORTS: 'includeWipAirports'
+    INCLUDE_WIP_AIRPORTS: 'includeWipAirports',
```
erikquinn: Remove extra comma here

src/assets/scripts/client/InputController.js
```diff
@@ -289,14 +289,16 @@ export default class InputController {
         // TODO: this should use early returns instead of the else if
         if (event.which === MOUSE_EVENT_CODE.MIDDLE_PESS) {
             UiController.ui_zoom_reset();
-        } else if (event.which === MOUSE_EVENT_CODE.LEFT_PRESS) {
+        } else if (event.which === MOUSE_EVENT_CODE.RIGHT_PRESS) {
+		    event.preventDefault();
```
erikquinn: Looks like this is misaligned

src/assets/scripts/client/InputController.js
```diff
@@ -713,6 +714,16 @@ export default class InputController {
     }
 
     /**
+     * Disable default behavior of right-click menu appearing on right click
+     *
+     * @for InputController
+     * @method _disableRightClickMenu
+     */
+    _disableRightClickMenu() {
```
n8rzz: @erikquinn why do we need this in addition to the `preventDefault` we already have?

his is a clever solution but I'm really not a fan.  We can solve this in a much simpler way.

src/assets/scripts/client/InputController.js
```diff
@@ -290,7 +292,6 @@ export default class InputController {
         if (event.which === MOUSE_EVENT_CODE.MIDDLE_PESS) {
             UiController.ui_zoom_reset();
         } else if (event.which === MOUSE_EVENT_CODE.RIGHT_PRESS) {
-		    event.preventDefault();
```
n8rzz: Why are we _not_ doing this here?

src/assets/scripts/client/InputController.js
erikquinn: Internet seems to indicate this is how the right click menu has to be handled. The standard preventDefault didn't prevent the right click menu from appearing

src/assets/scripts/client/InputController.js
erikquinn: Covered in _disableRightClickMenu()

src/assets/scripts/client/InputController.js
sam-irl: Just an example, this is what happens when you right click currently (screengrab from #576 deployment) 
image

erikquinn: For reference: 
[https://stackoverflow.com/questions/7477184/preventdefault-blocks-right-click-menu-in-firefox-on-mac-but-not-windows?rq=1](https://stackoverflow.com/questions/7477184/preventdefault-blocks-right-click-menu-in-firefox-on-mac-but-not-windows?rq=1)

n8rzz: do we see this behavior with 
```javascript
event.preventDefault();
event.stopPropagation();
``` 
or
```javascript
event.preventDefault();
event.stopImmediatePropagation();
```

It just seems odd that we would have to drill into the document this way.

src/assets/scripts/client/InputController.js
```diff
@@ -711,6 +714,16 @@ export default class InputController {
     }
 
     /**
+     * Disable default behavior of right-click menu appearing on right click
+     *
+     * @for InputController
+     * @method _disableRightClickMenu
+     */
+    _disableRightClickMenu() {
+        document.querySelector('body').addEventListener('contextmenu', (e) => e.preventDefault());
```
n8rzz: If we must do it this way, there are a few things that we should change:
- add a class property for `body` (see current `$window` prop) that gets set in the `_init()` method
- jquery is not necessary, we can keep it like it is here.
- move the attachment part of this to the `setupHandlers()` method
- use `event` instead of `e` it avoid confusion and add additional context to the call
- add a `removeEventListener()` call that removes this listener. this can live in the `disable()` method
- if the disable method does not exist, create one

sam-irl:
This is going to be my main priority, I have been a bit busy recently

erikquinn:
@sam-irl bump on this one? Could really do with getting it merged in so it doesn't conflict too much with #597.

src/assets/scripts/client/InputController.js
erikquinn: Tried the combinations of all options offered up above, both in right-click events' mouse-UP and mouse-DOWN events (since it seems to appear on mouse up anyway), but sadly got no results from them. :disappointed:

erikquinn:
@n8rzz adjusted, and ready for your :eyes:

CHANGELOG.md
```diff
@@ -7,8 +7,9 @@
 - Add new openScope emblem vector graphic [#572](https://github.com/openscope/openscope/issues/572)
 - Adds additional meta tags to index.html head [#484](https://github.com/openscope/openscope/issues/572)
 - Adds a link to the full command reference at the end of the tutorial [#581](https://github.com/openscope/openscope/issues/581)
+- The mouse button to drag the radar screen is now right click [#564](https://github.com/openscope/openscope/issues/564)
 - The distance separator behind aircraft on ILS is now toggleable [#536](https://github.com/openscope/openscope/issues/536)
-- Removes the blue line "departure window" [#411](https://github.com/openscope/openscope/issues/411)
+- Removes the blue line "departure window" [#411](https://github.com/openscope/openscope/issues/411
```
n8rzz: we need that closing `)` here to close the markdown link

src/assets/scripts/client/InputController.js
```diff
@@ -122,6 +122,7 @@ export default class InputController {
         this.input.mouseDown = [0, 0];
         this.input.isMouseDown = false;
 
+        // TODO: Do we want setupHandlers to be `enableHandlers`, and called from `.enable()`?
```
n8rzz: no. the `setupHandlers` method is not needed here.  I added these in error (all over the place 😱).

adding/removing event listeners should happen within `.enable()` and `disable()` and the two handler methods can be removed.

In es5 land, the `.setupHandlers()` method was used to properly bind event handlers: 
```javascript
this._onClickHandler = this.$element.on('click', this._onClick).bind(this);
```

With arrow functions, we don't need to do this anymore so these methods are no longer needed.

src/assets/scripts/client/InputController.js
```diff
@@ -132,6 +133,7 @@ export default class InputController {
      * @method _init
      */
     _init() {
+        this.$body = $('body')[0];
```
n8rzz: Can we use the root `$element` that is captured in `App` instead of looking for it manually here?  I can't see it in the diff, but be sure that `this.$body` is also included in the `constructor`.

src/assets/scripts/client/InputController.js
```diff
@@ -141,10 +143,25 @@ export default class InputController {
...
+    }
+
+    /**
+     * Disable handlers
+     *
+     * @for InputController
+     * @method setupHandlers
+     */
+    disableHandlers() {
+        this.$body.removeEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
```
n8rzz: this method can be removed with the contents moved to `.disable()`

src/assets/scripts/client/InputController.js
```diff
@@ -141,10 +143,25 @@ export default class InputController {
     }
 
     /**
+     * Enable handlers
+     *
      * @for InputController
      * @method setupHandlers
      */
     setupHandlers() {
+        this.$body.addEventListener('contextmenu', (event) => this._disableRightClickMenu(event));
```
n8rzz: this method can be removed with it's contents moved to `.enable()`

src/assets/scripts/client/InputController.js
```diff
@@ -187,6 +204,8 @@ export default class InputController {
 
         this._eventBus.off(EVENT.STRIP_CLICK, this.input_select);
 
+        this.disableHandlers();
```
n8rzz: remove

src/assets/scripts/client/InputController.js
```diff
@@ -289,14 +309,14 @@ export default class InputController {
         // TODO: this should use early returns instead of the else if
```
n8rzz: can you update this `TODO` to say: _This should use a `switch` on `event.which` instead of `if/else if`_

src/assets/scripts/client/InputController.js
```diff
@@ -711,6 +731,17 @@ export default class InputController {
...
 
     /**
+     * Disable default behavior of right-click menu appearing on right click
+     *
+     * @for InputController
+     * @method _disableRightClickMenu
+     * @param event {jQuery event}
+     */
+    _disableRightClickMenu(event) {
+        event.preventDefault();
```
n8rzz: do we also need to prevent event bubbling here with `event.stopPropagation();`?

src/assets/scripts/client/tutorial/TutorialView.js
```diff
@@ -205,6 +205,15 @@ export default class TutorialView {
             position: tutorial_position
         });
 
+	    this.tutorial_step({
```
n8rzz: someday I would love to re-do this entire tutorial.  makes me cringe a little.  but, alas, that is a very low priority.

n8rzz:
@erikquinn ^^

CHANGELOG.md
erikquinn: don't know how that happened!

fixed

src/assets/scripts/client/InputController.js
erikquinn: Ah, it all makes sense now. 👍 

Removed the `setupHandlers` and `disableHandlers` methods and moved the stuff to `.enable()` and `.disable()`.

src/assets/scripts/client/InputController.js
erikquinn: Added a `$body` property (initially null), then set to `this.$element[0]` in `_init()`.

src/assets/scripts/client/InputController.js
erikquinn: done

src/assets/scripts/client/InputController.js
erikquinn: done

src/assets/scripts/client/InputController.js
erikquinn: done

src/assets/scripts/client/InputController.js
erikquinn: done

src/assets/scripts/client/InputController.js
erikquinn: We can if that's a thing we should do? Not my area, so you tell me. Though since it is already at the `$body` level, there is basically nowhere for the event to bubble up to I would think? That's not to say we shouldn't put that in there anyway though.

src/assets/scripts/client/tutorial/TutorialView.js
erikquinn: ^ how I feel about a myriad of things in life...

src/assets/scripts/client/InputController.js
n8rzz: looks like we're good here.  disregard for now

n8rzz:
@erikquinn after you address that conflict (and update with `develop`) this is good to go!

:shipit:

n8rzz:
also this looks _way_ behind `develop`.  We should probably give this a run through one last time before we merge.  just to be safe.

PR #596, opened 2017-07-23
documentation/167

### LINKED ISSUE #167: Add new issue template
This should help us get a minimum set of data to begin investigating bugs.

```
Browser:
Airport:

#### Issue Description


#### Steps to reproduce
- 

```

sam-irl:
Adds templates for issues (resolves #167) and pull requests. This issue was opened some months previously but I think the idea could use discussion. Thoughts appreciated.


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1 @@
+Resolves #
```
n8rzz: We should add a short description line here too

erikquinn: Basically that would be the blank part of `The purpose of this pull request is to ________________________________________.`

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+Browser:
```
erikquinn: Since such an incredibly small number of our issues are caused by browser choice, do we really think it's necessary to ask for it every time?

PULL_REQUEST_TEMPLATE.md
sam-irl: gotcha

ISSUE_TEMPLATE.md
n8rzz: I say yes. If we don't need it fine, but I'd rather set the expectation that we _do_ need it. Then we don't need to ask later. Just makes things faster.

erikquinn: As you wish! :+1:

erikquinn:
¡Estamos de acuerdo!

(Ahem... Yeah, let's do it)

PR #611, opened 2017-07-30
enhancement/326

### LINKED ISSUE #326: Provide read backs for out of range performance commands
Add read backs from aircraft that are asked to exceed their performance limitations 
- climb above ceiling
- slow below model.speed.min
- speed up above model.speed.max

sam-irl:
Resolves openscope/openscope#326 

The purpose of this pull request is to check if an assigned altitude or speed is out of the aircraft's performance range.

If a speed exceeds `AircraftModel.model.speed.max`, or the inverse, or an altitude is above the ceiling, the player is notified.


sam-irl:
Note that every file edited is marked as having the first line changed. This is because the changes were made in Visual Studio; I'm not sure if it's a problem.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -117,6 +126,10 @@ export default class Pilot {
     maintainAltitude(currentAltitude, altitude, expedite, shouldUseSoftCeiling, airportModel) {
         const { minAssignableAltitude, maxAssignableAltitude } = airportModel;
         let clampedAltitude = clamp(minAssignableAltitude, altitude, maxAssignableAltitude);
+        
+        if (altitude > maxAssignableAltitude) {
+            this.assignedUnattainableAltitude = false;
```
n8rzz: two things here:
1) This property seems to be used only within this method, thus, it _should not_ be an instance variable. This should be declared and used only within this method.
1) boolean type properties should be prefixed with the word `is` and worded in a truthy manner. so this could be revised to `isAssignedAltitudeAttainable`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,17 +146,22 @@ export default class Pilot {
         const altitudeInstruction = radio_trend('altitude', currentAltitude, altitude);
         const altitudeVerbal = radio_altitude(readbackAltitude);
         let expediteReadback = '';
+        let altitudeUnattainable = '';
```
n8rzz: this name should also contain the word `readback`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,17 +146,22 @@ export default class Pilot {
...
         }
+        
+        if (this.assignedUnattainableAltitude) {
+            altitudeUnattainable = 'requested altitude unattainable, ';
+        }
 
         const readback = {};
-        readback.log = `${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
-        readback.say = `${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
+        readback.log = `${altitudeUnattainable}${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
```
n8rzz: I know we don't have any tests for the readbacks, in this method.  But, with the addition of another logic block, could you add one for when an altitude is attainable and when one is not?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -222,16 +240,22 @@ export default class Pilot {
      * @param {Number} speed - the speed to maintain, in knots
      * @return {Array} [success of operation, readback]
      */
-    maintainSpeed(currentSpeed, speed) {
+    maintainSpeed(currentSpeed, speed, aircraftModel) {
         const instruction = radio_trend('speed', currentSpeed, speed);
+        let speedUnattainable = '';
```
n8rzz: same as above
>boolean type properties should be prefixed with the word is and worded in a truthy manner

so this could be revised to `isAssignedSpeedUnattainable`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -222,16 +240,22 @@ export default class Pilot {
...
         this._mcp.setSpeedFieldValue(speed);
 
         // Build the readback
         const readback = {};
-        readback.log = `${instruction} ${speed}`;
-        readback.say = `${instruction} ${radio_spellOut(speed)}`;
+        readback.log = `${speedUnattainable}${instruction} ${speed}`;
+        readback.say = `${speedUnattainable}${instruction} ${radio_spellOut(speed)}`;
+        
+        if (speed > aircraftModel.model.speed.max || speed < aircraftModel.model.speed.min) {
```
n8rzz: could we add this logic to the `AircraftTypeModel` class instead of doing this inline?  something like:
```javascript
// in AircraftTypeModel.js
isValidSpeed(nextSpeed) {
  return nextSpeed > this.model.speed.max || 
    nextSpeed < this.model.speed.min;
}
```

so then here we can simply use: 
```javascript
if (this.isValidSpeed(speed)) {
 // ...
}
```

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -222,16 +240,22 @@ export default class Pilot {
...
+            speedUnattainable = 'requested speed unattainable, ';
```
n8rzz: does this assignment actually have an effect?  my guess is no.  could you add at least one test for this case?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -222,16 +240,22 @@ export default class Pilot {
...
+            return [false, readback]
```
n8rzz: there should be an empty line above any `return` statement

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -288,7 +289,7 @@ export default class AircraftCommander {
     runSpeed(aircraft, data) {
         const speed = data[0];
```
n8rzz: I know this was already there, but could you update this var name to `nextSpeed`?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2460,4 +2461,22 @@ export default class AircraftModel {
     removeConflict(conflictingAircraft) {
         delete this.conflicts[conflictingAircraft.callsign];
     }
+
+    /**
+     * @for AircraftModel
+     * @method isValidSpeed
+     * @type {boolean}
+     */
```
n8rzz: These `isValid` methods look good except they are handled n the wrong class. My mistake, I may not have explained that very well before. 

These two methods should be moved to the `AircraftTypeDefnitionModel` and references should be updated.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -114,7 +114,7 @@ export default class Pilot {
      * @param airportModel {AirportModel}
      * @return {array}            [success of operation, readback]
      */
-    maintainAltitude(currentAltitude, altitude, expedite, shouldUseSoftCeiling, airportModel) {
+    maintainAltitude(currentAltitude, altitude, expedite, shouldUseSoftCeiling, airportModel, aircraftModel) {
```
n8rzz: Could you add a `TODO` here to look at this later. 6 parameters is too many, there may be an easier way. This can be handled later though.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,18 +133,28 @@ export default class Pilot {
...
         if (expedite) {
             // including space here so when expedite is false there isnt an extra space after altitude
             expediteReadback = ' and expedite';
 
             this.shouldExpediteAltitudeChange();
         }
+        
+        
 
         const readback = {};
```
n8rzz: There should only be one empty line between statements not 3

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -114,7 +114,7 @@ export default class Pilot {
      * @param airportModel {AirportModel}
```
n8rzz: Be s Re to update the doc blocks if you're adding parameters

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -222,7 +232,7 @@ export default class Pilot {
      * @param {Number} speed - the speed to maintain, in knots
      * @return {Array} [success of operation, readback]
```
n8rzz: Missing doc block

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,25 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
 
         return aircraftIcao.toUpperCase();
     }
+    /**
+     * Checks if the given speed is valid.
+     *
+     * @for AircraftTypeDefinitionModel
+     * @param nextSpeed
```
n8rzz: be sure to include the data type here as well

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,25 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+     */
+     isValidSpeed(nextSpeed) {
+       return nextSpeed > this.speed.min && nextSpeed < this.speed.max;
+     }
+     
+     /**
+      * Checks if a given altitude is valid.
+      *
+      * @for AircraftTypeDefinitionModel
+      * @param nextAltitude
```
n8rzz: be sure to include the data type here as well

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,25 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
         return aircraftIcao.toUpperCase();
     }
+    /**
+     * Checks if the given speed is valid.
+     *
+     * @for AircraftTypeDefinitionModel
+     * @param nextSpeed
+     * @return {Boolean}
+     */
+     isValidSpeed(nextSpeed) {
```
erikquinn: This method name doesn't seem appropriate to me. What really makes a "speed" to be "valid" or not? I would say it's whether or not it is a number. Whether or not `nextSpeed` is _valid_ isn't really the question this method answers... perhaps a better name would be like `isAttainableSpeed()`?

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,25 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+     }
+     
+     /**
+      * Checks if a given altitude is valid.
+      *
+      * @for AircraftTypeDefinitionModel
+      * @param nextAltitude
+      * @return {Boolean}
+      */
+      isValidAltitude(nextAltitude) {
```
erikquinn: Same comment as with speed-- should have a better name, since there is no such thing as an altitude that is _itself_ invalid.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,18 +134,27 @@ export default class Pilot {
         const altitudeInstruction = radio_trend('altitude', currentAltitude, altitude);
         const altitudeVerbal = radio_altitude(readbackAltitude);
         let expediteReadback = '';
-
+
```
erikquinn: We want to remove these extra spaces, no?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,18 +134,27 @@ export default class Pilot {
...
-
+        
         if (expedite) {
             // including space here so when expedite is false there isnt an extra space after altitude
             expediteReadback = ' and expedite';
 
             this.shouldExpediteAltitudeChange();
         }
-
+
```
erikquinn: We want to remove these extra spaces, no?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -133,18 +134,27 @@ export default class Pilot {
...
+        
         const readback = {};
         readback.log = `${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
         readback.say = `${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
 
+        if (!aircraftModel.model.isValidAltitude(altitude)) {
+            let altitudeUnattainableReadback = 'requested altitude unattainable, ';
+
+            readback.log = `${altitudeUnattainableReadback}${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
+            readback.say = `${altitudeUnattainableReadback}${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
```
erikquinn: Let's simplify all this to like `unable to maintain ${readbackAltitude} due to performance`.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -232,6 +243,14 @@ export default class Pilot {
         const readback = {};
         readback.log = `${instruction} ${speed}`;
         readback.say = `${instruction} ${radio_spellOut(speed)}`;
+        
+        if (!aircraftModel.model.isValidSpeed(speed)) {
+            let speedUnattainableReadback = 'requested speed unattainable, ';
+            readback.log = `${speedUnattainableReadback}${instruction} ${speed}`;
+            readback.say = `${speedUnattainableReadback}${instruction} ${radio_spellOut(speed)}`;
```
erikquinn: `unable to maintain ${speed} knots due to performance`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
sam-irl: I started, but... it could have a family. Think of the *children*.

sam-irl:
I finally got a build to pass, so as soon as I can compile this I will. But all seems good.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2405,6 +2405,7 @@ export default class AircraftModel {
         return shouldEnterHold;
     }
 
+    //TODO: can this be removed? It doesn't do anything.
```
n8rzz: it actually does do stuff.  

This method is called via the `EventBus`: https://github.com/openscope/openscope/tree/develop/src/assets/scripts/client/aircraft/AircraftController.js#L172 which gets triggered in another class (likely the `AircraftController`)

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,24 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
         return aircraftIcao.toUpperCase();
     }
+
+    /**
+     * @for AircraftTypeDefinitionModel
+     * @method assignedAttainableAltitude
+     * @param altitude
+     * @return {Boolean}
+     */
+    assignedAttainableAltitude(altitude) {
```
erikquinn: Still not loving the name here... since it's a method of the aircraft definition, we're basically asking:

`Boeing 737 with winglets, are you assigned an attainable altitude?`

This sounds like it wouldn't take an argument. And it needs to take one, because the type definition won't have access to the MCP of a particular instance. Therefore, I think we'd be better with:

`Boeing 737 with winglets, are you able to maintain 330 knots?`

This would result in a name of `AircraftTypeDefinitionModel.isAbleToMaintainAltitude(altitude)`. I think that form for the altitude and speed methods will make much more sense. It's a bit clunkier, but overall more clear I think.

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,24 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+        return altitude < this.ceiling;
+    }
+
+    /**
+     * @for AircraftTypeDefinitionModel
+     * @method assignedAttainableSpeed
+     * @param speed
+     * @return {Boolean}
+     */
+    assignedAttainableSpeed(speed) {
```
erikquinn: `isAbleToMaintainSpeed(speed) {`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -232,6 +244,13 @@ export default class Pilot {
         const readback = {};
         readback.log = `${instruction} ${speed}`;
         readback.say = `${instruction} ${radio_spellOut(speed)}`;
+        
+        if (!aircraftModel.model.assignedAttainableSpeed(speed)) {
+            readback.log = `unable to maintain ${speed} due to performance limits`;
+            readback.say = `unable to maintain ${radio_spellOut(speed)} due to performance limits`;
```
erikquinn: please add `knots` in both of these readbacks.

sam-irl:
@erikquinn So it *works* fine, but the build fails?

sam-irl:
Okay, I believe the issue is that although we create an `AircraftModel` in the test case, `this.model` isn't getting updated to use the def we give it. It's important to note that it work s fine when deployed - the checks and readbacks are working fine. I'll look into this further, unless you decide it's easier to move them to the AircraftModel

n8rzz:
Be sure the `AircraftModel` fixture is actually using an `AircraftTypeDefinitionModel` and not a simple JS object.  If it's just an object, which is my hunch, that would explain why the methods aren't available under test.

test/aircraft/Pilot/maintainAltitude.spec.js
```diff
@@ -43,7 +42,7 @@ ava('.maintainAltitude() should set mcp.altitudeMode to `HOLD` and set mcp.altit
     const shouldExpediteMock = false;
     const shouldUseSoftCeilingMock = false;
     const pilot = new Pilot(modeControllerFixture, fmsArrivalFixture);
-    const model = new AircraftModel(new AircraftTypeDefinitionModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK), navigationLibraryFixture);
```
n8rzz: Whoa how did I miss that in a PR. Yikes!

.gitignore
```diff
@@ -1,8 +1,9 @@
-*~
+﻿*~
 .DS_Store
 _atc-call-stack.js
 _site/
 coverage/
+.vs/
```
n8rzz: Is this for Visual Studio?

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -298,9 +299,9 @@ export default class AircraftCommander {
      * @param data {array}
      */
     runSpeed(aircraft, data) {
-        const speed = data[0];
+        const nextSpeed = data[0];
 
-        return aircraft.pilot.maintainSpeed(aircraft.speed, speed);
+        return aircraft.pilot.maintainSpeed(aircraft.speed, nextSpeed, aircraft);
```
n8rzz: If we're passing the entire `aircraftModel`, there is little point in sending `aircraft.speed` separately. The receivers my method can get that value itself.

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,24 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
 
         return aircraftIcao.toUpperCase();
     }
+
+    /**
+     * @for AircraftTypeDefinitionModel
+     * @method isAbleToMaintainAltitude
+     * @param altitude
```
n8rzz: Missing data type

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,24 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+     * @return {Boolean}
+     */
+    isAbleToMaintainAltitude(altitude) {
+        return altitude < this.ceiling;
+    }
+
+    /**
+     * @for AircraftTypeDefinitionModel
+     * @method isAbleToMaintainSpeed
+     * @param speed
```
n8rzz: Missing data type

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -203,4 +203,24 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+     * @return {Boolean}
+     */
+    isAbleToMaintainSpeed(speed) {
+        return speed > this.speed.min && speed < this.speed.max;
```
n8rzz: Should these be `>=` and `<=`?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -112,9 +112,10 @@ export default class Pilot {
      * @param expedite {boolean}  whether to use maximum possible climb/descent rate
      * @param shouldUseSoftCeiling {boolean}
      * @param airportModel {AirportModel}
+     * @param aircraftModel {AircraftModel}
      * @return {array}            [success of operation, readback]
      */
-    maintainAltitude(currentAltitude, altitude, expedite, shouldUseSoftCeiling, airportModel) {
+    maintainAltitude(currentAltitude, altitude, expedite, shouldUseSoftCeiling, airportModel, aircraftModel) {
```
n8rzz: We don't need the first two params here if we also get the `aircraftModel`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -145,6 +146,16 @@ export default class Pilot {
         readback.log = `${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
         readback.say = `${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
 
+        if (!aircraftModel.model.isAbleToMaintainAltitude(altitude)) {
+            const requestedAltitude = altitude;
```
n8rzz: This reassignment is unnecessary

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -220,9 +231,10 @@ export default class Pilot {
      * @for Pilot
      * @method maintainSpeed
      * @param {Number} speed - the speed to maintain, in knots
+     * @param {AircraftModel} aircraftModel
```
n8rzz: Param name should come before type

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -220,9 +231,10 @@ export default class Pilot {
...
      * @return {Array} [success of operation, readback]
      */
-    maintainSpeed(currentSpeed, speed) {
+    maintainSpeed(currentSpeed, speed, aircraftModel) {
         const instruction = radio_trend('speed', currentSpeed, speed);
```
n8rzz: We can get rid of one of the speed params here since we also send the `aircraftModel`

.gitignore
sam-irl: Yes. I'll remove it if you want, I was just getting *very* annoyed with it. (I use Atom now, pretty happy but the linter isn't wonderful)

n8rzz: you can leave it, I was just curious.

n8rzz: you could try VSCode?  I've moved on from Atom to VSCode recently and am really liking it.  It's also quite a bit faster than Atom.

test/aircraft/Pilot/cancelApproachClearance.spec.js
```diff
@@ -23,26 +26,26 @@ ava('.cancelApproachClearance() returns early if #hasApproachClearance is false'
...
     const headingBeforeLocalizerInterceptionMock = 3.839724354387525; // 220 in degrees
     const headingAfterLocalizerInterceptionMock = 4.36332;  // appx 250 degrees
     const nextHeadingDegreesMock = 250;
-    const currentSpeedMock = 210;
     const nextSpeedMock = 180;
     const shouldExpediteDescentMock = false;
     const shouldUseSoftCeilingMock = false;
     const pilot = new Pilot(modeControllerFixture, fmsArrivalFixture);
+    const model = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK, navigationLibraryFixture);
+    const currentAltitudeMock = model.speed;
```
erikquinn: This is where the test is failing.

erikquinn: putting in an altitude they'd be descending through to get to the assigned altitude would fix the problem. For example, you can set currentAltitudeMock to `5700` and the test will pass.

sam-irl: 👍

sam-irl: i set it to speed

sam-irl: why do i do this

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -145,6 +145,19 @@ export default class Pilot {
         readback.log = `${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
         readback.say = `${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
 
+        if (!aircraftModel.model.isAbleToMaintainAltitude(altitude)) {
+            const verbalRequestedAltitude = radio_altitude(altitude);
+
+            readback.log = `unable to maintain ${altitude} due to performance`;
+            readback.say = `unable to maintain ${verbalRequestedAltitude} due to performance`;
+
+            this._mcp.setAltitudeFieldValue(aircraftModel.altitude);
```
erikquinn: Once we move this block up, it should not be setting anything at all. Only returning early with a response that explains why the pilot decided not to follow the instruction.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -145,6 +145,19 @@ export default class Pilot {
...
+
+            return [false, readback];
+        }
+
+        this._mcp.setAltitudeFieldValue(clampedAltitude);
```
erikquinn: This should be moved up next to the `.setAltitudeHold()`

test/aircraft/Pilot/maintainAltitude.spec.js
```diff
@@ -5,133 +5,161 @@ import _isArray from 'lodash/isArray';
...
     t.true(shouldExpediteAltitudeChangeSpy.calledOnce);
 });
 
 ava('.maintainAltitude() returns the correct response strings when shouldExpedite is false', (t) => {
-    const currentAltitudeMock = 5000;
     const nextAltitudeMock = 13000;
     const shouldExpediteMock = false;
     const shouldUseSoftCeilingMock = false;
     const pilot = new Pilot(modeControllerFixture, fmsArrivalFixture);
+    const model = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK, navigationLibraryFixture);
```
erikquinn: `aircraftModel` contains a Pilot and MCP. Since we are now creating the entire aircraft, we must remove the `pilot` const above, and use `aircraftModel.pilot` in all below. This is the case in multiple tests.

src/assets/scripts/client/aircraft/Pilot/Pilot.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -145,6 +145,19 @@ export default class Pilot {
         readback.log = `${altitudeInstruction} ${readbackAltitude}${expediteReadback}`;
         readback.say = `${altitudeInstruction} ${altitudeVerbal}${expediteReadback}`;
 
+        if (!aircraftModel.model.isAbleToMaintainAltitude(altitude)) {
+            const verbalRequestedAltitude = radio_altitude(altitude);
+
+            readback.log = `unable to maintain ${altitude} due to performance`;
```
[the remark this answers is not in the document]

src/assets/scripts/client/aircraft/Pilot/Pilot.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -219,10 +232,12 @@ export default class Pilot {
      *
      * @for Pilot
      * @method maintainSpeed
-     * @param {Number} speed - the speed to maintain, in knots
+     * @param aircraftModel {AircraftModel}
+     * @param speed {Number} - the speed to maintain, in knots
```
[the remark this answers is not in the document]

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -233,6 +248,13 @@ export default class Pilot {
         readback.log = `${instruction} ${speed}`;
         readback.say = `${instruction} ${radio_spellOut(speed)}`;
 
+        if (!aircraftModel.model.isAbleToMaintainSpeed(speed)) {
```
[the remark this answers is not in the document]

PR #628, opened 2017-08-13
refactor/481

n8rzz:
Resolves #481 

Adds LESS preprocessor and adds CSS folder structure

Issue #481, opened 2017-05-22
Add LESS and reorganize CSS


n8rzz:
- add css preprocessor that gives us variables and sane imports
- organize css in an object-oriented manner (base, helpers, landmark, modules)

PR #631, opened 2017-08-15
refactor/431

### LINKED ISSUE #431: Deactivate unused WIP airports on develop
~~While we get all airports to the improved standards recently laid out, all other airports should be commented out from the load list. We will get as many airports as humanly possible fixed up this sprint in order to keep the list of choices as healthy as possible for the June 1st release.~~

~~Of course (so the game can run), this will only be done once _any_ airport meeting the standards has been merged.~~

By examining the logs of which airports are not being used, remove unneeded WIP airports from the load list.

erikquinn:
Resolves #431.

Remove unused WIP airports from the load list, and add premium flag.

### DIFF
--- a/src/assets/style/airport.css
+++ b/src/assets/style/airport.css
@@ -65,15 +65,22 @@
   width: 5em;
   font-weight: bold;
   font-size: 120%;
-  margin-right: 0.25em;
   text-align: right;
   white-space: nowrap;
 }
 
 #airport-switch li .icao {
   display: inline-block;
-  width: 4em;
-  /*margin-right: */
+  width: 2.5em;
+  margin-left: 1em;
+  text-align: center;
+  font-weight: bold;
+}
+
+#airport-switch li .symbol {
+  display: inline-block;
+  width: 0.7em;
+  margin-left: 1em;
   text-align: center;
   font-weight: bold;
 }
@@ -81,6 +88,7 @@
 #airport-switch li .name {
   display: inline-block;
   font-weight: bold;
+  margin-left: 1em;
 }
 
 #airport-switch li:hover {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Extends departing spawnPatterns outside the airspace at MDSD to prevent point deduction [#615](https://github.com/openscope/openscope/issues/615)
 - Extends departing spawnPatterns outside the airspace at KBOS to prevent point deduction [#635](https://github.com/openscope/openscope/issues/635)
 - Implement scope themes, and changed default theme to blue-based [#15](https://github.com/openscope/openscope/issues/15)
+- Deactivate unused WIP airports from the load list, and add premium flag [#431](https://github.com/openscope/openscope/issues/431)
 
 
 
--- a/assets/airports/airportLoadList.js
+++ b/assets/airports/airportLoadList.js
@@ -45,403 +45,470 @@ window.AIRPORT_LOAD_LIST = (function() { // eslint-disable-line wrap-iife
             icao: 'ebbr',
             level: 'easy',
             name: 'Brussels-National',
+            premium: false,
             wip: true
         },
         {
             icao: 'eddf',
             level: 'medium',
             name: 'Frankfurt Airport',
+            premium: false,
             wip: true
         },
         {
             icao: 'eddh',
             level: 'easy',
             name: 'Hamburg Airport',
+            premium: false,
             wip: true
         },
         {
             icao: 'eddm',
             level: 'hard',
             name: 'Franz Josef Strauß International Airport',
+            premium: false,
             wip: true
         },
         {
             icao: 'eddt',
             level: 'medium',
             name: 'Berlin Tegel Airport',
+            premium: false,
             wip: true
         },
         // {
         //     icao: 'egcc',
         //     level: 'hard',
         //     name: 'Manchester Airport',
+        //     premium: false,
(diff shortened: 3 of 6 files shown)


n8rzz:
If this is what you want to do thats cool.  I just want to make sure you understand what doing this could mean to our users, and that there will likely be some consequences after the fact.

erikquinn:
I'm not opposed to leaving the `show wip airports` option open a bit longer, but we do need to make those airports that are completely broken to be inaccessible on the production site. Since some of them (such as EHAM) are broken in a way that causes a crash with looping errors, _and_ refreshing the page may bring users back to that broken airport (since the most recent airport is in local storage), it could appear to them that the entire site is broken, when its a simple case of an airport file not conforming to the latest designs needed to run properly. Most users won't know how to flush the cache, and each time they come back they may be continually unable to do anything.

As an alternative option, and only if there is significant support for it, we could kind of create a third level of airport status, akin to `reliable`, `unreliable`, and `broken`, and then deactivate only the broken ones to fix the above core concern. I know many people may not use the site like I do, but for me personally, the fact that my home airport (KIAD) is not available for me to use is only a slight bummer. Since there is enough variety in the other available airports, it doesn't bother me much. That said, perhaps for some it is a game changer to not have access to their home airport file (which may even be mostly functional)-- so I don't seek to alienate anybody, and if many would be upset by this, then we can consider keeping it around. But in the end, the goal of the overall project is to be the highest _quality_, and it would be counter to that goal for us to knowingly use airports that do not work properly.

The real solution I think is to look at the analytics and make a list of critical airports to repair, and work with our airport guys to get them fixed up. And if we're fairly quick about that, someone's home airport being out of commission for 30 days for critical repairs sounds reasonable to me.

Perhaps it would be advisable to get some form of vote on whether or not this is a good idea? There are definitely downsides to either course of action, and I'm happy to hear from anybody who has thoughts on either side.

n8rzz:
@openscope/simulator-collaborators looking for opinions from the team on the conversation @erikquinn and I are having about WIP airports.  

I'll add my opinions later tonight.

Fechulo:
I'm leaning towards leaving WIP airports as they are for now. Slowly, more and more airports are being added and at some point I think we'll have to get rid of the old ones. Despite this, I can understand why it would also be beneficial to deactivate them now.

A big problem seems to be that users are unaware of the _risks_ of using WIP airports. This could be solved by adding a warning when a user chooses to use a WIP airport.

Additionally, I'd be willing to make a short video on my channel explaining the issue. perhaps people will give their opinions on the topic and help us reach a decision. I know that only a very small portion of our user base actually watches my videos, but it might help raise some awareness.

indianbhaji:
I'd keep that WIP option there... personally I've stopped any progress on my pull requests due to the wip/premium segregation

georgeenciu:
i'd go with the wip remaining in the game and a message box or notification telling the player that not all features might be available or the gameplay might be buggy

panther2:
Though still not feeling qualified for giving a substantial answer I would keep the WIP airports in the game.
It might be good to give an impression about the constant improvement and work to users not familiar with the development process here on Github.

n8rzz:
@erikquinn Mmmm, broken airports.  I forgot that is a thing.  Those I feel we should definitely remove.  Looking at the list from the last 30 days, I would guess that _those_ are working for the most part.  I'd be ok with removing everything but the ones in the list above.  

That way we can cut out a lot of the sub-par airports while also keeping the ones that get used regularly.

erikquinn:
So I've added back the WIP option here, and will keep available all of the WIP airports that have been used (even if only once) in the past two weeks.

This includes EBBR, EDDF, EDDH, EDDM, EDDT, EGKK, EGLC, EGLL, EGNM, EHAM, EIDW, EKCH, ENGM, GCRR, KBOS, KLAS, KLAX, KPDX, KPHX, KSEA, KSFO, LOWW, MDSD, OMDB, SAWH, SBGR, TNCM, and VHHH.

All of these airports I have personally tested, and as luck would have it, only _one_ has issues with freezing up (EHAM). All others may have some minor flaws, but at least run properly. So they will remain available by enabling the `Include WIP Airports` option.

Eventually, we will remove the option altogether, but since none of these are _totally broken_, we can keep them on as popular demand requests. Once we get a bunch of these repaired, and expand the choices in the load list more, we can eventually come back to the topic-- maybe around February or so. But this will be good for now. Appreciate all the input :smiley:

erikquinn:
And if there are any WIP airports other than the ones that were recently used that somebody wants included, just advise, and as long as it "works" it can be made available as well.

src/assets/style/airport.css
```diff
@@ -65,22 +65,30 @@
...
-  width: 4em;
-  /*margin-right: */
+  width: 2.5em;
+  margin-left: 1em;
+  text-align: center;
+  font-weight: bold;
+}
+
+#airport-switch li .symbol {
+  display: inline-block;
```
n8rzz: Property order is a thing with CSS rules, but these are all going to be refactored with #481, so you can leve this as-is for now.

https://github.com/thenerdery/html-css-standards/blob/master/standards/css.md#property-order

erikquinn: ah! perfect. I tried to conform with what I was seeing elsewhere, but there were no exact examples so I figured I had it wrong. Will read the linked document when I get bored :smiley:

n8rzz: you'll see better examples once #628 is in 😉

PR #634, opened 2017-08-20
refactor/354

### LINKED ISSUE #354: Tutorial lists an arriving aircraft callsign on the first page
should be the first departing aircraft that is listed

Dabea:
Resolves #354 .

~~This reverse the airport list so the newest spawned airport will be chosen for the tutorial to display~~

Ensures tutorial selects first departure aircraft

### DIFF
--- a/src/assets/scripts/client/tutorial/TutorialView.js
+++ b/src/assets/scripts/client/tutorial/TutorialView.js
@@ -193,6 +193,7 @@ export default class TutorialView {
         prop.tutorial.open = false;
 
         const tutorial_position = [0.1, 0.85];
+        const departureAircraft = prop.aircraft.list.filter((aircraftModel) => aircraftModel.isDeparture())[0];
 
         this.tutorial_step({
             title: 'Welcome!',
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/tutorial/TutorialView.js
```diff
@@ -201,8 +201,8 @@ export default class TutorialView {
                 if (prop.aircraft.list.length <= 0) {
                     return t;
                 }
-
-                return t.replace('{CALLSIGN}', prop.aircraft.list[0].callsign);
+                const reverseList = [...prop.aircraft.list].reverse();
```
erikquinn: I believe this does not need the spread into an array-- this is the same as `prop.aircraft.list.reverse();`. And we are trying to eliminate use of the `prop` as well.

Also, we actually would want the closest controllable arrival and first departure, while this would give us the furthest. Perhaps something like `_filter(aircraftList, (aircraft) => aircraft.flightPhase === FLIGHT_PHASE.APRON)[0]` or similar would be what we need here.

n8rzz: Using the spread here actually gives us a new array, leaving the original unaffected by the following .reverse()

n8rzz: @erikquinn your solution, though, is cleaner. 👍

Dabea: I belive we would still need to reverse it though with ```_filter(aircraftList, (aircraft) => aircraft.flightPhase === FLIGHT_PHASE.APRON)[0]```since as they are loaded in they are added to the end if the array

erikquinn: @n8rzz oh, I guess you're saying `.reverse()` mutates. I see what's going on. That's a clever solution then!

@Dabea The list of aircraft in the AircraftController still has new aircraft added to the end, so the `[0]` index of the filter restults will be the "first" arrival or departure, like we want. The aircraft we want _is_ the one whose strip is at the bottom of all other aircraft of their category, so it should get the right one without any reversing.

PR #601, opened 2017-07-26
refactor/365

sam-irl:
Resolves #365 

The purpose of this pull request is to rename documentation/index.md to commands.md, and add system commands.

I added system commands, as well as `squawk`, but I'm still unsure what some commands _do_, specifically `auto` and `transmit`. If someone could enlighten me that would be appreciated.

PR #643, opened 2017-08-22
refactor/639

### LINKED ISSUE #639: Deprecate `abort` command
The `abort` command is not specific enough to be realistic, nor is it really comprehensive, since only very few actions can actually be "aborted". Similar commands could be created such as `go around / hold position` etc later, though they're hardly needed.

sam-irl:
Resolves #639 

The purpose of this pull request is to deprecate `abort`

Depends on #601 for the docs, I know I updated them here but that's more because I wasn't thinking


sam-irl:
@n8rzz That really isn't any better, what's the point of just displaying that it *will* be removed, and nothing else? The amount of useful information displayed is the same.

n8rzz:
perhaps.  but at least there is some feedback for the users that are expecting the command to work (and haven't read the documentation)

erikquinn:
I agree with both... Leaving it partially in place only makes the removal process that much more painful. _BUT_ with a command like `abort`, people probably do actually use it (unlike `sr`). So pulling it out from under them may be less tasteful than turning it off and having the game explain to them what happened.

I suppose @n8rzz is right though-- it's better to err on the side of keeping the users informed. We need to save our `we're doing this because it's what's best for the project even if you don't like it` points for later rather than spending them all early and losing the loyalty of some users. :wink:

src/assets/scripts/client/commandParser/commandDefinitions.js
```diff
@@ -71,10 +71,7 @@ const ZERO_ARG_COMMANDS = {
     },
 
     // Aircraft commands
-    abort: {
-        validate: zeroArgumentsValidator,
-        parse: noop
```
erikquinn: This needs to be present if the command is to register upon a user entering it

src/assets/scripts/client/commandParser/commandMap.js
```diff
@@ -108,7 +108,6 @@ export const COMMAND_MAP = {
     direct: 'direct',
     pd: 'direct',
     dct: 'direct',
-    abort: 'abort',
```
erikquinn: This needs to be present if the command is to register upon a user entering it

sam-irl: Hurray, I'm an idiot

n8rzz: @sam-irl not an idiot at all.  This is why we do pull requests.  Another set (or sets) of eyes is always a good idea!

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -216,6 +216,18 @@ export default class AircraftCommander {
...
     /**
+     * Aborts an action. Deprecated.
+     *
+     * @for AircraftCommander
+     * @method runAbort
+     * @param {AircraftModel} aircraft
+     * @return {array} [success of operation, readback]
+     */
+    runAbort(aircraft) {
+        return [false, 'this command has been deprecated, clear an aircraft to another fix or heading instead'];
```
erikquinn: let's change to `the 'abort' command has been deprecated, please see documentation for help`

PR #661, opened 2017-08-29
bugfix/660

### LINKED ISSUE #660: MDSD: Arrivals spawn too high to descend in time
Browser: Chrome
Airport: MDSD

#### Issue Description
Numerous aircraft (particularly from the NW) struggle to descend from their spawn altitude quickly enough to be controllable. This can be resolved by either:
1. lowering the spawn altitude
1. extending the arrrival routes further out so the aircraft have additional time to descend

Fechulo:
Resolves #660 

Lowers the spawn altitude of aircraft inbound on the RIBES arrival.

### DIFF
--- a/assets/airports/mdsd.json
+++ b/assets/airports/mdsd.json
@@ -1152,7 +1152,7 @@
             "exitPoints": {
                 "AVRES": ["AVRES"]
             },
-            "draw": []
+            "draw": [[]]
         },
         "BIBOK1": {
             "icao": "BIBOK1",
@@ -1178,7 +1178,7 @@
             "exitPoints": {
                 "BIBOK": ["BIBOK"]
             },
-            "draw": []
+            "draw": [[]]
         },
         "DELPO1": {
             "icao": "DELPO1",
@@ -1204,7 +1204,7 @@
             "exitPoints": {
                 "DELPO": ["DELPO"]
             },
-            "draw": []
+            "draw": [[]]
         },
         "KANAM": {
             "icao": "KANAM",
@@ -1247,7 +1247,7 @@
             "exitPoints": {
                 "KARIB": ["KARIB"]
             },
-            "draw": []
+            "draw": [[]]
         },
         "KERSO": {
             "icao": "KERSO",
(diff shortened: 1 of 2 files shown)


assets/airports/mdsd.json
```diff
@@ -1152,7 +1152,7 @@
             "exitPoints": {
                 "AVRES": ["AVRES"]
             },
-            "draw": [[]]
+            "draw": []
```
erikquinn: Sorry, I didn't say anything yet because it literally just went in (#480), but any empty draw arrays will have to be a double empty array (technical reasons)

Fechulo: Oh ok! May I ask what those technical reasons are? (ELI5 please)

erikquinn: lol sure! here's the relevant section:
```
/**
 * Returns an array of unique fix names used by any portion of the procedure
 *
 * @for StandardRouteModel
 * @method getAllFixNamesInUse
 * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
 */
getAllFixNamesInUse() {
    const drawFixNames = this.draw.reduce((list, fix) => list.concat(fix))
        .map((fixName) => fixName.replace('*', ''));

    const allFixNames = [
        ...this._entryCollection.getAllFixNamesInUse(),
        ...this._bodySegmentModel.getAllFixNamesInUse(),
        ...this._exitCollection.getAllFixNamesInUse(),
        ...drawFixNames
    ];
    const uniqueFixNames = _uniq(allFixNames);

    return uniqueFixNames;
}
```

You see we access `this.draw`, which is the value you're assigning here to `"draw"`, and then we step through each fix in each array included in the draw section. We do this with two functions that are native to an `Array`, being `Array.reduce()` and `Array.map()`.

Once you step into the `.reduce()` part, we ask `.map()` to look through the fix names. But if the array is only one-dimensional, like `[]`, we end up being able to do `.reduce()` into the first array, and then _there is no second array to use .map() on_. Sooooo it throws an error because I can only use map on an array (even if empty), I can't use .map() on... nothing.

PR #665, opened 2017-08-30
bugfix/664

### LINKED ISSUE #664: Flight strips out of proportion
Browser: Chrome
Airport: All

#### Issue Description
Flight strips have become out of proportion, due to their being too tall. Regardless of the strips contents, it must maintain its dimensions. When the route overflows, it will need to be truncated; this is a separate issue (#633). In the meantime, overflow beyond the dimensions  of any field may simply be hidden.
image

---

Related: #633.

n8rzz:
Resolves #664 

Updates order of elements in StripViewTemplate to properly float flight rules to the right

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Bugfixes
 - Fix VNAV descents on STARs with only "at/above" and "at/below" restrictions [#618](https://github.com/openscope/openscope/issues/618)
+- Updates order of elements in the StripViewTemplate so the FlightRules element is properly floated to the right [#664](https://github.com/openscope/openscope/issues/664)
+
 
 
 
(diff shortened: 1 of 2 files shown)


n8rzz:
Isn't there an existing issue for that already?  The change you're requesting here has nothing to do with this bug so I'd prefer to do it in another PR.

erikquinn:
There's a second issue for getting the strip to hide the _middle part_ of a route when it's too long for the strip. But this issue #664 was about clamping the height of the strip so it could never be a different size regardless of the content.

n8rzz:
Ahh, you're right. Once I found the float thing I had it in my head that that was the issue.

I'll see what I can do.

erikquinn:
Added #672 for the remainder, but this does enough for now. :shipit:

PR #680, opened 2017-09-04
enhancement/679

### LINKED ISSUE #679: Add transparency to StripView

n8rzz:
Resolves #679 

modifies StripView to use `rgba` instead of `hex` for background color allowing for transparent background


n8rzz:
Not high maintenance at all, those are good points. I'll make some adjustments.

n8rzz:
@erikquinn well, the transparency of the stripList works well along with the hover effects.  Its the transparency of the stipItems themselves that looked hideous.  Give it a try and see what you think.

I'm ok either way.

erikquinn:
Disregard, I built the wrong branch and was like "soooooo nothing is different?" 🤦‍♂️

Issue #46, opened 2017-01-16
Adjustable Radar Update Rate


erikquinn:
The rate at which the positions of the aircraft depicted on the scope are updated should be dependent on a user-defined value within the airport file (or eventually, the *facility* file, following further refactors).

Originally reported in zlsa/atc#741.

PR #682, opened 2017-09-05
feature/125

### LINKED ISSUE #125: Genuine Radar Emulation
_From @erikquinn on December 15, 2016 21:46_

Depends on n8rzz/atc#212.

Since we will already have to be dealing with sending updated aircraft position information, some small measures can be taken to more properly emulate the nuances of radar.

### Delayed Drawing
Updated positions of aircraft received from the server would not simply be displayed immediately. A specifiable "update rate" consistent with the 360 degree sweep period of the radar site would be implemented, and the displayed positions updated at the same frequency. For example, ASR9 radar antennas make a full sweep appx. every 4.5 seconds. So despite receiving maybe 50 updates from the server on where the aircraft are, we would only redraw the aircraft every 4.5 seconds, at their latest reported position.

---

### Artificial Sweep
Maybe at some point, an artificial sweep can be enabled. As with the above example, the aircraft's updates are delayed. However, with an artificial sweep, the positions would be updated as the radar antenna rotates ([briefly viewable here](https://youtu.be/-FTYnWID3DI?t=329)). This would be done independently of the sending of updated positions from the server. The server just sends the updated info when it's available, and all the rest is done client-side.

### Transponder-Based Identification
Despite giving lectures on primary and secondary radar a few hundred times, I'll skip that here, because there are plenty of available resources for that. The basic idea can be deduced from the following statement I've always liked to use: "A meteorite falling from the sky will be shown on primary radar, but not secondary." Primary will show *anything* that has a return echo. Secondary is quite different: the transponders within aircraft (with mode c *on*) are broadcasting a 4-digit numerical (octal) "squawk code" in every direction, as well as their current altitude. Any radar site that receives that information is able to display it.

The squawk code is one of the following (more or less):
 - `1200` (VFR)
 - `7500` (hijack)
 - `7600` (radio failure)
 - `7700` (emergency)
 - `("discrete")` - this would be any non-reserved squawk code, they are always unique for each aircraft.

For our purposes, since we only have normal IFR traffic, they would all be assigned individual discrete codes. The way it *really* works is that if the radar system's computers detect a discrete squawk code that matches one listed in the flightplan repository, it associates the thing in the sky with the flightplan, from which it can pull the callsign, flightplan, filed altitude, aircraft type, etc.

_Copied from original issue: n8rzz/atc#214_

sam-irl:
Resolves #125 

The purpose of this pull request is to add a realistic delay between radar updates

### DIFF
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -237,7 +237,7 @@ export default class CanvasController {
         const alpha = extrapolate_range_clamp(0.1, elapsed, 0.4, 0, 1);
         const framestep = Math.round(extrapolate_range_clamp(1, GameController.game.speedup, 10, 30, 1));
 
-        if (this.canvas.dirty || (!GameController.game_paused() && prop.time.frames % framestep === 0) || elapsed < GameController.getGameOption(GAME_OPTION_NAMES.RADAR_UPDATE_DELAY)) {
+        if (this.canvas.dirty || (!GameController.game_paused() && prop.time.frames % framestep === 0) || elapsed > GameController.getGameOption(GAME_OPTION_NAMES.RADAR_UPDATE_DELAY)) {
             const cc = this.canvas_get('navaids');
             const fading = elapsed < 1;
 
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -236,7 +236,7 @@ export default class CanvasController {
         const elapsed = GameController.game_time() - AirportController.airport_get().start;
         const alpha = extrapolate_range_clamp(0.1, elapsed, 0.4, 0, 1);
         const framestep = Math.round(extrapolate_range_clamp(1, GameController.game.speedup, 10, 30, 1));
-        const shouldUpdate = this.canvas.dirty || (!GameController.game_paused() && prop.time.frames % framestep === 0);
+        const shouldUpdate = this.canvas.dirty || !GameController.game_paused() && prop.time.frames % framestep === 0;
         const updateDelay = GameController.getGameOption(GAME_OPTION_NAMES.RADAR_UPDATE_DELAY);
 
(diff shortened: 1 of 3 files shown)


sam-irl:
Jesus Christ, that isn't seconds in the CanvasController.

n8rzz:
@sam-irl there is some translation needed between game option labels and game option values:

screen shot 2017-09-07 at 9 56 29 am

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -237,7 +237,7 @@ export default class CanvasController {
         const alpha = extrapolate_range_clamp(0.1, elapsed, 0.4, 0, 1);
         const framestep = Math.round(extrapolate_range_clamp(1, GameController.game.speedup, 10, 30, 1));
 
-        if (this.canvas.dirty || (!GameController.game_paused() && prop.time.frames % framestep === 0) || elapsed < 1) {
+        if (this.canvas.dirty || (!GameController.game_paused() && prop.time.frames % framestep === 0) || elapsed > GameController.getGameOption(GAME_OPTION_NAMES.RADAR_UPDATE_DELAY)) {
```
n8rzz: oh my this is some tricky logic.  This was here before, so I know it's not your fault.  

at the very least, could you pull some of this out to `const`s before the if block?
```javascript
const updateDelay = GameController.getGameOption(GAME_OPTION_NAMES.RADAR_UPDATE_DELAY);
const shouldUpdate = this.canvas.dirty || !GameController.game_paused() && prop.time.frames % framestep === 0;

if (shouldUpdate || elapsed > updateDelay) {
  // ...
}
```

I still don't like it, but at least its easier to read.  The `CanvasController` will get a good fixing some day in the future.

erikquinn:
Also, this is only a tiny portion of #125. Really what this PR aims to accomplish is #46. When you get a chance, please replace this PR with a new one for a new `enhancement/46` branch.

sam-irl:
Replaced with #691

sam-irl:
Replaced with #694

PR #686, opened 2017-09-06
bugfix/685

### LINKED ISSUE #685: Up/down arrows should insert command instead of non-existent character
Browser: All

Airport: All

#### Issue Description
When using the altitude change hotkeys (up/down arrow), a "not found" character is displayed. This should be changed to insert the command.

#### Steps to reproduce
- Select aircraft
- Use arrow keys to change altitude

sam-irl:
Resolves #685 

The purpose of this pull request is to insert `altitude` into the command bar instead of the unicode arrow char which does not exist in the default font

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -484,7 +484,7 @@ export default class InputController {
                 break;
             case KEY_CODES.UP_ARROW:
                 if (this._isArrowControlMethod()) {
-                    this.$commandInput.val(`${currentCommandInputValue} \u2B61 `);
+                    this.$commandInput.val(`${currentCommandInputValue} altitude `);
                     e.preventDefault();
                     this.onCommandInputChangeHandler();
                 } else {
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/InputController.js
```diff
@@ -484,7 +484,7 @@ export default class InputController {
                 break;
             case KEY_CODES.UP_ARROW:
                 if (this._isArrowControlMethod()) {
-                    this.$commandInput.val(`${currentCommandInputValue} \u2B61 `);
+                    this.$commandInput.val(`${currentCommandInputValue} altitude `);
```
erikquinn: Let's keep the directional context here by using `c` for climb.

src/assets/scripts/client/InputController.js
```diff
@@ -505,7 +505,7 @@ export default class InputController {
                 break;
             case KEY_CODES.DOWN_ARROW:
                 if (this._isArrowControlMethod()) {
-                    this.$commandInput.val(`${currentCommandInputValue} \u2B63 `);
+                    this.$commandInput.val(`${currentCommandInputValue} altitude `);
```
erikquinn: Let's keep the directional context here by using `d` for descend.

src/assets/scripts/client/commandParser/commandMap.js
```diff
@@ -33,8 +33,6 @@ export const SYSTEM_COMMANDS = {
  * @final
  */
 const UNICODE_COMMANDS = {
-    '\\u2B61': 'altitude',
-    '\\u2B63': 'altitude',
     '\\u2BA2': 'heading',
     '\\u2BA3': 'heading',
```
erikquinn: These two are also no longer used. Since we're already here, might as well remove them also.

erikquinn: done

PR #687, opened 2017-09-06
bugfix/623

### LINKED ISSUE #623: Altitudes on aircrafts strips are too precise
Browser: Chrome
Airport: All

#### Issue Description
If "Allow departures via climb" is set to yes and an aircraft is given an altitude higher than the airspace ceiling the target altitude is set to airspace ceiling + 1 foot (e.g. FL150.01) which is fine. The aircraft strip however should only show FL150
altitude

#### Steps to reproduce
- Set "Allow departures via climb" to yes
- Give an aircraft an altitude higher than the airspace ceiling
- Observe the aircraft's altitude on the strip

sam-irl:
Resolves #623 

The purpose of this pull request is to resolve FL in the StripView being too specific

### DIFF
--- a/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
+++ b/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
@@ -701,11 +701,11 @@ export default class StripViewModel extends BaseModel {
      * @return roundedAltitude
      */
     _roundAltitude(altitude) {
-        const roundedAltitude = _round(altitude);
-
-        if (_isNaN(roundedAltitude)) {
+        if (_isNaN(altitude)) {
             return '-';
         }
+
+        const roundedAltitude = _round(altitude);
 
         return roundedAltitude;
     }
--- a/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
+++ b/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import _round from 'lodash/round'
 import BaseModel from '../../base/BaseModel';
 import EventBus from '../../lib/EventBus';
 import { STRIP_VIEW_TEMPLATE } from './stripViewTemplate';
@@ -448,7 +449,7 @@ export default class StripViewModel extends BaseModel {
         this.$aircraftTypeView.text(this._aircraftType);
         this.$cidView.text(this.cid);
         this.$transponderView.text(this._transponder);
-        this.$assignedAltitudeView.text(this._assignedAltitude);
+        this.$assignedAltitudeView.text(_round(this._assignedAltitude));
         this.$flightPlanAltitudeView.text(this._flightPlanAltitude);
         this.$departureAirportView.text(this._departureAirport);
         this.$arrivalAirportView.text(this._arrivalAirport);
--- a/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
+++ b/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
@@ -1,5 +1,6 @@
 import $ from 'jquery';
-import _round from 'lodash/round'
+import _round from 'lodash/round';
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -448,7 +449,7 @@ export default class StripViewModel extends BaseModel {
         this.$aircraftTypeView.text(this._aircraftType);
         this.$cidView.text(this.cid);
         this.$transponderView.text(this._transponder);
-        this.$assignedAltitudeView.text(this._assignedAltitude);
+        this.$assignedAltitudeView.text(_round(this._assignedAltitude));
```
n8rzz: this is a good attempt, however, the `_redraw()` method should not contain any logic or manipulations.  The use of `_round()` should instead occur within both the `_init()` and `_updateStripView()` methods.

n8rzz: and, since you're in this method, could you also add that note to the docblock and update the `method` name?

sam-irl:
@n8rzz 56b8e81

n8rzz:
Ha, I must have just missed that diff. Sorry about that.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -684,12 +684,28 @@ export default class StripViewModel extends BaseModel {
...
+    /**
+     * @for StripViewModel
+     * @method roundAltitude
+     * @param altitude
+     * @return roundedAltitude
+     */
+    _roundAltitude(altitude) {
+        const roundedAltitude = _round(altitude);
+
+        if (roundedAltitude === 'NaN') {
```
n8rzz: `NaN` is a specific javascript object.  here, you are actually looking for the _string_ `NaN`. 
```javascript
'NaN' === NaN; // false
```

instead you should use lodash for this check (just because thats what we're using elsewhere):

```javascript
_isNaN(roundedAltitude)
```

question: what causes `_round(altitude)` to be `NaN`?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -684,12 +684,28 @@ export default class StripViewModel extends BaseModel {
...
         this._departureAirport = departureAirportId;
         this._flightPlan = flightPlan;
 
         return this._redraw();
     }
+
+    /**
+     * @for StripViewModel
+     * @method roundAltitude
+     * @param altitude
```
n8rzz: be sure to add the data type here:
```
@param altitude {number|string}
```

_(assuming there is a string involved due to the `NaN`)_

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -684,12 +684,28 @@ export default class StripViewModel extends BaseModel {
...
+     * @return roundedAltitude
```
n8rzz: be sure to add the data type here as well

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
sam-irl: I believe that, by default, the string '-' is set as altitude assigned, in the AircraftModel.

n8rzz: thats what I thought 👍 

https://lodash.com/docs/4.17.4#isNaN

erikquinn:
Thanks for the work on this one, but after some thought, we should be thinking about the problem differently. Upon further investigation of the corresponding issue, I think we should be forcing the climb in a different way, which would prevent the need for us to use the `+ 0.01` hacky solution. And honestly, this is one place where I really hesitate to conduct any rounding or truncating.

Closing PR and corresponding issue. Further work will be on #692.

PR #690, opened 2017-09-07
refactor/670

### LINKED ISSUE #670: GameOptions not getting stored in localStorage
Browser:  Google Chrome Version 60.0.3112.113 (Official Build) (64-bit)

Airport:  Any

Approximate play time before issue:  Immediate on Launch

#### Issue Description
Launching the 5.4.0 Beta in a new Chrome window (new tab) everything is disabled by default, no matter which airport I start with (cleared the cache to make sure that wasn't part of it).  See screen shot.  Runways & data labels, terrain, restricted areas, routes, etc. all off.   

image

#### Steps to reproduce
- Open browser, launch game in a new tab
- Turn on any options for any airport (e.g. fix and runway labels)
- Close tab, then launch again in a new tab
- Options will be off again

n8rzz:
Resolves #670 

The purpose of this pull request is to:
Adds `localStorage.setItem()` to `GameOptions.setOptionByName()` and adds test file for `GameOptions`

### DIFF
--- a/src/assets/scripts/client/game/GameOptions.js
+++ b/src/assets/scripts/client/game/GameOptions.js
@@ -13,13 +13,37 @@ export default class GameOptions {
      * @constructor
      */
     constructor() {
+        /**
+         * @property _eventBus
+         * @type EventBus
+         * @private
+         */
         this._eventBus = EventBus;
+
+        /**
+         * @property _options
+         * @type {Object}
+         * @default {}
+         * @private
+         */
         this._options = {};
+
+        /**
+         * Model properties will be added for each game option
+         * dynamically via `.addGameOptions()`
+         *
+         * @property {*}
+         * @type {string}
+         *
+         * this[OPTION_NAME] = OPTION_VALUE;
+         */
 
         this.addGameOptions();
     }
 
     /**
+     * Add available game options to `_options` dictionary
+     *
      * @for GameOptions
      * @method addGameOptions
      */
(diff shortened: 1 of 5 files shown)


src/assets/scripts/client/game/GameOptions.js
```diff
@@ -34,22 +58,25 @@ export default class GameOptions {
...
-        let dataName = data.defaultValue;
+    addOption(optionProps) {
+        const optionStorageKey = this.buildStorageName(optionProps.name);
+        const storedOptionValue = global.localStorage.getItem(optionStorageKey);
+        this._options[optionProps.name] = optionProps;
+        let optionValue = optionProps.defaultValue;
 
-        if (_has(localStorage, optionStorageName)) {
-            dataName = localStorage[optionStorageName];
+        if (typeof storedOptionValue !== 'undefined') {
```
erikquinn: just curious, why aren't we using `!_isNil()` here?

n8rzz: It's more explicit. LocalStorage will return `undefined` when a key does not exist.   I suppose I could switch it, probably should to stay consistent.

n8rzz:
@erikquinn that is possible. I whipped this up quick using only the tests. Note that options won't get saved until they are changed.

n8rzz:
@erikquinn the one that doesn't appear to work at the moment is the `theme` option.  it is getting stored correctly, but it seems that option isn't getting reused correctly.  I'll look into that.

n8rzz:
@erikquinn added #696 to address the above, otherwise all options appear to be stored and loading correctly.

n8rzz:
@erikquinn 🤔, well now I feel silly.

The `GameOptions` were not getting stored in `localStorage`, so that is fixed with this PR.  As far as the control buttons (terrain, SID lines, fix names, etc), we haven't ever stored those option values before.  Though I think that would be a good idea, I'll create a new issue for that one (#698).

Lets call this one good with what is here because it does fix broken code and then use #698 for that new work.

PR #691, opened 2017-09-08
enhancement/46

### LINKED ISSUE #46: Adjustable Radar Update Rate

sam-irl:
Resolves #46 

The purpose of this pull request is to add radar delays.


sam-irl:
Replaced by #694 for the correct branch names.

PR #706, opened 2017-09-18
enhancement/296

### LINKED ISSUE #296: Pause simulation when browser tab not in focus
Currently, if the browser window with the simulator in it does not have focus, the game _kind of_ pauses (the aircraft do not move), but the screen continues to be redrawn, the positions recalculated, etc. If the game does not have focus, it should pause itself, and resume when the user presses the play button.

Additionally, an option could be added to the settings menu to disable this behavior, so the simulation will continue to run, even if the browser tab does not have the focus.

n8rzz:
Resolves #296 

Adds `TimeKeeper` singleton and moves `App.incrementFrame()` logic to this new class


src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -252,10 +255,8 @@ export default class CanvasController {
 
                 this.canvas_clear(cc);
                 this.canvas_fill_background(cc);
-                cc.translate(
```
n8rzz: completely unrelated to this PR but I couldn't resist.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -283,22 +285,20 @@ export default class CanvasController {
                 round(this.canvas.size.width / 2 + this.canvas.panX),
                 round(this.canvas.size.height / 2 + this.canvas.panY)
             );
-            // TODO: this is incorrect usage of a ternary. ternaries should be used for a ssignment not function calls.
-            // draw airspace border
-            AirportController.airport_get().airspace
-                ? this.canvas_draw_airspace_border(cc)
-                : this.canvas_draw_ctr(cc);
+
+            if (AirportController.airport_get().airspace) {
```
n8rzz: also, completely unrelated to this PR

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -396,10 +384,11 @@ export default class CanvasController {
      * @method canvas_should_draw
      */
     canvas_should_draw() {
-        const elapsed = time() - this.canvas.last;
+        const currentTime = TimeKeeper.gameTime;
```
n8rzz: need to cache this otherwise we get different values each time

src/assets/scripts/client/game/GameController.js
```diff
@@ -362,12 +361,12 @@ class GameController {
      * @method update_pre
      */
     update_pre() {
+        this.game.delta = Math.min(TimeKeeper.deltaTime * this.game.speedup, 100);
```
n8rzz: this line concerns me.  it appears as though we're keeping track of `deltaTime` in a few places.  This will need to be looked at in the near future

erikquinn:
I have a theory. I remember seeing `$.blur(method);`, and that would not be what we want. We would want to do `$window.addEventListener('blur', method);`. Give that a try with some console logs and see if it's firing properly then?

n8rzz:
yep, we're on the same track there.  

we were using jquery with `$(window).blur()` but there is litte sense in that since it's just a wrapper for `window.addEventListener('blur')`.  The problem is _how_ the blur is handled.  we seem to be detecting it correctly (using either method above) but for some reason there is still a jump on window focus.  I haven't quite narrowed it down the why.

Even stranger is that blur/focus to dev tools _does_ work correctly and it's using the same block as the window blur/focus.  (like, literally, *_the same handler_*).  that seems strange to me why one would work and one doesn't.  I'm sure I'm missing something somewhere.

n8rzz:
and _thats_ why we still see the jump:

screen shot 2017-09-24 at 2 02 25 pm

`deltaTime` should be `0` and not 7

n8rzz:
@erikquinn that was it.

give this a try when you can.

n8rzz:
Yeah, I was afraid of that and didn't think if it till after I pushed. Let me see what I can do with that before we merge.

n8rzz:
@erikquinn ready for another go.

small change here, timewarp should work now at any setting.  However, if you navigate to a new tab and come back, timewarp will be reset to `0`.  What's happening is on `window.blur` we reset the `#game.speedup` prop to 0 which allows us to look at `#game.delta` **_and_** `game.speedup` to determine if we should rest `#game.delta`.  

Its a little hackey, but I've added comments on why and it allows us to avoid setting transitional properties like `isTransitioningFromBlur` etc.

erikquinn:
> small change here, timewarp should work now at any setting.  However, if you navigate to a new tab and come back, timewarp will be reset to `0`.

I don't _love_ that to be honest, but it seems like it may be a bit of a headache to keep the timewarp going _and_ get it to do what it should on refocus. Is that the case? Because if it's anything other than a painless fix to also maintain a persistent timewarp, I'm good with just merging it and checking the box for fixing the problem. Either way.

n8rzz:
>Is that the case

@erikquinn  Yep

erikquinn:
Royer! :shipit:

PR #708, opened 2017-09-19
enhancement/707

### LINKED ISSUE #707: Initial CanvasController refactor
this is a precursor to existing issue #236.  There are several _'quick wins'_ that we can do to increase performance.

- refactor:
  - `$.each()` 
  - `_forEach()`
- consolidate math operations
- move some logic out of render cycle
- create separate canvases for moving elements (aircraft) and non-moving elements (everything else)

n8rzz:
Resolves #707 

Initial canvas controller refactor

_depends on #706 (#296)_


src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -140,76 +239,68 @@ export default class CanvasController {
...
-        this.canvas.draw_restricted = false;
-        this.canvas.draw_sids = false;
-        this.canvas.draw_terrain = true;
+        this.canvas.last = -1;
+        this._shouldShallowRender = true;
+        this._shouldDeepRender = true;
+        this._shouldDrawLabels = false;
+        this._shouldDrawRestrictedAreas = false;
+        this._shouldDrawSidMap = false;
+        this._shouldDrawTerrain = true;
```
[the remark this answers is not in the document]
n8rzz: we have been storing these values in the `CanvasController`, I simply gave them better names.  This is the source of truth for these value, whenever one of the control buttons is clicked, we trigger an event that updates the value in the `CanvasController`.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -225,40 +253,35 @@ export default class CanvasController {
...
+        this.canvas_add(CANVAS_NAME.AIRCRAFT);
+        this.canvas_add(CANVAS_NAME.AIRSPACE);
+        this.canvas_add(CANVAS_NAME.BACKGROUND);
+        this.canvas_add(CANVAS_NAME.NAVAIDS);
+        this.canvas_add(CANVAS_NAME.TERRAIN);
+        this.canvas_add(CANVAS_NAME.PROCEDURE_PATH);
+        this.canvas_add(CANVAS_NAME.RESRICTED_AIRSPACE);
+        this.canvas_add(CANVAS_NAME.RUNWAYS);
+        this.canvas_add(CANVAS_NAME.VIDEO_MAP);
+        this.canvas_add(CANVAS_NAME.WIND_VANE);
```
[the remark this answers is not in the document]
n8rzz: yep, still working that part out.  It'll happen during a `deepRender` cycle where we'll update pan and zoom on all the canvases.

At least thats my plan 😬

src/assets/scripts/client/canvas/CanvasController.js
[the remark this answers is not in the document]
n8rzz: oh, and doc blocks are coming for everything in the `CanvasController`.  I just haven't filled them all in yet 😉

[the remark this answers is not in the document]
erikquinn: > This is the source of truth for these value, whenever one of the control buttons is clicked, we trigger an event that updates the value in the CanvasController.

That does seem a bit odd though, wouldn't you agree? Them toggle buttons are a part of the UI, and I would think their state would be stored as a property of the UiController, and then accessed from the canvas controller via `UiController.shouldDrawLabels` etc.

> oh, and doc blocks are coming for everything in the CanvasController. I just haven't filled them all in yet 😉

oh thank GOD for that 😌

src/assets/scripts/client/AppController.js
```diff
@@ -279,7 +279,7 @@ export default class AppController {
     updateViewControls() {
         const { current: airport } = AirportController;
 
-        this.canvasController.canvas.dirty = true;
+        this._eventBus.trigger(EVENT.MARK_DIRTY_CANVAS);
```
erikquinn: gotta love that event bus! this will be _SUCH_ a better way to deal with that sort of thing.

src/assets/scripts/client/UiController.js
```diff
@@ -628,7 +642,8 @@ class UiController {
      */
     canvas_terrain_toggle(event) {
         $(event.target).closest(SELECTORS.DOM_SELECTORS.CONTROL).toggleClass(SELECTORS.CLASSNAMES.ACTIVE);
-        prop.canvas.draw_terrain = !prop.canvas.draw_terrain;
+
+        this._eventBus.trigger(EVENT.SHOULD_TOGGLE_TERRAIN);
```
erikquinn: how about `EVENT.TOGGLE_TERRAIN` instead of should?

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -176,6 +176,16 @@ export default class AirportModel {
          */
         this.airspace = null;
 
+        // TODO: this should really be its own class possibly separate from the `AirportModel`
```
erikquinn: ^^ amen. one day :)

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -283,68 +404,52 @@ export default class CanvasController {
...
-            // TODO: this is incorrect usage of a ternary. ternaries should be used for a ssignment not function calls.
-            // draw airspace border
-            AirportController.airport_get().airspace
-                ? this.canvas_draw_airspace_border(cc)
-                : this.canvas_draw_ctr(cc);
+
+            if (AirportController.airport_get().airspace) {
+                this.canvas_draw_airspace_border(cc);
+            } else {
+                this.canvas_draw_ctr(cc);
```
erikquinn: FWIW, a `airspace` property is now guaranteed by design. This is left over from when we could have circular airspaces around the center of the airport, and only define the radius. Now we must have GPS coordinates for the airspace border. So this conditional and the `canvas_draw_ctr()` could definitely be removed if you're so inclined.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -283,68 +404,52 @@ export default class CanvasController {
...
-            if (this.canvas.dirty || this.canvas_should_draw() || true) {
+            if (this.canvas_should_draw() || true) {
                 cc.save();
                 cc.globalAlpha = alpha;
-                cc.translate(
-                    calculateMiddle(this.canvas.size.width),
-                    calculateMiddle(this.canvas.size.height)
-                );
+                cc.translate(middleWidth, middleHeight);
                 this.canvas_draw_all_aircraft(cc);
```
erikquinn: Generally speaking, it's quite silly we've been passing the canvas context around between every method involved in the draw. I assume you have plans to do so, but it would make a lot of sense to have the 2d context as a property of the CanvasController itself so we can instead call the methods without any arguments, and they can just do `this._ctx` or something to get it.

n8rzz: yes it is a little awkward, however, it means we only need to get the `context` once.  And, we are actually storing it too.  `this.canvas.contexts[CANVAS_NAME].context` is where that lives.  for now, it's just easier grabbing it once and passing it around.

This will likely change, though, as I make my way through these changes.

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: excellent 🎉

n8rzz:
@erikquinn there are so many little things I want to tackle with this but it would quickly get out of hand.  I'm going to stop here with major changes and begin addressing some of the items you've posted above.

Work on the canvases will continue after #726 and #725 are completed under a new issue.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -52,45 +51,212 @@ import { LOG } from '../constants/logLevel';
...
+        /**
+         * Enum used for storing canvas size values
+         *
+         * @property size
+         * @type {object}
+         */
         this.canvas.size = {
             height: DEFAULT_CANVAS_SIZE.HEIGHT,
             width: DEFAULT_CANVAS_SIZE.WIDTH
         };
```
erikquinn: Why not just `this.canvas.height` and `this.canvas.width`?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -132,108 +321,127 @@ export default class CanvasController {
...
-
-        $(hidefCanvas).attr('width', w * dpr);
-        $(hidefCanvas).attr('height', h * dpr);
-        $(hidefCanvas).css('width', w);
-        $(hidefCanvas).css('height', h);
+    canvas_add(name) {
+        const canvasTemplate = `<canvas id='${name}-canvas'></canvas>`;
 
-        const ctx = hidefCanvas.getContext('2d');
+        $(SELECTORS.DOM_SELECTORS.CANVASES).append(canvasTemplate);
```
erikquinn: can we do this with a property created on instantiation from `$element` rather than using jQuery here?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -686,55 +888,53 @@ export default class CanvasController {
...
      * @for CanvasController
      * @method canvas_draw_separation_indicator
-     * @param cc
-     * @param aircraft
+     * @param cc {HTMLCanvasContext}
+     * @param aircraft {AircraftModel}
      */
     canvas_draw_separation_indicator(cc, aircraft) {
-        if (!GameController.shouldUseTrailingSeparationIndicator(aircraft)) {
+        if (!GameController.shouldUseTrailingSeparationIndicator(aircraft) || !this.theme.RADAR_TARGET.TRAILING_SEPARATION_INDICATOR_ENABLED) {
```
erikquinn: This is already be included in the `.shouldUseTrailingSeparationIndicator()` method, so there's no need for the addition here.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1356,13 +1568,12 @@ export default class CanvasController {
         cc.textBaseline = 'top';
 
         for (let i = 90; i <= 360; i += 90) {
+            let angle = i;;
```
erikquinn: double-semi!

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: Yes we can!  Good catch. 

In the future I think we can just find the elements in the DOM instead of creating them like this. I'd like to move away from dynamically creating DOM elements unless we absolutely have to.

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: This block is mostly as I found it, I just removed an unnecessary parenthesis. This will be simplified in a future PR, I just didn't want to add more to the pile for this PR.

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: 😱

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: I agree!  I found it like this. This will be going away under #726

src/assets/scripts/client/UiController.js
```diff
@@ -520,11 +529,10 @@ class UiController {
 
     /**
      * @for UiController
-     * @method ui_after_zoom
+     * @method storeZoomLevel
      */
-    ui_after_zoom() {
+    storeZoomLevel() {
         localStorage[STORAGE_KEY.ATC_SCALE] = this.scale;
```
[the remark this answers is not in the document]
n8rzz: done

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: right, not the block i'm concerned about, just the if statement. you added an or that is redundant. it should read what it was before:
`if (!GameController.shouldUseTrailingSeparationIndicator(aircraft)) {`

src/assets/scripts/client/UiController.js
```diff
@@ -653,7 +668,7 @@ class UiController {
     canvas_labels_toggle(event) {
         $(event.target).closest(SELECTORS.DOM_SELECTORS.CONTROL).toggleClass(SELECTORS.CLASSNAMES.ACTIVE);
 
-        prop.canvas.draw_labels = !prop.canvas.draw_labels;
+        this._eventBus.trigger(EVENT.SHOULD_TOGGLE_LABELS);
```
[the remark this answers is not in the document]
n8rzz: done

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -52,45 +51,212 @@ import { LOG } from '../constants/logLevel';
...
+         * @property panY
+         * @type {number}
+         * @default 0
+         */
         this.canvas.panY = 0;
+
+        // TODO: this property will be moving out of the `CanvasController` under issue:
+        // [#726](https://github.com/openscope/openscope/issues/726)
+        /**
+         * @property panX
```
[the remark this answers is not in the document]
n8rzz: done

src/assets/scripts/client/canvas/CanvasController.js
n8rzz: and I don't like this block, either.  the sooner we can get rid of it (as confusing as it is) the better!

n8rzz: seems my espresso hasn't quite kicked in yet, I was thinking this was an entirely different block.

**_this_** block is actually a combination of two if statements.  Isn't it possible for the theme to specify yes and the `GameController` no?

n8rzz:
@erikquinn added more documentation.  Only one item left from your initial feedback:
- https://github.com/openscope/openscope/pull/708#discussion_r140960124

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: I know there was a second if statement below this that was consolidated into the line being commented on. What I'm trying to say is the if statement that was moved should have been _REmoved_ altogether. It is already being handled in the `.shouldUseTrailingSeparationIndicator()` method:

```
/**
 * Check whether or not the trailing distance separator should be drawn.
 *
 * Used by the `CanvasController` to determine whether or not to proceed with
 * `canvas_draw_separation_indicator`.
 *
 * @for GameController
 * @method shouldUseTrailingSeparationIndicator
 * @return {boolean}
 */
shouldUseTrailingSeparationIndicator(aircraft) {
    const userSettingsValue = this.getGameOption(GAME_OPTION_NAMES.DRAW_ILS_DISTANCE_SEPARATOR);
    let isIndicatorEnabled = userSettingsValue === 'yes';

    if (userSettingsValue === 'from-theme') {
        isIndicatorEnabled = this.theme.RADAR_TARGET.TRAILING_SEPARATION_INDICATOR_ENABLED;
    }

    return isIndicatorEnabled && aircraft.isArrival();
}
```

n8rzz: marvelous.  I'll make that change!

n8rzz: done

PR #713, opened 2017-09-20
feature/709

### LINKED ISSUE #709: Ensure "last airport" exists before attempting to load it
During initialization of the app, check local storage for `localstorage[STORAGE_KEY.ATC_LAST_AIRPORT]` (the one we will subsequently attempt to load). If it is _not present in the load list_, that means there will be no airport file to load, and the app will never make it through the initialization process. If this is the case, reset the local storage (either the last airport value to the default airport), or all of the values... whatever it takes to be _sure_ that users that had an airport cached that no longer exists will not have the site stuck in a state where it can never complete the load without manually flushing the cache.

sam-irl:
Resolves #709 

The purpose of this pull request is to check if airport exists in localStorage before attempting to load it

### DIFF
--- a/src/assets/scripts/client/index.js
+++ b/src/assets/scripts/client/index.js
@@ -12,12 +12,8 @@ import { STORAGE_KEY } from './constants/storageKeys';
 const getInitialAirport = () => {
     let airportName = DEFAULT_AIRPORT_ICAO;
 
-    if (_has(localStorage, STORAGE_KEY.ATC_LAST_AIRPORT)) {
+    if (_has(localStorage, STORAGE_KEY.ATC_LAST_AIRPORT) && AirportController.hasAirport(STORAGE_KEY.ATC_LAST_AIRPORT)) {
         airportName = _lowerCase(localStorage[STORAGE_KEY.ATC_LAST_AIRPORT]);
-
-        if (!AirportController.hasAirport(airportName)) {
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/index.js
```diff
@@ -13,6 +14,9 @@ const getInitialAirport = () => {
 
     if (_has(localStorage, STORAGE_KEY.ATC_LAST_AIRPORT)) {
         airportName = _lowerCase(localStorage[STORAGE_KEY.ATC_LAST_AIRPORT]);
+        if (!AirportController.hasAirport(airportName)) {
```
n8rzz: be sure to add an empty line above `if` statements

src/assets/scripts/client/index.js
```diff
@@ -13,6 +14,10 @@ const getInitialAirport = () => {
 
     if (_has(localStorage, STORAGE_KEY.ATC_LAST_AIRPORT)) {
         airportName = _lowerCase(localStorage[STORAGE_KEY.ATC_LAST_AIRPORT]);
+
+        if (!AirportController.hasAirport(airportName)) {
```
erikquinn: We can simplify this logic by combining the conditions of these two if statements. We are initially setting to the default airport, and then _if an airport is in local storage and it is accessible, THEN we look for that airport_.

Issue #21, opened 2017-01-03
Command Reference


erikquinn:
Particularly as we get more and more commands, it would be nice to have a way to see a reference sheet that would remind you the parameters required under each command while in-game.

Idea from @sam-irl was an autocomplete-esque thing where the text box would show in faint color the arguments available for the given command (such as typing `AAL123 turn` and after that it showing a faint `[L/R] [magnetic heading]` to clue people in on what the parmeters are for the `turn` command). 

Ideally, a full solution to this issue would include _both_ the above behavior in the text box as well as a list of commands that users can show/hide when they need to see what commands are available, what the differences are, etc.

Originally reported in zlsa/atc#463.

n8rzz:
@erikquinn I suppose the wiki page wouldn't count as a command reference?

erikquinn:
@n8rzz It counts as a thorough how-to, but sometimes a shortlist would be nice, for example, to remind oneself what command you use to "tell that plane to skip straight to WETOR" (`pd`) via a small in-game cheat sheet.

ron-eo:
Is there a command that will bring your airport back to the center of the screen? I'm in empty space and feeling lost! :(

indianbhaji:
Try double clicking an arriving flight on the flight strip - don't try this with a departure

ron-eo:
Excellent! Thank you, indianbhaji! Something that would be useful in a command reference, that's for sure! Cheers ! :)

scott-j-smith:
Actually, I do that with departures, but only after they are already waiting at the runway.

n8rzz:
we did fix that bug this sprint (#207) that would break the app when double clicking on an aircraft strip for an aircraft that hasn't taxied.

n8rzz:
@suhaildawood this story would be included with any sort of UI updating.  Just calling this one out for you because its old and you may not have seen it 😉

PR #716, opened 2017-09-21
enhancement/21

sam-irl:
Resolves openscope/openscope#21 

The purpose of this pull request is to add an in game command reference


sam-irl:
`help altitude`. Maybe for the syntax there could be a thing in the command bar? So if I type:
`AAL123 a`
the command bar shows
`AAL123 a [altitude]`

erikquinn:
Unfortunately, this isn't quite what we're looking to do with #21. Either a short list of available commands that can be shown/hidden, or your great idea of an autocomplete-esque thing where the text box would show in faint color the arguments available for the given command (such as typing `AAL123 turn` and after that it showing a faint `[L/R] [magnetic heading]` to clue people in on what the parmeters are for the `turn` command. Ideally, a full solution to #21 would include _both_ of these elements.

Closing this PR.

PR #722, opened 2017-09-22
refactor/721

### LINKED ISSUE #721: Remove `version` command
The `version` command is using `prop.version`, which is no longer maintained, so when it attempts to log the version number, it instead says:

image

The only place the version number is stored now is in `package.json`, which ends up in `layout.hbs` and `loadingView.hbs`. Somehow we should get this command printing the correct version number.

sam-irl:
Resolves #721 

The purpose of this pull request is to remove the version command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
     - updates styles to use flexbox with properly organized children
 
 
+
 ### Enhancements & Refactors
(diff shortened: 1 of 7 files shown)


erikquinn:
Changing from `bugfix` to `refactor`, since instead of fixing the bug with the command, we're just removing the command! 😉

PR #731, opened 2017-09-26
documentation/29

### LINKED ISSUE #29: Airport Guide
Some background information should be made available for each airport, in order for users to have access to the diagrams, explanations, limitations on procedures, airspace boundaries, and other information required to understand how to effectively work in that particular airspace. This would be achieved through `*.md` files, as well as links to further readings, and charts, etc.

Additionally, each page should contain some links to relevant charts, diagrams, etc.

Related: #35.
Originally reported in zlsa/atc#612.

erikquinn:
Resolves #29.

Add airport guide files and start ksea as an example

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
 
 An ATC simulator in HTML5. Visit http://openscope.co to begin playing now!
 
+For more information on each airport, see the [airport guide](documentation/airport-guides/airport-guide-directory.md).
+
 
 ## Quick Start
 
--- a/documentation/airport-guides/ksea.md
+++ b/documentation/airport-guides/ksea.md
@@ -1,0 +1,74 @@
+### About this facility
+Seattle TRACON (S46), Seattle, Washington, USA
+582,394 annual operations in 2016 ([FAA ATADS](https://aspm.faa.gov/opsnet/sys/Tracon.asp))
+
+Located less than a quarter mile west of SeaTac's Runway 16R-34L, this consolidated approach control facility works all traffic in the Seattle metropolitan area. The facility is fairly new, with a large, open control floor sporting about 15-20 STARS scopes, and an in-house Traffic Management Unit.
+
+### Airspace
+For the most part, S46 airspace extends up to 15,000 feet over a fairly large area surrounding northwestern Washington state. There are various shelves that extend S46's ability to prepare an arrival sequence, including the Olympia and Arlington shelves (SFC-9,000), and the Paine shelf which extends the northern reach of the airspace.
+
+Class B airspace exists with tops of 10,000 over the Seattle-Tacoma metropolitan area, with a fairly complex structure designed to deconflict noncommunicative VFR traffic from IFR arrivals on arrival and departure procedures in and out of Seattle.
+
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -8,6 +8,8 @@
 
 An ATC simulator in HTML5. Visit http://openscope.co to begin playing now!
 
+For more information on each airport, see the [airport guide](documentation/airport guide/airport-guide-directory.md).
```
n8rzz: can you add a dash between airport and guide: `airport-guide`.  spaces are a no-no in urls

documentation/airport guides/ksea.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### KOLM
+Olympia Regional  
+45,838 annual operations  
+37.6nm southwest of KSEA
+
+World War II era airport now open for public use. Has relatively little traffic, much of which are helicopters.
+
+### Additional Resources
+_Add links to relevant videos, files, etc here_
```
n8rzz: would this include links to procedure publications?  how are we going to manage that when the procedures get updated, are previous versions still available for us to link to?

erikquinn: We would probably link to [this](https://flightaware.com/resources/airport/KBFI/APD/AIRPORT+DIAGRAM), for example, which will not break when they update.

README.md
erikquinn: done

documentation/airport guides/ksea.md
n8rzz: true.  but what if, as an example, procedure `GOPHR6` is changed to `GOPHR7`?

erikquinn: we're still good. If you look carefully at that particular page, down the right side there is a list of all procedures. They, as well as the APD will be updated without any need to change the link.

documentation/airport-guides/ksea.md
```diff
@@ -0,0 +1,74 @@
...
+
+Class B airspace exists with tops of 10,000 over the Seattle-Tacoma metropolitan area, with a fairly complex structure designed to deconflict noncommunicative VFR traffic from IFR arrivals on arrival and departure procedures in and out of Seattle.
+
+Of unique note here is the location of Boeing Field (KBFI), which is less than 5nm north of SeaTac (KSEA), right along the approach course to the 16s. For this reason, extra restrictions on localizer intercept altitudes or crossing restrictions are needed to ensure appropriate separation. Particularly when in good visual conditions on a south/east flow, due to the numerous conveniently placed landmarks, controllers heavily utilize visual reference points with their traffic advisories.
+
+### Instrument Procedures
+For an American airport, the procedures are fairly automated, with the design of the SIDs and STARs taking a lot of weight off the controllers. Arrivals climb and descend via the procedures, and by design the traffic flow requires minimal intervention other than running the final. Aircraft are still manually vectored to the final approach course.
```
Fechulo: > Arrivals climb and descend via the procedures  

Should probably be `Aircraft climb and descend via the procedures`

documentation/airport guides/ksea.md
Fechulo: While this will work fine for US airports, it won't work for a lot of other airports. In Europe, for example, the eAIP's URL changes for each AIRAC cycle and some countries require a registration to access the charts. Some countries don't even publish their charts (I'm looking at you Argentina 😠)

documentation/airport-guides/ksea.md
erikquinn: done

documentation/airport guides/ksea.md
erikquinn: yeah, not sure. I guess we'll just have to deal with updating them in those cases ☹️

PR #734, opened 2017-09-27
bugfix/732

### LINKED ISSUE #732: Aircraft do not clear the scope when uncontrolled for a bit
Browser: Chrome

Airport: CYOW (#161)

Approximate play time before issue: When browser is out of focus and plane is outside of controlzone

#### Issue Description
When aircraft are outside control zone (landed or outside of airspace) and the browser loses focus aircraft seem to freeze mid air. The scope can still be moved though

#### Steps to reproduce

1. Use the timewarp button and set it to x5
2. Clear aircraft out of your airspace, keep the browser in focus
3. The aircraft will continue moving outside the airspace but eventually they will not clear the scope and will just stay still in the air

erikquinn:
Resolves #732.

Ensure removal of radar targets on aircraft deletion

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -157,6 +157,7 @@ export default class AircraftController {
         this._stripViewController = new StripViewController();
 
         return this.init()
+            .setupHandlers()
             .enable();
     }
 
@@ -170,6 +171,20 @@ export default class AircraftController {
     }
 
     /**
+     * Set up event handlers
+     *
+     * @for AircraftController
+     * @method _setupHandlers
+     * @private
+     * @chainable
+     */
+    _setupHandlers() {
+        this._onRemoveAircraftHandler = this.aircraft_remove.bind(this);
+
+        return this;
+    }
+
+    /**
      * @for AircraftController
      * @method enable
      * @chainable
@@ -179,7 +194,7 @@ export default class AircraftController {
         this._eventBus.on(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickhandler);
         this._eventBus.on(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this.onSelectAircraftStrip);
         this._eventBus.on(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
-        this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this.aircraft_remove);
+        this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
         this._eventBus.on(EVENT.REMOVE_AIRCRAFT_CONFLICT, this.removeConflict);
 
         return this;
@@ -195,7 +210,7 @@ export default class AircraftController {
         this._eventBus.off(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickhandler);
         this._eventBus.off(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this._onSelectAircraftStrip);
         this._eventBus.off(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
-        this._eventBus.off(EVENT.REMOVE_AIRCRAFT, this.aircraft_remove);
+        this._eventBus.off(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
         this._eventBus.off(EVENT.REMOVE_AIRCRAFT_CONFLICT, this.removeConflict);
 
         return this;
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -381,7 +396,7 @@ export default class AircraftController {
 
                 GameController.events_recordNew(GAME_EVENTS.ARRIVAL);
                 aircraft.setIsRemovable();
-                this.aircraft_remove(aircraft);
+                this._eventBus.trigger(EVENT.REMOVE_AIRCRAFT, aircraft);
```
n8rzz: is this event being listened to elsewhere?  otherwise I'd prefer to just call the method.  There may be instances where this is _triggered_ externally (like when a user issues a `del` command), but I would really like to avoid using the `EventBus` within the same class.  That means, naturally, that we should minimize the number of listeners to a particular event.

We've had this discussion before, I know, I'm just re-iterating 😄

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -179,7 +194,7 @@ export default class AircraftController {
         this._eventBus.on(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickhandler);
         this._eventBus.on(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this.onSelectAircraftStrip);
         this._eventBus.on(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
-        this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this.aircraft_remove);
+        this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
```
n8rzz: very nice!

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: yeah it's listened to by the `AircraftController` _and_ `ScopeModel.radarTargetCollection`. I hear what you're saying though. Here's my only thought:

We can change to calling the local method `aircraft_remove()`, and _within_ that method trigger the event for the external listeners. The only potential pitfall is if we start following this pattern, and later on we want to trigger the `REMOVE_AIRCRAFT` event from a different class (say, `AircraftConflict` when they crash or something), we would naturally choose to trigger the event from within that other class. But then since `aircraft_remove()` is not a listener, we likely won't remember to call it and stuff could get hairy.

Penny for your thoughts on that.

n8rzz: I'm all for `AircraftController` listening to the event, that part shouldn't change.  Other classes will need to be able to trigger a `REMOVE_AIRCRAFT` event.  My concern is using it here, in this block.  Really all we're doing is calling the local method.  And, if the `ScopeModel` is the only other thing listening for the event, my recommendation would be to simply call `ScopeModel.remove()` (or whatever method) directly within `aircraft_remove`.

The high level idea with the`EventBus` is to provide a way for some/multiple classes to trigger an event, but have the event trigger a method in a single class.  That's the idea anyway, and I'm not sure if that's how we have things implemented.  That's how I'd like it to work 😏 

_[caller] -> single source of change_

erikquinn: I guess we actually do have a local reference to the scope model, so we are able to do that. When in cases where we don't though, how would we do it? For example an `AircraftModel` goes too fast and breaks its wings off and then wants to delete itself (granted we would probably deal with this in the aircraft controller, but hey). It won't have access to the scope, so how would we handle it then?

n8rzz: that would be a case for triggering the event.  but since _we do_ have access to the `ScopeModel` here, it makes sense to just call the method.

erikquinn: I have to say though... writing this
```javascript
this.aircraft_remove(aircraft);
this._scopeModel.radarTargetCollection.removeRadarTargetModelForAircraftModel(aircraft);
```

...makes me kinda nervous. Seems like leaving the door open in the future to forgetting to call the right things when "triggering" the removal from other places. I could really see a benefit to registering the events and just saying "trigger this thing" as opposed to manually spelling out the methods needed to execute the event in multiple places. Obviously we don't want _everything_ happening with an eventbus event, but it's a system that really helps to simplify things and I wouldn't mind more of it myself.

erikquinn: Well actually, nevermind. That isn't really the case. What I should be doing here is calling the radar target removal method _within_ `aircraft_remove()`. Then there's still a single method that does all the work, which could be (or not be) invoked by the event bus later on.

erikquinn: ^ and I think that's what you're getting at above as well. That we should avoid having a hundred different event listeners, and rather have one listener on a high-level controller that goes through the list of actions needed for all the individual classes involved. I think I'm catching on now.

n8rzz: now you smell what I'm stepping in 😉 

thats exactly right 👍

erikquinn: fixed

erikquinn:
I was in the process of doing that, but had some problems. For another day I guess.

PR #736, opened 2017-09-27
bugfix/711

### LINKED ISSUE #711: Game Options not being stored correctly
Browser: Chrome
Airport: All

#### Issue Description
openscope/openscope#690 (refactor #670) appears to have caused some issues with the local storage of game options. As can currently be seen on the dev server, game options are not preserved, and are defaulted to the _first_ option in the list, rather than the default specified in `gameOptionConstants.js`.

#### Steps to reproduce
- Clear cache and load dev server
- Observe that the default values are not honored
- ~~Change some settings and refresh~~
- ~~Observe that the changes to options are not preserved~~

erikquinn:
Resolves #711.

Ensure game options initialize to correct default values.

### DIFF
--- a/src/assets/scripts/client/game/GameOptions.js
+++ b/src/assets/scripts/client/game/GameOptions.js
@@ -1,3 +1,4 @@
+import _isNil from 'lodash/isNil';
 import EventBus from '../lib/EventBus';
 import { GAME_OPTION_VALUES } from '../constants/gameOptionConstants';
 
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/game/GameOptions.js
```diff
@@ -66,7 +66,7 @@ export default class GameOptions {
         this._options[optionProps.name] = optionProps;
         let optionValue = optionProps.defaultValue;
 
-        if (typeof storedOptionValue !== 'undefined') {
+        if (!_isNil(storedOptionValue)) {
```
n8rzz: wait, what, really?  In the instance a key doesn't exist in `localStorage` it returns `undefined`, right?

can you add a test case for this?

erikquinn: it's returning `null` instead of `undefined` in all cases i was seeing.

erikquinn: so the isNil will cover both.

erikquinn: done

PR #746, opened 2017-09-28
hotfix/744

### LINKED ISSUE #744: Logic for checking existence of a localStorage icao should be moved from index.js
The original solution implemented in openscope/openscope#713 is incorrect.  I should have caught it in the PR, but I missed it.  That's my fault.

The original solution had us checking for an airports existence in the `AirportController` within the `index.js` file on initial load.  This will not work.  `index.js` is the initial JS file the creates the entire app and when it runs the `AirportController` does not yet exist.  

This has now created issue #741.

The solution was correct in spirit, it just needs to happen in a different place later on in the initialization lifecycle.

sam-irl:
Resolves openscope/openscope#744 

The purpose of this pull request is to check if an airport exists before loading it, this time for *real*, I swear

### DIFF
--- a/src/assets/scripts/client/App.js
+++ b/src/assets/scripts/client/App.js
@@ -85,8 +85,8 @@ export default class App {
         )
             .done((airportResponse, airlineResponse, aircraftResponse) => {
                 const initialAirport = typeof airportResponse[0] !== 'undefined' 
-                ? airportResponse[0] 
-                : null;
+                    ? airportResponse[0] 
+                    : null;
 
                 this.setupChildren(
                     airportLoadList,
--- a/src/assets/scripts/client/App.js
+++ b/src/assets/scripts/client/App.js
@@ -84,7 +84,9 @@ export default class App {
             $.getJSON('assets/aircraft/aircraft.json')
         )
             .done((airportResponse, airlineResponse, aircraftResponse) => {
-                const initialAirport = typeof airportResponse[0] !== 'undefined' ? airportResponse[0] : null;
+                const initialAirport = typeof airportResponse[0] !== 'undefined' 
+                ? airportResponse[0] 
+                : null;
 
                 this.setupChildren(
                     airportLoadList,
--- a/src/assets/scripts/client/App.js
+++ b/src/assets/scripts/client/App.js
@@ -76,6 +76,22 @@ export default class App {
      * @param airportLoadList {array<object>}  List of airports to load
      */
     initiateDataLoad(airportLoadList, initialAirportToLoad) {
+        // Check if the json we try to load exists
+        // If it doesn't, return early. Otherwise the program hangs.
+        if ($.getJSON(`assets/airports/${initialAirportToLoad.toLowerCase()}.json`) == undefined) {
+            console.error(`couldn't find json for ${initialAirportToLoad} (cleared storage key)`);
+            localStorage.removeItem(STORAGE_KEY.ATC_LAST_AIRPORT);
+            return;
+        }
+        if ($.getJSON('assets/airlines/airlines.json') == undefined) {
+            console.error('couldn\'t find json for airlines, are all files present?');
+            return;
+        }
+        if ($.getJSON('assets/aircraft/aircraft.json') == undefined) {
+            console.error('couldn\'t find json for aircraft, are all files present?');
+            return;
+        }
+
         // This is provides a way to get async data from several sources in the app before anything else runs
         // TODO: this is wrong. move this and make it less bad!
         $.when(
--- a/src/assets/scripts/client/App.js
+++ b/src/assets/scripts/client/App.js
@@ -1,6 +1,5 @@
 import $ from 'jquery';
 import AppController from './AppController';
-import AirportController from './airport/AirportController';
 import EventBus from './lib/EventBus';
 import TimeKeeper from './engine/TimeKeeper';
 import { EVENT } from './constants/eventNames';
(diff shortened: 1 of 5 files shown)


sam-irl:
@n8rzz I've only just seen #745, whoops! Close if needed

src/assets/scripts/client/App.js
```diff
@@ -31,6 +32,11 @@ export default class App {
      * @param initialAirportToLoad {string}    ICAO id of the initial airport. may be the default or a stored airport
      */
     constructor(element, airportLoadList, initialAirportToLoad) {
+        // We need to put this in App, because it doesn't work in index.js
```
n8rzz: this is still the incorrect spot.  at this point in the app, `AirportController` still hasn't been hydrated with airports.

this logic should happen somewhere around here: https://github.com/openscope/openscope/blob/096f38934a9d6798df6f1e621b5120b0901b8e2a/src/assets/scripts/client/airport/AirportController.js#L132

sam-irl: @n8rzz is this already implemented?

https://github.com/openscope/openscope/blob/096f38934a9d6798df6f1e621b5120b0901b8e2a/src/assets/scripts/client/airport/AirportController.js#L139

n8rzz: the logic above, yes that already exists in `develop`

hmm...

I think we have a different problem entirely here.  I think, in fact, this: https://github.com/openscope/openscope/blob/096f38934a9d6798df6f1e621b5120b0901b8e2a/src/assets/scripts/client/App.js#L88 is where are problems are coming from.

The initial issue described a browser hang, that line would do it when an airport file doesn't exist.

src/assets/scripts/client/App.js
```diff
@@ -82,6 +76,22 @@ export default class App {
      * @param airportLoadList {array<object>}  List of airports to load
      */
     initiateDataLoad(airportLoadList, initialAirportToLoad) {
+        // Check if the json we try to load exists
+        // If it doesn't, return early. Otherwise the program hangs.
+        if ($.getJSON(`assets/airports/${initialAirportToLoad.toLowerCase()}.json`) == undefined) {
```
n8rzz: I think you're starting down the right track here.  the problem with this implementation is you've now doubled the  network requests by making duplicate calls.

since we know that airlines and aircraft will always exist, we may be able to do:
```javascript
// within .done() 
const initialAirport = typeof airportResponse[0] !== 'undefined' ? airportResponse[0] : null;

this.setupChildren(
  airportLoadList,
  initialAirport,
  airlineResponse[0].airlines,
  aircraftResponse[0].aircraft
);
```

the only problem with that is I don't know if we hit the `.fail()` method if `initialAirportToLoad` comes back undefined.  might be something to try though?

src/assets/scripts/client/App.js
```diff
@@ -100,9 +84,11 @@ export default class App {
             $.getJSON('assets/aircraft/aircraft.json')
         )
             .done((airportResponse, airlineResponse, aircraftResponse) => {
+                const initialAirport = typeof airportResponse[0] !== 'undefined' ? airportResponse[0] : null;
```
n8rzz: oh I like this!  I'm spinning up a review app to check this out.  But I think this could be a winner.

small style gripe:
```javascript
// instead of
const initialAirport = typeof airportResponse[0] !== 'undefined' ? airportResponse[0] : null;

// please use
const initialAirport = typeof airportResponse[0] !== 'undefined' 
    ? airportResponse[0] 
    : null;
```

if just makes the ternaries a little easier to read.   I'm not a huge fan of them to begin with, but they are useful in cases such as this.

src/assets/scripts/client/App.js
```diff
@@ -84,7 +84,9 @@ export default class App {
             $.getJSON('assets/aircraft/aircraft.json')
         )
             .done((airportResponse, airlineResponse, aircraftResponse) => {
-                const initialAirport = typeof airportResponse[0] !== 'undefined' ? airportResponse[0] : null;
+                const initialAirport = typeof airportResponse[0] !== 'undefined' 
+                ? airportResponse[0]
```
n8rzz: so close!

the lines after the ternary conditional should be indented 4 spaces:
```javascript
const initialAirport = typeof airportResponse[0] !== 'undefined' 
    ? airportResponse[0] 
    : null;
```

src/assets/scripts/client/index.js
```diff
@@ -3,30 +3,19 @@ require('raf').polyfill();
...
-};
-
 /**
  * Entry point for the application.
  *
  * Provides a way to grab the `body` element of the document and pass it to the app.
  */
 export default (() => {
     const airportLoadList = window.AIRPORT_LOAD_LIST;
-    const initialAirportToLoad = getInitialAirport();
```
n8rzz: we still need this function in order to get any `localStorage` value.  the new code you've added within the `$.when()` should hopefully handle cases where that value doesn't exist.

n8rzz:
@sam-irl so I went ahead and played with this a little more.  It seems the problems were a little deeper than I originally anticipated.  I have had re-worked my original implementation quite a bit to get things working correctly again.  I think it makes more sense to do it this way, though.

n8rzz:
Mmm, yeah, those changes are going break things with the tests. I'll get those fixed soon.

src/assets/scripts/client/App.js
```diff
@@ -62,6 +63,9 @@ export default class App {
      */
     createHandlers() {
         this.eventBus.on(EVENT.PAUSE_UPDATE_LOOP, this.updateRun);
```
erikquinn: should this be in an `.enable()` method?

If it stays here, will need a blank line below to separate from the property assignments

src/assets/scripts/client/App.js
```diff
@@ -62,6 +63,9 @@ export default class App {
...
+        this.onLoadStorageIcaoFailureHandler = this.onLoadStorageIcaoFailure.bind(this);
+        this.onLoadAirlinesAndAircraftHandler = this.onLoadAirlinesAndAircraft.bind(this);
+        this.setupChildrenHandler = this.setupChildren.bind(this);
```
erikquinn: wouldn't we want our handlers to be private?

n8rzz:
closing in favor of hotfix branch #753

src/assets/scripts/client/App.js
n8rzz: it should, yes, but I have a hunch we use it before that method is called and I wanted to keep the changes here directed at our issue.

src/assets/scripts/client/App.js
n8rzz: this is the top level file, so they are all in a sense public.  the only exception is `_init()` which has both a private and public version in this file.

PR #747, opened 2017-09-28
bugfix/742

### LINKED ISSUE #742: 'No such aircraft' response is not red
Browser: Chrome
Airport: All

#### Issue Description
`No such aircraft, say again` is not red, and it should be.

#### Steps to reproduce
- Press enter in aircraft command bar with nothing in it.

sam-irl:
Resolves #742 

The purpose of this pull request is to make the `no such aircraft, say again` response red


sam-irl:
wrong branch, see #748

PR #753, opened 2017-10-01
hotfix/744

### LINKED ISSUE #744: Logic for checking existence of a localStorage icao should be moved from index.js

n8rzz:
Resolves #744 

The purpose of this pull request is to:
Adds graceful handling to failures that occur after attempting to load an airport using a localStorage icao value for an airport that may not exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,3 @@
-## 5.6.0 (November 1, 2017)
----
-### New Features
-
-
-
-### Bugfixes
-
-
-
-### Enhancements & Refactors
-
-
 ## 5.5.1 (October 1, 2017)
 --
 ### Hotfix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 --
 ### Hotfix
 - Checks if the airport in localStorage exists before loading it [#709](https://github.com/openscope/openscope/issues/709) and [#744](https://github.com/openscope/openscope/issues/744)
-- Updates `.eslint` rules to ignore unused `event` and `error` parameters. [#710](https://github.com/openscope/openscope/issues/710) and [#744]
+- Updates `.eslint` rules to ignore unused `event` and `error` parameters. [#710](https://github.com/openscope/openscope/issues/710) and [#744](https://github.com/openscope/openscope/issues/744)
 
 
 ## 5.5.0 (October 1, 2017)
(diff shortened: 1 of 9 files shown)


n8rzz:
@sam-irl I added you as a reviewer on this one because I want you to see how I ended up solving this.  It is quite a bit different that what I had originally though we would need to do.  But it works, and in the process, cleans things up a little bit.

I don't normally hi-jack branches, I'm sorry I did that.  In this case I had a thought it would work and just went ahead and implemented it.  1,000 apologies.

CHANGELOG.md
```diff
@@ -1,9 +1,28 @@
...
+
+
+### Enhancements & Refactors
+
+
+## 5.5.1 (October 1, 2017)
+--
+### Hotfix
+- Checks if the airport in localStorage exists before loading it [#709](https://github.com/openscope/openscope/issues/709) and [#744](https://github.com/openscope/openscope/issues/744)
+- Updates `.eslint` rules to ignore unused `event` and `error` parameters. [#710](https://github.com/openscope/openscope/issues/710) and [#744]
```
sam-irl: Should have a link here, not just a bracketed number

sam-irl:
No problem with the branch hijack! I wouldn't have been able to solve this (at least, not before v5.6.0) anyway. I haven't had any experience with JQuery or JavaScript before I started here a few months ago, and I'm learning lots!

sam-irl:
whoops, didn't mean to approve. obviously wait for @erikquinn

n8rzz:
@sam-irl good catch!  don't know how I missed that :man_facepalming:

n8rzz:
@sam-irl these jquery methods are using a thing called `promises` under the hood.  Promises are a good concept to gain an understanding of.  This article may be helpful in that regard: https://davidwalsh.name/promises.

Another thing to call out is the `setupHandlers()` method and what is going on there.  

Javascript gets a little fussy with scope and the context of `this`.  If would have called the methods directly from within those callbacks, it would puke because the context of `this` is that of the callback and not the class itself.  using that `.bind(this)` stuff locks in the scope of `this` to the class and gives us that handler method to use wherever we choose.

erikquinn:
@sam-irl You can approve things! If your opinion is being sought and you have no objections, go ahead and approve it. In most cases, it won't be merged based only on your approval, but it is helpful feedback for all involved to see that you don't see anything wrong with it. :smiley:

PR #757, opened 2017-10-02
documentation/756

### LINKED ISSUE #756: extraneous '--' in CHANGELOG

sam-irl:
Resolves #756 

The purpose of this pull request is to remove a '--' from the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
 
 
 ## 5.5.1 (October 1, 2017)
---
 ### Hotfix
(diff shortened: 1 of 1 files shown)


erikquinn:
@sam-irl This follows the same pattern used in the rest of the file, so I'm not sure i follow?

sam-irl:
@erikquinn v5.5.0: 
image
v5.5.1:
image

erikquinn:
Just being a _single character, in the changelog_, I really don't think necessitates all this. We can just take care of it in another PR. I know we always say one issue to one pull request, but to me this feels too literal an interpretation of that, and just generally over the top. I mean, we wouldn't want to make a changelog entry about fixing the changelog, and therefore, I think we should just do this elsewhere.

Will be done in #766. Closing.

PR #762, opened 2017-10-03
enhancement/4

### LINKED ISSUE #4: Minor variations in wind
Some degree of variation in the wind would be nice and realistic, rather than it being static, as defined in the airport's json file. The variations would be minor, but ranging in direction maybe 30 degrees, and in velocity about +/- 5 knots. Would add a nice bit of variety when takeoff clearances are read back, and generally make it feel more realistic.

Related: zlsa/atc#68.

sam-irl:
Resolves #4 

The purpose of this pull request is to add minor wind variations

### DIFF
--- a/src/assets/scripts/client/airport/AirportModel.js
+++ b/src/assets/scripts/client/airport/AirportModel.js
@@ -8,6 +8,7 @@ import _map from 'lodash/map';
 import AirportController from './AirportController';
 import EventBus from '../lib/EventBus';
 import GameController from '../game/GameController';
+import WindController from '../weather/WindController';
 import AirspaceModel from './AirspaceModel';
 import DynamicPositionModel from '../base/DynamicPositionModel';
 import RunwayCollection from './runway/RunwayCollection';
@@ -520,8 +521,8 @@ export default class AirportModel {
             return;
         }
 
-        this.wind.speed = currentWind.speed;
-        this.wind.angle = degreesToRadians(currentWind.angle);
+        this.wind.speed = WindController.updateWindSpeed(currentWind.speed);
+        this.wind.angle = degreesToRadians(WindController.updateWindAngle(currentWind.angle));
     }
 
     /**
@@ -553,19 +554,6 @@ export default class AirportModel {
      */
     getWind = () => {
         return this.wind;
-
-        // TODO: leaving this method here for when we implement changing winds. This method will allow for recalculation of the winds?
-        // TODO: there are a lot of magic numbers here. What are they for and what do they mean? These should be enumerated.
-        // const wind = Object.assign({}, this.wind);
-        // let s = 1;
-        // const angle_factor = sin((s + GameController.game_time()) * 0.5) + sin((s + GameController.game_time()) * 2);
-        // // TODO: why is this var getting reassigned to a magic number?
-        // s = 100;
-        // const speed_factor = sin((s + GameController.game_time()) * 0.5) + sin((s + GameController.game_time()) * 2);
-        // wind.angle += extrapolate_range_clamp(-1, angle_factor, 1, degreesToRadians(-4), degreesToRadians(4));
-        // wind.speed *= extrapolate_range_clamp(-1, speed_factor, 1, 0.9, 1.05);
-        //
-        // return wind;
     };
 
     /**
@@ -642,26 +630,16 @@ export default class AirportModel {
         return this._runwayCollection.getRunwayRelationshipForRunwayNames(primaryRunwayName, comparatorRunwayName);
     }
 
-    // TODO: Implement changing winds, then bring this method back to life
     /**
      * @for AirportModel
      * @method updateRunway
      */
     updateRunway() {
-        // const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(this.getWind);
-        //
-        // this.setArrivalRunway(bestRunwayForWind);
-        // this.setDepartureRunway(bestRunwayForWind);
-    }
+        const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(this.getWind);
 
-    // TODO: leaving this here for when we implement variable winds
-    // /**
-    //  * @for AirportModel
-    //  * @method setRunwayTimeout
-    //  */
-    // setRunwayTimeout() {
-    //     this.timeout.runway = GameController.game_timeout(this.updateRunway, Math.random() * 30, this);
-    // }
+        this.setArrivalRunway(bestRunwayForWind);
+        this.setDepartureRunway(bestRunwayForWind);
+    }
 
     /**
      * Return a `RunwayModel` for the provided name
--- a/src/assets/scripts/client/canvas/CanvasController.js
+++ b/src/assets/scripts/client/canvas/CanvasController.js
@@ -448,10 +448,12 @@ export default class CanvasController {
      * @method canvas_update_post
      */
     canvas_update_post() {
-        const elapsed = GameController.game_time() - AirportController.airport_get().start;
+        const airport = AirportController.airport_get();
+        const elapsed = GameController.game_time() - airport.start;
         const alpha = extrapolate_range_clamp(0.1, elapsed, 0.4, 0, 1);
         const framestep = Math.round(extrapolate_range_clamp(1, GameController.game.speedup, 10, 30, 1));
         const shouldUpdate = !GameController.game_paused() && TimeKeeper.frames % framestep === 0;
+        const shouldUpdateWind = !GameController.game_paused() && TimeKeeper.frames % 50 === 1;
         const fading = elapsed < 1;
 
         // TODO: to be implemented in the future as, potentially, another method `.deepRenderUpdate()` or something
@@ -459,6 +461,11 @@ export default class CanvasController {
         //     this is where we update static drawings like terrain, airspace, video map, etc
         //     updates that happen here should be infrequent because they are considered expensive
         // }
+
+        // We want to update the wind before it's drawn.
+        if (shouldUpdateWind) {
+            airport.updateCurrentWind(airport.getWind());
+        }
 
         if (this._shouldShallowRender || shouldUpdate || fading) {
             const cc = this.canvas_get(CANVAS_NAME.STATIC);
--- a/src/assets/scripts/client/math/core.js
+++ b/src/assets/scripts/client/math/core.js
@@ -236,4 +236,18 @@ export const generateRandomOctalWithLength = (length = 1) => {
     }
 
     return leftPad(value.join(''), length);
+};
+
+/**
+ * Generates a random number, on an (approximate) bell curve.
+ * AKA - Normal Distribution, Gaussian Distribution
+ * 
+ * @function getNormalDistributedNumber
+ * @param average
+ * @return {number}
+ */
+export const getNormalDistributedNumber = (average) => {
+    const initialNumber = ((Math.random() + Math.random() + Math.random() + Math.random() + Math.random() + Math.random()) % 6);
+
+    return initialNumber * average * 2;
 }
--- a/src/assets/scripts/client/weather/WindController.js
+++ b/src/assets/scripts/client/weather/WindController.js
@@ -1,0 +1,63 @@
+import AirportModel from '../airport/AirportModel';
+import AirportController from '../airport/AirportController';
+import { getNormalDistributedNumber } from '../math/core';
+
+/**
+ * @class WindController
+ */
+class WindController {
+    /**
+     * Constructor. Generates the initial wind for
+     * an airport. Should be called on initialization, or
+     * when an airport is changed.
+     * 
+     * @method constructor
+     */
+    constructor() {
+        /**
+         * Our {AirportModel} to use.
+         */
+        this.airport = AirportController.current;
+
+        /**
+         * Initial wind speed.
+         */
+        this._defaultWindSpeed = airport.wind.speed;
+
+        /**
+         * Initial wind angle
+         */
+        this._defaultWindAngle = airport.wind.angle;
+
+        this.updateWindSpeed(_defaultWindSpeed);
+        this.updateWindAngle(_defaultWindAngle);
+    }
+
+    /**
+     * Gets a new speed for the wind.
+     * 
+     * @param averageSpeed {Number} - the average windspeed [knots]
+     * @return {Number} - the new windspeed
+     */
+    updateWindSpeed(averageSpeed) {
+        return getNormalDistributedNumber(averageSpeed);
+    }
+
+    /**
+     * Gets a new wind angle for the wind.
+     * 
+     * @param averageAngle {Number} - the average wind angle [degrees]
+     * @return {Number} - the new wind angle, from 0 to 359
+     */
+    updateWindAngle(averageAngle) {
+        const newAngle = getNormalDistributedNumber(averageAngle);
+
+        if (newAngle >= 360) {
+            return newAngle - 360;
+        }
+
+        return newAngle;
+    }
+}
+
+export default new WindController();


n8rzz:
work on this will very much be blocked by #754.  There are several changes to the `CanvasController` and other classes that may affect how this is implemented.  

I will add notes to your current implementation, but plan for some refactor/conflicts once #754 is merged.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -14,6 +14,7 @@ import RunwayCollection from './runway/RunwayCollection';
 import StaticPositionModel from '../base/StaticPositionModel';
 import { isValidGpsCoordinatePair } from '../base/positionModelHelpers';
 import { degreesToRadians, parseElevation } from '../utilities/unitConverters';
+import { getNewWind } from '../utilities/windUtilities';
```
n8rzz: avoid using the term `get` with methods or functions.  you will see it used sparingly, but should be used only to denote a getter type method.

in this case, the word `calculate` would be more appropriate.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -553,19 +556,6 @@ export default class AirportModel {
      */
     getWind = () => {
```
n8rzz: see note below on where this is used, but this could probably be removed entirely

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -642,26 +632,16 @@ export default class AirportModel {
...
      * @for AirportModel
      * @method updateRunway
      */
     updateRunway() {
-        // const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(this.getWind);
-        //
-        // this.setArrivalRunway(bestRunwayForWind);
-        // this.setDepartureRunway(bestRunwayForWind);
-    }
+        const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(this.getWind);
```
n8rzz: you could either call the function as a parameter here or set it to a const and pass it along:
```javascript
// option one
const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(getWind());

// option two (preferred)
const currentWind = getWind();
const bestRunwayForWind = this._runwayCollection.findBestRunwayForWind(currentWind);
```

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -642,26 +632,16 @@ export default class AirportModel {
         return this._runwayCollection.getRunwayRelationshipForRunwayNames(primaryRunwayName, comparatorRunwayName);
     }
 
-    // TODO: Implement changing winds, then bring this method back to life
     /**
      * @for AirportModel
      * @method updateRunway
      */
     updateRunway() {
```
n8rzz: out of curiosity, where is this method called?  is it part of the update loop?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -448,10 +448,12 @@ export default class CanvasController {
      * @method canvas_update_post
      */
     canvas_update_post() {
-        const elapsed = GameController.game_time() - AirportController.airport_get().start;
+        const airport = AirportController.airport_get();
+        const elapsed = GameController.game_time() - airport.start;
         const alpha = extrapolate_range_clamp(0.1, elapsed, 0.4, 0, 1);
         const framestep = Math.round(extrapolate_range_clamp(1, GameController.game.speedup, 10, 30, 1));
         const shouldUpdate = !GameController.game_paused() && TimeKeeper.frames % framestep === 0;
+        const shouldUpdateWind = !GameController.game_paused() && TimeKeeper.frames % 50 === 1;
```
n8rzz: this method is concerned with drawing the current state of all the drawable game objects.  currently, this method is gross.  like, really gross.  we are doing so much here that we shouldn't be doing here.

the updating of the wind should happen elsewhere completely outside of this class so that once it comes time to draw it we are just grabbing values that already exist.  Any math or calculations that happen slow down the drawing process and lead to dropped frames (currently a big problem with this method)

src/assets/scripts/client/math/core.js
```diff
@@ -236,4 +236,18 @@ export const generateRandomOctalWithLength = (length = 1) => {
...
+
+/**
+ * Generates a random number, on an (approximate) bell curve.
+ * AKA - Normal Distribution, Gaussian Distribution
+ * 
+ * @function getNormalDistributedNumber
+ * @param average
+ * @return {number}
+ */
+export const getNormalDistributedNumber = (average) => {
```
n8rzz: try to avoid the word `get` here.  again, like I mentioned above, the word `calculate` would be more appropriate.

this should also be a normal function instead of an arrow function: 
```javascript
export function getNormalDistributedNumber(average) {
 // ...
}
```

src/assets/scripts/client/math/core.js
```diff
@@ -236,4 +236,18 @@ export const generateRandomOctalWithLength = (length = 1) => {
...
 
     return leftPad(value.join(''), length);
+};
+
+/**
+ * Generates a random number, on an (approximate) bell curve.
+ * AKA - Normal Distribution, Gaussian Distribution
+ * 
+ * @function getNormalDistributedNumber
+ * @param average
```
n8rzz: be sure to include the data type after the `param`

src/assets/scripts/client/math/core.js
```diff
@@ -236,4 +236,18 @@ export const generateRandomOctalWithLength = (length = 1) => {
...
+ * @return {number}
+ */
+export const getNormalDistributedNumber = (average) => {
+    const initialNumber = ((Math.random() + Math.random() + Math.random() + Math.random() + Math.random() + Math.random()) % 6);
```
n8rzz: why the repeated `Math.random()`s and what is the significance of the `% 6` at the end?  this would be a good spot for an inline comment for those less fluent in math (like myself)

src/assets/scripts/client/utilities/windUtilities.js
```diff
@@ -0,0 +1,18 @@
+import { getNormalDistributedNumber } from '../math/core';
+
+export const getNewWind = (data) => {
```
n8rzz: another word to avoid, `new`.  that is a reserved word in JS (when used by itself).  a better choice here would be `next`.  so: `calculateNextWind` is, I think, a better name.

src/assets/scripts/client/utilities/windUtilities.js
```diff
@@ -0,0 +1,18 @@
...
+    const speed = getNormalDistributedNumber(data.speed);
+    const initialAngle = getNormalDistributedNumber(data.angle);
+    let newAngle = initialAngle;
```
n8rzz: there is that `new` word again.  try `nextAngle`

src/assets/scripts/client/utilities/windUtilities.js
```diff
@@ -0,0 +1,18 @@
...
+
+    if(initialAngle > 360) {
+        newAngle = initialAngle - 360;
```
n8rzz: what happens if this receives `750`?  is that even possible?  If so, what happens?

test/utilities/windUtilities.spec.js
```diff
@@ -0,0 +1,14 @@
+import ava from 'ava';
+
+import { getNewWind } from '../../src/assets/scripts/client/utilities/windUtilities';
+
+const windMock = {
+    speed: 10,
+    angle: 400
+};
+
+ava('.getNewWind() does not return a value greater than 360', (t) => {
```
n8rzz: you should add a test case for the other logical branch branch; when this function is passed values greater than 360

src/assets/scripts/client/airport/AirportModel.js
sam-irl: I don't think it's actually called anywhere

n8rzz: actually, I found it [here](https://github.com/openscope/openscope/pull/762/files#diff-db21fe1f5294d56680983f0c7dd43c24R467)

src/assets/scripts/client/utilities/windUtilities.js
erikquinn: we have a helper function for this actually, `degrees_normalize()` I think it's called.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -376,7 +377,7 @@ export default class AirportModel {
         this.setActiveRunwaysFromNames(data.arrivalRunway, data.departureRunway);
         this.buildAirportMaps(data.maps);
         this.buildRestrictedAreas(data.restricted);
-        this.updateCurrentWind(data.wind);
+        AirportWind.setStaticWind(data.wind);
```
n8rzz: I should have warned you that doing it this way, though easier, adds a little more complexity to getting things connected.  Since we are now using a class for `AirportWind`, you will need to use the `this.wind` property to hold this instance:
```javascript
this.wind = new AirportWindModel.(data.wind);
```

_(sorry, I'm changing my mind on the name again.  I think `AirportWindModel` is better)_

src/assets/scripts/client/airport/AirportWind.js
```diff
@@ -0,0 +1,35 @@
+import AirportModel from './AirportModel';
```
n8rzz: In javascript, importing the `AirportModel` here creates a circular reference, which is bad.  Good news is you don't need this import and it can be removed 🎉

src/assets/scripts/client/airport/AirportWind.js
```diff
@@ -0,0 +1,35 @@
...
+import { calculateNextWind } from '../utilities/windUtilities';
+
+/**
+ * Mutable array that holds the information on the airport's current wind.
+ */
+export let airportWind = {
+   speed: 10,
```
n8rzz: these two properties can be made class properties of the `AirportWindModel`

src/assets/scripts/client/airport/AirportWind.js
```diff
@@ -0,0 +1,35 @@
...
+   angle: 0
+};
+
+export default class AirportWind {
+    /**
+     * Constructor
+     */
+    constructor() {}
```
n8rzz: so then this line becomes:
```javascript
constructor() {
  this.angle = -9999;
  this.speed = -1;

  return this._init();
}
```

notice we initialize the properties with invalid values.  within the `_init()` method you would then set the default values for those properties.

src/assets/scripts/client/math/core.js
erikquinn: interesting way to do that. This would produce a proper gaussian random distribution (aka the classic "bell curve"). And then the assumption is that you would measure the difference from `3` at the end, and use that to shift the wind up and down. The resulting behavior would be that it would usually have very slight increases or decreases, with dramatic changes possible, but far less likely.

The trouble with trying to do just `Math.random()` is that this function produces an even distribution, meaning that `0, 0.1, 0.5, 0.77, 0.89, 1` are all just as likely to be selected as each other. When trying to emulate wind changes, this would just be too jittery.

This does seem heavy for what it does, but it actually seems to be one of the best performing ways to get it done. We could reduce it from 6 to like 3 or 4, losing some integrity, but I'm not sure the time spent on this is really much of a concern anyway. Would probably be best to keep it accurate, and just call it every couple of seconds or something-- definitely not every frame.

erikquinn: If you're interested: 
https://jsperf.com/normal-random-number-generators/8

src/assets/scripts/client/airport/AirportWind.js
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Sets the initial airportWind, instead of the default.
+     * Also can be used to force-set the wind.
+     * 
+     * @method setStaticWind
+     * @param nextWind {Array} [speed, angle]
+     */
+    setStaticWind(nextWind) {
+        airportWind = nextWind;
```
n8rzz: now you can use the class properties instead of updating the object.

that is an object, by the way, not an array.  you can also use `const` with objects even if the internal properties change.  as long as the reference to the object does not change, `const` is still good.

src/assets/scripts/client/airport/AirportWind.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    /**
+     * Updates the wind, using a bell curve.
+     */
+    updateWind(wind) {
+        return calculateNextWind(wind);
```
n8rzz: because we now have a class that holds all of our wind logic and values, we can probably bring this `calculateNextWind` function in and use it as a class method.

src/assets/scripts/client/airport/AirportWind.js
erikquinn: @n8rzz Would it be prudent to set up some sort of broad-scope weather class? Like an `AtmosphereModel` or something? That could have wind/pressure/etc properties and methods all in a single place. just a thought.

n8rzz: Maybe someday. It I feel that would be an over abstraction at the moment.

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+    _init(data) {
+        this.speed = data.speed;
+        this.angle = data.angle;
+
+        const nextWind = {
+            speed: this.speed,
+            angle: this.angle
+        };
+
+        return this._calculateNextWind(nextWind);
```
n8rzz: the `nextWind` arg here is not needed.  you could simply call the method.  This would overwrite any default values, though.  Is that intended?

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Actual math function to find the wind on a bell curve.
+     * 
+     * @method calculateNextWind
+     * @param {Object} data 
+     * @private
+     */
+    _calculateNextWind(data) {
+        const speed = calculateNormalDistributedNumber(data.speed);
```
n8rzz: you can use `this.speed` here

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Actual math function to find the wind on a bell curve.
+     * 
+     * @method calculateNextWind
+     * @param {Object} data 
+     * @private
+     */
+    _calculateNextWind(data) {
```
n8rzz: the `data` param can be removed

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+        const speed = calculateNormalDistributedNumber(data.speed);
+        const initialAngle = calculateNormalDistributedNumber(data.angle);
```
n8rzz: you can use `this.angle` here

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+        let nextAngle = initialAngle;
+    
+        if (initialAngle > 360) {
+            // How many times can we subtract 360 and not get a negative number?
+            const factorsOfThreeSixty = _floor(initialAngle / 360);
+    
+            nextAngle = initialAngle - (360 * factorsOfThreeSixty);
+        }
+    
+        const nextWind = {
```
n8rzz: here you will want to update the class properties:
```javascript
this.angle = nextAngle;
this.speed = speed;
```

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,95 @@
...
+        this.speed = 10;
+        this.angle = 0;
+    }
+
+    /**
+     * Getter for the current wind.
+     * 
+     * @return {Object} wind, with properties `speed` and `angle`.
+     */
+    get wind() {
```
n8rzz: this is likely not needed.  Since the `AirportModel` has access to an instance of this class, these values would be referenced thusly:
```javascript
// AirportModel.js
this.wind = new AirportWindModel();

// current wind angle
this.wind.angle

// current wind speed
this.wind.speed
```

src/assets/scripts/client/airport/AirportWindModel.js
sam-irl: So, what I'm trying to get my head around is airport-specific wind defaults: so, just pass the data to `_init`, and overwrite the defaults. If it isn't passed a `speed` or `angle` KV-pair, it should just leave it as default.

sam-irl:
I'm thinking a wind update every 5 minutes is a good number right now

src/assets/scripts/client/airport/AirportWindModel.js
n8rzz: ahh, ok.

First question to ask there is wind required to be defined in the airport file?  You'd have to check the docs for that answer.  If it is, then there may not be a need to call `_calculateNextWind()` here.  you can simply set the data and be done.  You may need to translate the data, though.  hmm 🤔

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Creates the timer to update the wind.
+     * 
+     * @method createWindUpdateTimer
+     * @private
+     * @return {gameTimeout} an instance of the new game timeout.
+     */
+    _createWindUpdateTimer() {
+        return GameController.game_timeout(
```
n8rzz: `game_interval()` may actually be more appropriate here since we want this to run continuously.  Unless we're changing the delay time after each run?

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,109 @@
...
+            _doUpdateTimer(),
```
n8rzz: `_doUpdateTimer()` will be undefined as written here.  since this is a class method, be sure to include `this`: `this._doUpdateTimer()`.

let me know if you run into scope issues with that.  you may run into a situation where you'll see an `this method is undefined` error (after adding `this`).  but we can cross that bridge if we come to it.

n8rzz:
@sam-irl I made a few updates for you to get your new class working.  I saw your last commit message and wanted to nudge you along a little bit.  Don't loose faith, you are doing a good job getting this going and are actually really close to getting this implemented.

With my changes, you should now have an `AirportWindModel` class that instantiates with the correct properties and also sets up an interval that calls `_calculateNextWind()` to update values.  

Next task is to take this class and use it throughout the `AirportModel`.  Which should hopefully be pretty straightforward.  

If you'd like, I can whip up the tests for the new `AirportWindModel` class?  Just let me know!

n8rzz:
@sam-irl I went ahead and added some tests for this. I also resolved the conflicts that resulted from #754 getting merged.  I also made a few other small changes to names and doc blocks and such.

When you start back up on this one, be sure to pull first 😉

sam-irl:
@n8rzz unblocked! great!

n8rzz:
updated this branch with develop and resolved conflicts from #799.  Also added `.calculateWindVector()` method to `AirportWindModel` and implemented it within `AircraftModel` classes.

Be sure to pull before continuing work!

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,150 @@
...
+         * @property angle
+         * @type {number}
+         * @default -9999
+         */
+        this.angle = -9999;
+
+        /**
+         *
+         *
+         * @property speed
```
erikquinn: Complete docblocks

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,150 @@
...
+export default class AirportWindModel {
+    /**
+     * @constructor
+     * @param initialAirportWind {object{ speed: number, angle: number }}
+     */
+    constructor(initialAirportWind) {
+        /**
+         *
+         *
+         * @property angle
```
erikquinn: Complete docblocks

felixscheffer:
@sam-irl Do you still plan to work on this one? Would be much appreciated.

sam-irl:
@n8rzz I've finally figured out what was frustrating me so much with this PR - it turns out `GameController.game_interval` had a bug that made the update loop call continuously after it fires initially.

Things to do:
[ ] add a handler to `AirportController` to recalculate the best runway
[ ] finish documentation

if you can see anything i've missed please let me know

src/assets/scripts/client/airport/AirportWindModel.js
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @for AirportWindModel
+     * @method createWindUpdateTimer
+     * @return {gameTimeout} an instance of the new game timeout.
+     * @private
+     */
+    _createWindUpdateTimer() {
+        GameController.game_interval(
+            this._calculateNextWindHandler,
+            300,
```
n8rzz: This value should be in milliseconds.  so the 'fires constantly' thing you mentioned is because this value is too small.  As written, this is 3/10s of a second.  I think you may have meant to use `3000` instead, which is 3 seconds.  I'd actually suggest a larger value here, something more like once every 5 minutes or something?

sam-irl: Actually, that's after it's been fixed - the wind updates after 5 minutes of in-game time. It appears that the `game_interval` accepts seconds.

sam-irl:
Should the angle be rounded to the nearest 5 or 10? It seems more natural that way

src/assets/scripts/client/airport/AirportWindModel.js
felixscheffer: I double checked and @sam-irl is correct. `game_interval` is using seconds, not milliseconds.

felixscheffer:
@sam-irl `game_interval` should be fixed in the `develop` branch now.

JPmAn24:
While testing this, I found that the angle is never an integer after the first update. Is it okay if I force it to be an integer, or do I just have the wind display show an integer?

erikquinn:
The wind display should always show a rounded value. The value of the wind should be allowed to be a floating type, and whenever we want to show (or say) what the wind is, we would round the actual value to an integer (or more likely to the nearest ten degrees, eg `_round(windDirection, -1)`.

JPmAn24:
I have got progress on this. Tested with KAUS and KSEA, the wind shift does work, and I can have aircraft take-off in the opposite direction than the assigned runway.

PR #763, opened 2017-10-04
documentation/764

TheBear44:

Fix the dead link in airport guide documentation.


sam-irl:
Could we also make this a bulleted list? I know that this isn't the case now, but it would look a lot better.
```
- EIDW
- MDSD
- EGLL
...etc
```

TheBear44:
I add a bulleted list and airport name. It should look better now.

erikquinn:
Hi @TheBear44! And thanks for the pull request, that's definitely much better. I've taken the liberty of adding you to the organization, so you will be able to create branches owned by the `openscope` organization, because we try not to use forks (as you can see, our tests are not able to run on PRs originating from forks). So check your email for that, and feel free to [join us on slack](http;//slack.openscope.co), where somebody can be reached basically 24/7, and where we discuss bugs, features, and help each other along.

This PR will be merged into a _local_ branch, where we will continue the review process. Thanks again, and welcome!

EQ

PR #770, opened 2017-10-07
bugfix/567

### LINKED ISSUE #567: MCP altitude set to 0 when issuing `dvs`
Arriving aircraft given the command to "DVS" get an assigned altitude of 0 in the flight strip. See BCY858.

>@Fechulo : 
>Arriving aircraft descend via the STAR by default, so there's no need to give them the 'dvs' command. >However, the assigned altitude shouldn't be 0, so that's something to look into.

screenshot 5screenshot 8 _li

---

Steps to rectify: 
1. The aircraft are reaching 0 feet, but are not considered to have crashed. Resolve whatever is causing this, and get them to crash.
1. Flight strip shows NaN upon the aircraft reaching 0 feet altitude. It should at least be showing `0` instead of `NaN`.
1. Despite the fact that the aircraft have had their autopilots set to 0 feet, this should never be allowed except during landing. The catch-all keeping them at least 1000ft above the airport elevation should still be in effect here.
1. The `dvs` command should not be setting the MCP altitude to 0, but rather the bottom altitude of the procedure (if there is one). If there is not one, we may want to consider setting the MCP altitude to the airspace ceiling if it is higher (to force a descent so we can control the aircraft on STARs that don't have built in crossing restrictions).

erikquinn:
Resolves #567.

Fix bug of aircraft descending via STAR to '0' altitude.

### DIFF
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -495,7 +495,7 @@ export default class Pilot {
             nextAltitude = this._fms.getBottomAltitude();
         }
 
-        if (_isNil(nextAltitude) || nextAltitude === Infinity) {
+        if (isNaN(nextAltitude) || nextAltitude === Infinity) {
             return [false, 'unable to descend via STAR'];
         }
 
@@ -503,12 +503,7 @@ export default class Pilot {
         this._mcp.setAltitudeVnav();
         this._mcp.setSpeedVnav();
 
-        // Build readback
-        const readback = {};
-        readback.log = 'descend via the arrival';
-        readback.say = 'descend via the arrival';
-
-        return [true, readback];
+        return [true, 'descend via STAR'];
     }
 
     /**
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -418,10 +418,8 @@ export default class Pilot {
         }
 
         const airport = AirportController.airport_get();
-        const altitudeToMaintain = Math.max(
-            Math.min(aircraftModel.altitude, this._mcp.altitude),
-            airport.minAssignableAltitude
-        );
+        const descentAltitude = Math.min(aircraftModel.altitude, this._mcp.altitude);
+        const altitudeToMaintain = Math.max(descentAltitude, airport.minAssignableAltitude);
 
         this._mcp.setAltitudeFieldValue(altitudeToMaintain);
         this._mcp.setAltitudeHold();
@@ -487,12 +485,21 @@ export default class Pilot {
      *
      * @for Pilot
      * @method descendViaStar
-     * @param altitude {number}  (optional) altitude at which the descent will end (regardless of fix restrictions)
-     *                                      this should be the altitude of the lowest fix restriction on the STAR
-     * @return {array}           [success of operation, readback]
+     * @param bottomAltitude {number} (optional) altitude at which the descent will end (regardless of fix restrictions)
+     * @return {array}                [success of operation, readback]
      */
-    descendViaStar(altitude = 0) {
-        this._mcp.setAltitudeFieldValue(altitude);
+    descendViaStar(bottomAltitude) {
+        let nextAltitude = bottomAltitude;
+
+        if (typeof nextAltitude === 'undefined') {
+            nextAltitude = this._fms.getBottomAltitude();
+        }
+
+        if (_isNil(nextAltitude) || nextAltitude === Infinity) {
+            return [false, 'unable to descend via STAR'];
+        }
+
+        this._mcp.setAltitudeFieldValue(nextAltitude);
         this._mcp.setAltitudeVnav();
         this._mcp.setSpeedVnav();
 
@@ -505,7 +512,7 @@ export default class Pilot {
     }
 
     /**
-     * Abort the landing attempt; maintain present heading/speed, and climb to a reasonable alttiude
+     * Abort the landing attempt; maintain present heading/speed, and climb to a reasonable altitude
      *
      * @for Pilot
      * @method goAround
--- a/test/aircraft/Pilot/descendViaStar.spec.js
+++ b/test/aircraft/Pilot/descendViaStar.spec.js
@@ -9,8 +9,8 @@ import {
 let pilot = null;
 
 // Mocks
-const successResponse = [true, 'descend via STAR'];
-const failureResponse = [false, 'unable to descend via STAR'];
+const successResponseMock = [true, 'descend via STAR'];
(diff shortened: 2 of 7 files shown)


src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -488,12 +488,19 @@ export default class Pilot {
...
-     * @param altitude {number}  (optional) altitude at which the descent will end (regardless of fix restrictions)
-     *                                      this should be the altitude of the lowest fix restriction on the STAR
-     * @return {array}           [success of operation, readback]
+     * @param bottomAltitude {number} (optional) altitude at which the descent will end (regardless of fix restrictions)
+     * @return {array}                [success of operation, readback]
      */
-    descendViaStar(altitude = 0) {
-        this._mcp.setAltitudeFieldValue(altitude);
+    descendViaStar(bottomAltitude) {
+        if (bottomAltitude === undefined) {
```
n8rzz: This should be `typeof bottomAltitude === 'undefined'`

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -488,12 +488,19 @@ export default class Pilot {
...
+            bottomAltitude = this._fms.getBottomAltitude();
+        }
+
+        if (_isNil(bottomAltitude) || bottomAltitude === Infinity) {
```
n8rzz: Isn't the `_isNil()` duplicative here?

erikquinn: that was an attempt to be defensive in case `.getBottomAltitude()` returned a non-numeric value, which it can in many cases.

src/assets/scripts/client/aircraft/Pilot/Pilot.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -488,12 +488,19 @@ export default class Pilot {
      *
      * @for Pilot
      * @method descendViaStar
-     * @param altitude {number}  (optional) altitude at which the descent will end (regardless of fix restrictions)
-     *                                      this should be the altitude of the lowest fix restriction on the STAR
-     * @return {array}           [success of operation, readback]
+     * @param bottomAltitude {number} (optional) altitude at which the descent will end (regardless of fix restrictions)
```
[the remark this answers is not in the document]
erikquinn: correct

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -488,12 +488,19 @@ export default class Pilot {
...
+     * @return {array}                [success of operation, readback]
      */
-    descendViaStar(altitude = 0) {
-        this._mcp.setAltitudeFieldValue(altitude);
+    descendViaStar(bottomAltitude) {
+        if (typeof bottomAltitude === 'undefined') {
+            bottomAltitude = this._fms.getBottomAltitude();
+        }
+
+        if (_isNil(bottomAltitude) || bottomAltitude === Infinity) {
```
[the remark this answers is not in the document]
erikquinn: very well then, all done

test/aircraft/Pilot/descendViaStar.spec.js
```diff
@@ -14,7 +14,7 @@ ava('.descendViaStar() sets the correct Mcp modes and values', (t) => {
 
     t.true(pilot._mcp.altitudeMode === 'VNAV');
     t.true(pilot._mcp.speedMode === 'VNAV');
-    t.true(pilot._mcp.altitude === 0);
+    t.true(pilot._mcp.altitude === 8000);
```
n8rzz: I'd really like to see some additional test cases for the branching logic in this method:
```javascript
if (typeof nextAltitude === 'undefined') {
    nextAltitude = this._fms.getBottomAltitude();
}

if (_isNil(nextAltitude) || nextAltitude === Infinity) {
     return [false, 'unable to descend via STAR'];
}
```

I count 6 additional test cases here:
#### first `if`
- when `nextAltitude` is defined
- when `nextAltitude` is `undefined`

#### second `if`
- when `nextAltitude` `isNil`
- when `nextAltitude` is not `isNil`
- when `nextAltitude` is `Infinity`
- when `nextAltitude` is  not `Infinity`

erikquinn: done! 😃

test/aircraft/Pilot/descendViaStar.spec.js
```diff
@@ -1,43 +1,75 @@
...
 } from '../../fixtures/aircraftFixtures';
 
+// Fixtures
+let pilot = null;
 
-ava('.descendViaStar() sets the correct Mcp modes and values', (t) => {
-    const pilot = new Pilot(modeControllerFixture, fmsArrivalFixture);
+// Mocks
+const successResponse = [true, 'descend via STAR'];
+const failureResponse = [false, 'unable to descend via STAR'];
```
n8rzz: these two vars should include `mock` in their name

test/fixtures/aircraftFixtures.js
```diff
@@ -13,3 +13,15 @@ const runwayAssignmentMock = airportModelFixture.getRunway('19L');
...
+
+export function createFmsArrivalFixture() {
+    return new Fms(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK, runwayAssignmentMock, AIRCRAFT_DEFINITION_MOCK, navigationLibraryFixture);
+}
+
+export function createFmsDepartureFixture() {
+    return new Fms(DEPARTURE_AIRCRAFT_INIT_PROPS_MOCK, runwayAssignmentMock, AIRCRAFT_DEFINITION_MOCK, navigationLibraryFixture);
+}
+
+export function createModeControllerFixture() {
```
n8rzz: why are you using functions here instead of just exporting an instance?
```javascript
export const modeControllerFixture = new ModeController();
```

erikquinn: when i simpy export and import an instance, that instance can be mutated throughout subsequent tests in the same test file. I need the instance to be generated fresh each time. I could do that in the test file, but that would require a myriad of extra imports. Instead I just added a create-a-new-instance helper function that I can import and run to achieve the desired result.

test/aircraft/Pilot/descendViaStar.spec.js
erikquinn: done

test/fixtures/aircraftFixtures.js
n8rzz: ok, that makes sense.  

_(there should probably be a comment explaining the why somewhere in that file 😉 )_

erikquinn: Fair enough, done!

PR #772, opened 2017-10-08
bugfix/667

### LINKED ISSUE #667: Aircraft cleared twice for ILS approach do not capture glideslope
Browser: Chrome
Airport: All

#### Issue Description
Aircraft that are given a turn and cleared for an ILS approach _twice_ before actually intercepting the localizer are not descending along the glideslope. They remain at the last assigned altitude until crossing the runway threshold, then complete the descent.

#### Steps to reproduce
1. Position an aircraft where they can be assigned a new heading to intercept the localizer, at an altitude below the glideslope altitude where they are to intercept.
1. In one command, issue the aircraft a turn to the intercept heading, and clear them for the approach, eg `T L 160 ILS 16R`.
1. Verify the projection shows the aircraft joining the localizer.
1. Before they have joined the localizer, enter the same command for the same aircraft (eg `T L 160 ILS 16R`).
1. Observe aircraft join the localizer, but never descend along the glideslope.

erikquinn:
Resolves #667.

Fix bug where aircraft cleared twice for ILS won't join the glideslope.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -207,10 +207,9 @@ export default class AircraftCommander {
      *
      * @for AircraftCommander
      * @method runAbort
-     * @param {AircraftModel} aircraft
      * @return {array} [success of operation, readback]
      */
-    runAbort(aircraft) {
+    runAbort() {
         return [false, "the 'abort' command has been deprecated, please see documentation for help"];
     }
 
@@ -251,7 +250,7 @@ export default class AircraftCommander {
         let direction = data[0];
         const heading = data[1];
         const incremental = data[2];
-        const readback = aircraft.pilot.maintainHeading(aircraft.heading, heading, direction, incremental);
+        const readback = aircraft.pilot.maintainHeading(aircraft, heading, direction, incremental);
 
         if (direction === null) {
             direction = '';
@@ -260,7 +259,7 @@ export default class AircraftCommander {
         aircraft.target.turn = direction;
 
         if (aircraft.hasApproachClearance) {
-            aircraft.cancelApproachClearance(aircraft.altitude, aircraft.heading);
+            aircraft.cancelApproachClearance(aircraft);
         }
 
         return readback;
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -260,7 +259,7 @@ export default class AircraftCommander {
         aircraft.target.turn = direction;
 
         if (aircraft.hasApproachClearance) {
-            aircraft.cancelApproachClearance(aircraft.altitude, aircraft.heading);
+            aircraft.cancelApproachClearance(aircraft);
```
n8rzz: You could probably drop the aircraft param here too and just use`this` when in `.cancelApproachClearance()`

n8rzz: Do we need the aircraft param here?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -409,24 +409,23 @@ export default class Pilot {
...
+     * @return {array} [success of operation, readback]
      */
-    cancelApproachClearance(currentAltitude, currentHeading) {
+    cancelApproachClearance(aircraftModel) {
         if (!this.hasApproachClearance) {
             return [false, 'we have no approach clearance to cancel!'];
         }
 
         const airport = AirportController.airport_get();
         const altitudeToMaintain = Math.max(
```
n8rzz: Could you break this up a bit more?  I'm not a fan of the nested `Math` calls here

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: Thanks for pointing this out, because it doesn't look right for an entirely different reason: `.cancelApproachClearance()` is not a property of the aircraft at all, but rather the pilot. Fixed that mistake right up!

It is for that reason that we actually do have to pass the aircraft (or at least a couple of its properties) to the pilot, since he does not have access to it for reasons of circularity (possibly a word?).

That's actually a pattern I think would be wise with the pilot rather than passing hundreds of individual properties to the pilot when he completes an action: let him take a look at the entire aircraft, and then manipulate the mcp and fms appropriately. I suppose we could even remove his local references if desired. That's all beyond the scope of this anyway, but food for thought?

erikquinn: see above

src/assets/scripts/client/aircraft/Pilot/Pilot.js
erikquinn: done!

PR #779, opened 2017-10-13
bugfix/27

### LINKED ISSUE #27: Using runways not valid for assigned procedure causes crash
When either of the following are true:
 - taxiing a departure to a runway, and then assigning them a SID that is not valid
 - an aircraft is expecting to land on runway A, and is assigned a STAR only valid for runway B

...the game will crash.

Originally reported in zlsa/atc#573, zlsa/atc#519, zlsa/atc#580.

erikquinn:
Resolves #27.

Improve error handling when procedure is incompatible with runway assignment.

---

Blocked by #809.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -510,15 +510,8 @@ export default class AircraftCommander {
       * @return {array}   [success of operation, readback]
       */
     runReroute(aircraft, data) {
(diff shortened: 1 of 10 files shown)


erikquinn:
To difficult to come up with a good solution. Will be postponing this until resolution of #809.

erikquinn:
Too many conflicts, too many changes since work on this began. Didn't get too far with it anyway, so will just discard and try another time.

PR #788, opened 2017-10-17
feature/24

### LINKED ISSUE #24: Command to inform arrivals of the arrival runway to expect
Need a command we can use to inform an aircraft of what runway they can expect for arrival. This would be used primarily to help them determine which branches of a given STAR they should be following, as there are often different paths to follow in the chart based on the runway they are to land on.

See zlsa/atc#500.

erikquinn:
Resolves #24.

Add command to inform arrivals of assigned landing runway.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -609,7 +609,7 @@ export default class InputController {
             throw error;
         }
 
-        if (aircraftCommandParser.command !== 'transmit') {
+        if (aircraftCommandParser.command !== PARSED_COMMAND_NAME.TRANSMIT) {
             return this.processSystemCommand(aircraftCommandParser);
         }
 
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -406,7 +406,7 @@ export default class InputController {
 
                 break;
             case KEY_CODES.MULTIPLY:
-                this.$commandInput.val(`${currentCommandInputValue} \u2b50 `);
+                this.$commandInput.val(`${currentCommandInputValue} * `);
                 event.preventDefault();
                 this.onCommandInputChangeHandler();
 
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -406,7 +406,7 @@ export default class InputController {
 
                 break;
             case KEY_CODES.MULTIPLY:
-                this.$commandInput.val(`${currentCommandInputValue} \u2B50 `);
+                this.$commandInput.val(`${currentCommandInputValue} \u2b50 `);
                 event.preventDefault();
                 this.onCommandInputChangeHandler();
 
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -1,4 +1,5 @@
 import _has from 'lodash/has';
+import _isNil from 'lodash/isNil';
 import _map from 'lodash/map';
 import _round from 'lodash/round';
 import AirportController from '../airport/AirportController';
@@ -227,7 +228,7 @@ export default class AircraftCommander {
      * @return {array} [success of operation, readback]
      */
     runClearedAsFiled(aircraft) {
-        return aircraft.pilot.clearedAsFiled();
+        return aircraft.pilot.clearedAsFiled(aircraft);
     }
 
     /**
@@ -326,7 +327,19 @@ export default class AircraftCommander {
      */
     runExpectArrivalRunway(aircraft, data) {
         const airportModel = AirportController.airport_get();
-        const runwayModel = airportModel.getRunway(data[0]);
+        const runwayName = data[0];
+        const runwayModel = airportModel.getRunway(runwayName);
+
+        if (_isNil(runwayModel)) {
+            const oldRunwayModel = aircraft.fms.arrivalRunwayModel;
+            const readback = {};
+            readback.log = `unable to find Runway ${runwayName} on our charts, ` +
+                `expecting Runway ${oldRunwayModel.name} instead`;
+            readback.say = `unable to find Runway ${radio_runway(runwayName)} on our ` +
+                `charts, expecting Runway ${oldRunwayModel.getRadioName()} instead`;
+
+            return [false, readback];
+        }
 
         return aircraft.pilot.setArrivalRunway(aircraft, runwayModel);
     }
@@ -562,19 +575,23 @@ export default class AircraftCommander {
             taxiDestination = airport.departureRunwayModel.name;
         }
 
-        const runway = AirportController.airport_get().getRunway(taxiDestination.toUpperCase());
+        const runwayModel = AirportController.airport_get().getRunway(taxiDestination.toUpperCase());
 
-        if (!runway) {
-            return [false, `no runway ${taxiDestination.toUpperCase()}`];
+        if (!runwayModel) {
+            return [false, `unable to find Runway ${taxiDestination.toUpperCase()} on our charts`];
         }
 
-        const readback = aircraft.pilot.taxiToRunway(runway, isDeparture, flightPhase);
+        if (!aircraft.fms.isRunwayModelValidForSid(runwayModel)) {
+            aircraft.pilot.cancelDepartureClearance(aircraft);
+        }
+
+        const readback = aircraft.pilot.taxiToRunway(runwayModel, isDeparture, flightPhase);
 
         // TODO: this may need to live in a method on the aircraft somewhere
-        aircraft.fms.departureRunwayModel = runway;
+        aircraft.fms.departureRunwayModel = runwayModel;
         aircraft.taxi_start = TimeKeeper.accumulatedDeltaTime;
 
-        runway.addAircraftToQueue(aircraft.id);
+        runwayModel.addAircraftToQueue(aircraft.id);
         aircraft.setFlightPhase(FLIGHT_PHASE.TAXI);
 
         GameController.game_timeout(
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -331,12 +331,12 @@ export default class AircraftCommander {
         const runwayModel = airportModel.getRunway(runwayName);
 
         if (_isNil(runwayModel)) {
-            const oldRunwayModel = aircraft.fms.arrivalRunwayModel;
+            const previousRunwayModel = aircraft.fms.arrivalRunwayModel;
             const readback = {};
             readback.log = `unable to find Runway ${runwayName} on our charts, ` +
-                `expecting Runway ${oldRunwayModel.name} instead`;
+                `expecting Runway ${previousRunwayModel.name} instead`;
             readback.say = `unable to find Runway ${radio_runway(runwayName)} on our ` +
-                `charts, expecting Runway ${oldRunwayModel.getRadioName()} instead`;
+                `charts, expecting Runway ${previousRunwayModel.getRadioName()} instead`;
 
             return [false, readback];
         }
@@ -586,11 +586,9 @@ export default class AircraftCommander {
         }
 
         const readback = aircraft.pilot.taxiToRunway(runwayModel, isDeparture, flightPhase);
-
-        // TODO: this may need to live in a method on the aircraft somewhere
-        aircraft.fms.departureRunwayModel = runwayModel;
         aircraft.taxi_start = TimeKeeper.accumulatedDeltaTime;
 
+        aircraft.fms.setDepartureRunway(runwayModel);
         runwayModel.addAircraftToQueue(aircraft.id);
         aircraft.setFlightPhase(FLIGHT_PHASE.TAXI);
 
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -8,14 +8,14 @@ import NavigationLibrary from '../navigationLibrary/NavigationLibrary';
 import TimeKeeper from '../engine/TimeKeeper';
 import UiController from '../UiController';
 import { MCP_MODE } from './ModeControl/modeControlConstants';
-import { speech_say } from '../speech';
 import {
     FLIGHT_PHASE,
     FLIGHT_CATEGORY
 } from '../constants/aircraftConstants';
 import { EVENT } from '../constants/eventNames';
-// import { PROCEDURE_TYPE } from '../constants/routeConstants';
 import { round } from '../math/core';
+import { AIRCRAFT_COMMAND_MAP } from '../parsers/aircraftCommandParser/aircraftCommandMap';
+import { speech_say } from '../speech';
 import {
     radio_runway,
     radio_spellOut,
@@ -23,48 +23,6 @@ import {
     radio_altitude
 } from '../utilities/radioUtilities';
 import { radiansToDegrees } from '../utilities/unitConverters';
-
-/**
- * Enum of commands and thier corresponding function.
- *
- * Used to build a call to the correct function when a UI command, or commands,
- * for an aircraft have been issued.
- *
- * @property COMMANDS
- * @type {Object}
- * @final
- */
-const COMMANDS = {
-    abort: 'runAbort',
-    altitude: 'runAltitude',
-    clearedAsFiled: 'runClearedAsFiled',
-    climbViaSID: 'runClimbViaSID',
-    debug: 'runDebug',
-    delete: 'runDelete',
-    descendViaStar: 'runDescendViaStar',
-    direct: 'runDirect',
-    fix: 'runFix',
-    flyPresentHeading: 'runFlyPresentHeading',
-    heading: 'runHeading',
-    hold: 'runHold',
-    land: 'runLanding',
-    moveDataBlock: 'runMoveDataBlock',
-    route: 'runRoute',
-    reroute: 'runReroute',
-    sayAltitude: 'runSayAltitude',
-    sayAssignedAltitude: 'runSayAssignedAltitude',
-    sayHeading: 'runSayHeading',
-    sayAssignedHeading: 'runSayAssignedHeading',
-    sayIndicatedAirspeed: 'runSayIndicatedAirspeed',
-    sayAssignedSpeed: 'runSayAssignedSpeed',
-    sayRoute: 'runSayRoute',
-    sid: 'runSID',
-    speed: 'runSpeed',
-    squawk: 'runSquawk',
-    star: 'runSTAR',
-    takeoff: 'runTakeoff',
-    taxi: 'runTaxi'
-};
 
 /**
  *
@@ -188,17 +146,13 @@ export default class AircraftCommander {
      * @return {function}
      */
     run(aircraft, command, data) {
-        let call_func;
+        const { functionName } = AIRCRAFT_COMMAND_MAP[command];
 
-        if (COMMANDS[command]) {
-            call_func = COMMANDS[command];
-        }
-
-        if (!call_func) {
+        if (typeof functionName === 'undefined') {
             return [false, 'say again?'];
         }
 
-        return this[call_func](aircraft, data);
+        return this[functionName](aircraft, data);
     }
 
     /**
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -316,6 +316,22 @@ export default class AircraftCommander {
     }
 
     /**
+     * Set the arrival runway to expect for approach and landing
+     *
+     * @for AircraftCommander
+     * @method runExpectArrivalRunway
+     * @param aircraft {AircraftModel}
+     * @param data {array}
+     * @return {array} [success of operation, response]
+     */
+    runExpectArrivalRunway(aircraft, data) {
+        const airportModel = AirportController.airport_get();
+        const runwayModel = airportModel.getRunway(data[0]);
+
+        return aircraft.pilot.setArrivalRunway(aircraft, runwayModel);
+    }
+
+    /**
      * @for AircraftCommander
      * @method runFlyPresentHeading
      * @param aircraft {AircraftModel}
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -299,7 +299,7 @@ export default class Fms {
     }
 
     /**
-     * Initialize #arrivalAirportModel
+     * Initialize `#arrivalAirportModel`
      *
      * @for Fms
      * @method _initializeArrivalAirport
@@ -315,7 +315,7 @@ export default class Fms {
     }
 
     /**
-     * Initialize #arrivalRunwayModel
+     * Initialize `#arrivalRunwayModel`
      *
      * @for Fms
      * @method _initializeArrivalRunway
@@ -338,7 +338,7 @@ export default class Fms {
     }
 
     /**
-     * Initialize #departureAirportModel
+     * Initialize `#departureAirportModel`
      *
      * @for Fms
      * @method _initializeDepartureAirport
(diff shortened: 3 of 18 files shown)


src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -323,7 +288,12 @@ export default class AircraftCommandParser {
...
     _isSystemCommand(callsignOrSystemCommandName) {
-        return _has(SYSTEM_COMMANDS, callsignOrSystemCommandName) &&
-            callsignOrSystemCommandName !== SYSTEM_COMMANDS.transmit;
+        const command = AIRCRAFT_COMMAND_MAP[callsignOrSystemCommandName];
+
+        if (typeof command === 'undefined') {
+            return false;
+        }
+
+        return command.isSystemCommand && callsignOrSystemCommandName !== 'transmit';
```
n8rzz: `'transmit'` should be an enum

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandDefinitions.js
```diff
@@ -75,7 +75,7 @@ const ZERO_ARG_AIRCRAFT_COMMANDS = {
         validate: zeroArgumentsValidator,
         parse: noop
     },
-    climbViaSID: {
+    climbViaSid: {
```
n8rzz: 👍 

I've never been a fan of the `climbViaSID` capitalization.  thank you for changing this!

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
```diff
@@ -1,40 +1,6 @@
...
- * @property UNICODE_COMMANDS
- * @type {Object}
- * @final
- */
-const UNICODE_COMMANDS = {
-    '\\u2B50': 'land'
-};
+import _findKey from 'lodash/findKey';
 
+// FIXME: These keys should be upper-cased, but waiting for now to keep things working
```
n8rzz: why should they be upper-cased?  since these map directly to method names, I'd say it makes sense for them to be camelCase.  We should probably have a comment as to _why_ they are camelCased, though.

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
```diff
@@ -54,67 +20,191 @@ const UNICODE_COMMANDS = {
...
+        functionName: 'runHeading',
+        isSystemCommand: false
+    },
+    hold: {
+        aliases: ['hold'],
+        functionName: 'runHold',
+        isSystemCommand: false
+    },
+    land: {
+        aliases: ['\\u2B50', '*', 'i', 'ils'],
```
n8rzz: are we still handling unicode commands?

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
```diff
@@ -123,3 +213,14 @@ export const AIRCRAFT_COMMAND_MAP = {
...
 export const EXPEDITE = ['expedite', 'x'];
+
+/**
+ * Get the name of a command when given any of that command's aliases
+ *
+ * @function getCommandNameFromAlias
+ * @param alias {string}
+ * @return {string}
+ */
+export function getCommandNameFromAlias(alias) {
```
n8rzz: `.findCommandNameWithAlias()`

and could we change the param name to: `commandAlias`

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
```diff
@@ -54,67 +20,191 @@ const UNICODE_COMMANDS = {
  * @final
  */
 export const AIRCRAFT_COMMAND_MAP = {
```
n8rzz: nice work here!  I really like the simplification you've done here.

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
sam-irl: No; I believe if an arrow key is pressed it's replaced with `c` or `d` in the command bar. So a Unicode character would never be sent to be parsed.

erikquinn: this is the star that's inserted when pressing the numpad `*`. I'm not sure if it's in use, was one of the loose ends to be cleaned up when I get to it. Still a work in progress of course.

erikquinn: it was part of a `UNICODE_COMMANDS` enum that was `...`ed into the `AIRCRAFT_COMMANDS_MAP`. so i haven't changed anything really, but nonetheless intend to find out if it should be removed or not.

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: works for me. It's a helper function though-- you don't mean to add it to a class or anything do you? Just since you used a dot in front.

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -122,7 +119,7 @@ export default class AircraftCommandParser {
      * @return {string|array<string>}
      */
     get args() {
-        if (this.command !== SYSTEM_COMMANDS.transmit) {
+        if (this.command !== 'transmit') {
```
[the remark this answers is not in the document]
erikquinn: that's the plan

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
n8rzz: Oh, no, it's just fine as a function.

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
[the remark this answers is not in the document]
erikquinn: done

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -187,7 +184,7 @@ export default class AircraftCommandParser {
      * @private
      */
     _buildTransmitAircraftCommandModels(callsignOrSystemCommandName, commandArgSegments) {
-        this.command = SYSTEM_COMMANDS.transmit;
+        this.command = 'transmit';
```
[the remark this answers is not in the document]
erikquinn: done

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
erikquinn: done

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandDefinitions.js
erikquinn: yeah, it used to make me _crazy_ to see `Sid` because that's like seeing `Usa`: it's just wrong. But I've come around to the convention of keeping _EVERYTHING_ proper camel cased. Makes it easy and consistent, so here we are. 😃

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: Kinda just because it's been the pattern. If it doesn't bother you, I'll leave it because frankly I've grown tired of this PR anyway. 😉

erikquinn: turns out there was already a comment to this effect, so i simply removed the fixme.

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: 🎉

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: not really needed. removed, and redirected the hotkey to insert the `*` alias.

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: done

erikquinn:
likely to be done as part of the route/fms refactor underway.

erikquinn:
not done under that refactor. Will need to reassess/restart/resume this at some point.

src/assets/scripts/client/InputController.js
```diff
@@ -406,7 +406,7 @@ export default class InputController {
 
                 break;
             case KEY_CODES.MULTIPLY:
-                this.$commandInput.val(`${currentCommandInputValue} \u2B50 `);
+                this.$commandInput.val(`${currentCommandInputValue} * `);
```
n8rzz: why the change here?  

remember, the `*` is interpreted differently by operating system.  the `*` on windows is not the same as a `*` on osx, hence the unicode 😉

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -361,6 +316,34 @@ export default class AircraftCommander {
...
+     * @param data {array}
+     * @return {array} [success of operation, response]
+     */
+    runExpectArrivalRunway(aircraft, data) {
+        const airportModel = AirportController.airport_get();
+        const runwayName = data[0];
+        const runwayModel = airportModel.getRunway(runwayName);
+
+        if (_isNil(runwayModel)) {
+            const oldRunwayModel = aircraft.fms.arrivalRunwayModel;
```
n8rzz: what does `old` mean here?   previous, perhaps?

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -592,19 +575,23 @@ export default class AircraftCommander {
...
+        const runwayModel = AirportController.airport_get().getRunway(taxiDestination.toUpperCase());
+
+        if (!runwayModel) {
+            return [false, `unable to find Runway ${taxiDestination.toUpperCase()} on our charts`];
+        }
 
-        if (!runway) {
-            return [false, `no runway ${taxiDestination.toUpperCase()}`];
+        if (!aircraft.fms.isRunwayModelValidForSid(runwayModel)) {
+            aircraft.pilot.cancelDepartureClearance(aircraft);
```
n8rzz: does `.cancelDepartureClearance()` actually need the `aircraft` param?  seems a little weird, since it's the same instance calling that method.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -592,19 +575,23 @@ export default class AircraftCommander {
...
         }
 
-        const readback = aircraft.pilot.taxiToRunway(runway, isDeparture, flightPhase);
+        const readback = aircraft.pilot.taxiToRunway(runwayModel, isDeparture, flightPhase);
 
         // TODO: this may need to live in a method on the aircraft somewhere
-        aircraft.fms.departureRunwayModel = runway;
+        aircraft.fms.departureRunwayModel = runwayModel;
```
n8rzz: do we have a method in the `Fms` to update the runway?  I am wary about setting things like this outside the class, it makes it hard to track things down.

I understand it was already like this, buuuuuut, now's the time to tackle these.  Or maybe just update the comment, this method should live in the `Fms`.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -636,6 +636,30 @@ export default class Fms {
         this.currentPhase === FLIGHT_PHASE.CLIMB;
     }
 
+    /**
+     * Facade for #_routeModel.isRunwayModelValidForSid()
```
n8rzz: missing bat ticks around `#_routeModel.isRunwayModelValidForSid()`

n8rzz: could we also add a blurb about _why_ this facade is needed?  for this and the next one below please.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -638,6 +638,54 @@ export default class RouteModel extends BaseModel {
...
+     * Returns whether the specified runway is valid for this route's STAR leg (if it has one)
+     *
+     * If there is no STAR, there is no issue with changing runways, so we would treat this as "valid"
+     *
+     * @for RouteModel
+     * @method isRunwayModelValidForStar
+     * @param runwayModel {RunwayModel}
+     * @return {boolean}
+     */
+    isRunwayModelValidForStar(runwayModel) {
```
n8rzz: I'd encourage you to see where these two methods could share logic.  there is a fair amount of repeated code between these two methods.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -785,29 +834,51 @@ export default class RouteModel extends BaseModel {
...
+            const procedureModelName = starLegModel.getProcedureName();
+            const readback = {};
+            readback.log = `unable, Runway ${runwayModel.name} is not valid for the ${procedureModelIcao} arrival`;
+            readback.say = `unable, Runway ${runwayModel.name} is not valid for the ${procedureModelName} arrival`;
+
+            return [false, readback];
+        }
+
+        const amendedStarLegModel = this._createAmendedStarLegUsingDifferentExitName(nextExitName, starLegIndex);
+
```
n8rzz: remove empty line

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -381,7 +399,20 @@ export default class Pilot {
      * @param {Number} initialAltitude  the altitude aircraft can automatically climb to at this airport
      * @return {Array}                  [success of operation, readback]
      */
-    clearedAsFiled() {
+    clearedAsFiled(aircraft) {
+        const fms = aircraft.fms;
+        const departureRunwayModel = fms.departureRunwayModel;
```
n8rzz: `fms` is unnecessary here.  instead try:
```js
const { departureRunwayModel } = fms;
```

This also seems overly defensive.  We're checking that the `departureRunwayModel` that we've already set is valid?  Shouldn't that happen when we're setting it initially and not here?  This method should really just be like: _'Ok, we have your flight plan, we verified it on instantiation, go ahead and fly'_

n8rzz: I'm all for defensive programming, it just feels unnecessary here.

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -65,7 +63,7 @@ export default class AircraftCommandParser {
      * @for AircraftCommandParser
      * @param rawCommandWithArgs {string}  string present in the `$commandInput` when the user pressed `enter`
      */
-    constructor(rawCommandWithArgs = '') {
```
n8rzz: is this defaulted somewhere else?  why remove the default here?

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -299,7 +275,11 @@ export default class AircraftCommandParser {
      * @private
      */
     _validateCommandArguments() {
-        return _compact(_map(this.commandList, (command) => {
+        const validatedCommandList = _map(this.commandList, (command) => {
+            if (typeof command === 'undefined') {
+                return;
```
n8rzz: this should still return a falsey value like `null` because using the `compact` below will remove those values properly.  just makes things more clear.  yay verbosity!

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
```diff
@@ -54,67 +19,191 @@ const UNICODE_COMMANDS = {
...
+        functionName: 'runHeading',
+        isSystemCommand: false
+    },
+    hold: {
+        aliases: ['hold'],
+        functionName: 'runHold',
+        isSystemCommand: false
+    },
+    land: {
+        aliases: ['*', 'i', 'ils'],
```
n8rzz: be careful with the `*` again here.  this may not be interpreted the same across operating systems

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -738,7 +738,7 @@ ava('.updateSidLegForDepartureRunwayModel() calls LegModel.updateStarLegForArriv
     t.true(sidLegUpdateSidRunwaySpy.calledWithExactly(nextRunwayModel));
 });
 
-ava('.updateStarLegForArrivalRunwayModel() returns early when the route contains no STAR leg to update', (t) => {
+ava.skip('.updateStarLegForArrivalRunwayModel() returns early when the route contains no STAR leg to update', (t) => {
```
n8rzz: `ava.skip`!?  ☹️

test/aircraft/Pilot/clearedAsFiled.spec.js
```diff
@@ -20,7 +20,8 @@ ava.afterEach(() => {
     resetNavigationLibraryFixture();
 });
 
-ava('.clearedAsFiled() grants pilot departure clearance and returns the correct response strings', (t) => {
+// TODO: Needs AircraftModel fixture, not just pilot
```
n8rzz: I think one of those exists already?

src/assets/scripts/client/InputController.js
erikquinn: `u2B50` is the ⭐ character, which we aren't using anymore, it's been looking for a literal asterisk. Since that's what we're using in other cases, this simply updates the character being inserted. If you're saying that a string in js will be handled differently on a mac, then I guess that's something we'll need to deal with, albeit separately from here.

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: affirm, I can rename.

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: classic case of kill-the-pilot. When we get to removing it, it would simply be `aircraft.cancelDepartureClearance()`. Doing it this ugly way for now to keep with the convention.

src/assets/scripts/client/aircraft/AircraftCommander.js

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn:
```
Other classes will not have access to the `#_routeModel`, but can use this facade to determine whether or not a given runway assignment is allowable, given the departure procedure in the `#_routeModel`.
```
done

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: I guess they're kinda parallel, but I wouldn't say they're that similar... they're using different methods in like 3 different places. Plus, I can't even imagine a sensible name for a method I could abstract that out to, and my thought is that if it's too abstract to have a sensible name, it probably does more harm than good...

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js

src/assets/scripts/client/InputController.js
n8rzz: ahh, ok, I'll allow it 😉

src/assets/scripts/client/aircraft/AircraftCommander.js
n8rzz: #worksforme

src/assets/scripts/client/aircraft/Pilot/Pilot.js
erikquinn: Here's the scenario this protects against:
1. Aircraft is filed for SID valid only from Runway 36.
1. Aircraft is successfully  cleared as filed.
1. Aircraft is taxiied to Runway 18 (without complaints, expecting to get a new SID assigned), and the pilot silently cancels their own departure clearance because they cannot accept that rwy/sid combination.
1. Aircraft will refuse takeoff clearance until receiving a routing valid for their assigned runway, or after taxiied to a valid runway and cleared-as-filed.

So this is the portion where the pilot verifies their filed route is valid for the runway they've been assigned, and if it isn't, he will reject the route, prompting the controller to assign the appropriate SID instead.

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
erikquinn: Honestly, I'm not sure. Looked back and couldn't find any clues.... it's pretty apparent it's not _needed_, but at the same time I feel like there's no reason not to have it, so I'll add it back in.

erikquinn: done

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
erikquinn: done

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandMap.js
erikquinn: Added #919 to look into it

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: done!

test/aircraft/Pilot/clearedAsFiled.spec.js
erikquinn: It does. Due to some conflicts, i'd like to handle this test in #897 though if that's alright

src/assets/scripts/client/aircraft/AircraftCommander.js

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -772,6 +820,7 @@ export default class RouteModel extends BaseModel {
         return this.currentLeg.skipToWaypointName(waypointName);
     }
 
+    // FIXME: Presume I will need to do the same thing here as with .updateStarLegForArrivalRunwayModel() ?
```
n8rzz: Is this `FIXME` still valid?

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
```diff
@@ -323,7 +305,12 @@ export default class AircraftCommandParser {
      * @return {boolean}
      */
     _isSystemCommand(callsignOrSystemCommandName) {
-        return _has(SYSTEM_COMMANDS, callsignOrSystemCommandName) &&
-            callsignOrSystemCommandName !== SYSTEM_COMMANDS.transmit;
+        const command = AIRCRAFT_COMMAND_MAP[callsignOrSystemCommandName];
```
n8rzz: why not use `_has()` here?

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: nope, removed

src/assets/scripts/client/parsers/aircraftCommandParser/AircraftCommandParser.js
erikquinn: no longer needed, due to the rearrangement of the command map, and resulting adjustments to this method. should also read slightly clearer now as well, so that's also a plus!

done

PR #792, opened 2017-10-18
refactor/791

### LINKED ISSUE #791: Add console error when procedure draw is not 2D array
In airport files, all SIDs and STARs draw sections should be an array of arrays of strings. If `draw` is set to `[]`, the subsequent call of `.reduce()` will fail. A check should be added to this section creating a const for the draw section, and a console error added when the draw section array does not contain at least one empty array.

image

n8rzz:
Resolves #791 

Adds useful error message and tests for case when airport file has procedure with improperly defined `draw` segment

### DIFF
--- a/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
+++ b/src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
@@ -298,8 +298,18 @@ export default class StandardRouteModel extends BaseModel {
      * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
      */
     getAllFixNamesInUse() {
+        // if data has been incorrectly defined in the airport.json file, we throw and provide a useful message
+        // this should never throw in production, but should be useful for developers as they create and update airport files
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/navigationLibrary/StandardRoute/StandardRouteModel.js
```diff
@@ -298,9 +298,19 @@ export default class StandardRouteModel extends BaseModel {
      * @return {array<string>} ['fixname', 'fixname', 'fixname', ...]
      */
     getAllFixNamesInUse() {
+        // if data has been incorrectly defined in the airport.json file, we throw and provide a useful message
+        // this should never throw in production, but should be useful for developers as they create and update airport files
+        if (typeof this.draw[0] === 'undefined' || typeof this.draw[0] === 'string') {
```
erikquinn: why not a more strict `if (!_isArray(this.draw[0]) {`?

n8rzz: because screaming babies.  I should have done it that way, you're right.

n8rzz: done

PR #799, opened 2017-10-25
feature/236

n8rzz:
Resolves openscope/openscope#236 

- Simplifies canvas methods.  
- Splits canvas into `STATIC` and `DYNAMIC` canvases, only drawing updates when there are updates to draw.
- Moves `CanvasController` arrow functions to bound class methods

PR #800, opened 2017-10-25
bugfix/787

### LINKED ISSUE #787: SID command should not change destination
When using the `sid` command, the sid is written to the aircraft's `destination` property. We are now using that property differently (to represent the destination airport, not where they leave the airspace). The resulting effect is noticeable in the flight strip. The code being run on `sid` command should be examined to prevent this from setting any destination information.
image

#### Steps to reproduce
- At KSEA, any departure, issue `sid summa1`.

sam-irl:
Resolves #787 

The purpose of this pull request is to prevent the `sid` command from changing the `AircraftModel.destination` property


sam-irl:
blocked by #799

sam-irl:
No it isn't, wrong PR

sam-irl:
@n8rzz I tested on my machine and everything seems to work normally - the command does not change the destination property but the route changes to the new SID.

PR #846, opened 2017-12-17
refactor/809

### LINKED ISSUE #809: Standard Route & FMS Simplification
Refactors as needed to the standard route system and how the FMS interprets and works with them.

erikquinn:
Resolves #809.

The purpose of this pull request is to change some of the inner workings of the navigation systems to make future expansion and development easier.


src/assets/scripts/client/aircraft/AircraftConflict.js
```diff
@@ -114,7 +114,6 @@ export default class AircraftConflict {
 
         this._recalculateLateralAndVerticalDistances();
         this.checkCollision();
-        this.checkRunwayCollision();
```
n8rzz: Why is this removed?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1047,9 +1045,9 @@ export default class AircraftModel {
             head: 0
         };
 
-        const { wind } = AirportController.airport_get();
-        // const wind = airport.wind;
-        const angle = this.fms.currentRunway.calculateCrosswindAngleForRunway(wind.angle);
+        const { wind } = this.fms.arrivalAirportModel || this.fms.departureAirportModel;
+        const runwayModel = this.fms.arrivalRunwayModel || this.fms.departureRunwayModel;
```
n8rzz: I feel like these two lines should be encapsulated in methods. Destructuring with `||` here looks icky and is a pain in the ass to test.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1704,7 +1702,7 @@ export default class AircraftModel {
                 return headingToWaypoint;
             }
 
-            this.fms.nextWaypoint();
+            this.fms.moveToNextWaypoint();
```
n8rzz: 👍 on the name change here

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -79,41 +55,25 @@ export default class Fms {
     /**
      * @constructor
      * @param aircraftInitProps {object}
-     * @param initialRunwayAssignment {RunwayModel}
      * @param typeDefinitionModel {AircraftTypeDefinitionModel}
      * @param navigationLibrary {NavigationLibrary}
      */
-    constructor(aircraftInitProps, initialRunwayAssignment, typeDefinitionModel, navigationLibrary) {
+    constructor(aircraftInitProps, typeDefinitionModel, navigationLibrary) {
```
n8rzz: typeDefinitionModel -> aircraftTypeDefinitionModel

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,687 @@ export default class Fms {
...
-
-        return _flatten(waypointList);
+        return this._routeModel.waypoints;
     }
 
+    // ------------------------------ LIFECYCLE ------------------------------
+
     /**
-     * Initialize the instance and setup initial properties
+     * Initialize class properties
```
n8rzz: class -> instance 
 Class properties would signify a definition of the properties that belong to a class. Instance properties are the values of those properties for a particular instance.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,687 @@ export default class Fms {
...
      * @param aircraftInitProps {object}
+     * @param navigationLibrary {NavigationLibrary}
+     * @chainable
      */
-    init({ altitude, category, model, route }, initialRunwayAssignment) {
-        this._setCurrentPhaseFromCategory(category);
-        this._setInitialRunwayAssignmentFromCategory(category, initialRunwayAssignment);
-
-        this.flightPlanAltitude = altitude;
+    init({ altitude, category, destination, model, nextFix, origin, routeString }, navigationLibrary) {
```
n8rzz: This destructuring should happen on the next line. Doing it here is a little confusing, the method signature should match the comments.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,687 @@ export default class Fms {
...
      *
      * @for LegModel
-     * @method destroy
+     * @method reset
+     * @chainable
      */
-    destroy() {
-        this._navigationLibrary = null;
-        this._previousRouteSegments = [];
+    reset() {
```
n8rzz: Why the name change here?  We’ve reserved `.reset()` for now pooled classes. Is the Fms now pooped too?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,687 @@ export default class Fms {
...
+        this.arrivalAirportModel = null;
+        this.arrivalRunwayModel = null;
```
n8rzz: 😬 if we have the `airportModel`, why do we also need the `runwayModel`?  That already lives with the airport, no?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,687 @@ export default class Fms {
...
-        this.arrivalRunwayModel = runwayModel;
+        if (!this._routeModel.hasWaypointName(fixName)) {
+            throw new TypeError(`Expected initial fix to be in flight plan route, but received '${fixName}'`);
+        }
 
-        this._regenerateStarLeg();
+        this.skipToWaypointName(fixName);
     }
 
+    // ------------------------------ PUBLIC ------------------------------
```
n8rzz: Public methods should be next after init methods, then private after that.

src/assets/scripts/client/aircraft/AircraftConflict.js
erikquinn: This checked to see if two aircraft were intending to use the same runway opposite direction. It's an unnecessary check that's outside the scope of a radar sim. All that matters is distance and direction, which are checked elsewhere. This was a very old method.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: I agree in principle. This particular method is one I have an issue with because it asks for properties of an airport and of a runway without specifying which airport or which runway they're querying. I think that is a bad practice and the method itself should be changed. I would not want to add a new method that does the `||` logic because frankly that's not an operation we should be doing. We should be requesting information more specifically than that. As a similar example, you'll see that the `Fms.currentRunway` getter is gone for the same reason. There are better and more organized ways to get what we need. I just didn't take the time to redo this wind method. I can add a todo if nothing else because it _is_ quite gross.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: Ideally I'd think just about everything should be pooled, especially the classes that are reused in volume (WaypointModel, LegModel, RouteModel, Fms, AircraftModel). And I did not make the connection of why some were destroy and some were reset but since they reset the class (and I would assume destroy to also delete the instance), I just preferred the `reset` name. Also it didn't hurt that `reset` comes after `init` when `destroy` comes before `init` when you sort them. :smirk:

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: There are multiple runways at a given airport of course, so which one we are planning to use is needed. We could store the runways name as a string and look it up, but since `#arrivalRunwayModel` is only a pointer anyway, we might as well reference the model since it will frequently be needed for various things.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+     * @constructor
+     * @param navigationLibrary {NavigationLibrary}
+     * @param routeString {string}
+     */
+    constructor(navigationLibrary, routeString) {
+        if (!(navigationLibrary instanceof NavigationLibrary)) {
+            throw new TypeError(`Expected valid navigationLibrary, but received ${typeof navigationLibrary}`);
+        }
+
+        super();
```
n8rzz: this should be the first line of the constructor.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+    }
+
+    /**
+     * Return the next `WaypointModel`, from current or future leg
+     *
+     * @for RouteModel
+     * @property nextWaypoint
+     * @type {WaypointModel}
+     */
+    get nextWaypoint() {
```
n8rzz: there's a lot of logic here for a getter.  this is bordering on method territory.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+    }
+
+    /**
+    * Return an array of waypoints in the flight plan that have altitude restrictions
+    *
+    * @for RouteModel
+    * @method getAltitudeRestrictedWaypoints
+    * @return {array<WaypointModel>}
+    */
+    getAltitudeRestrictedWaypoints() {
```
n8rzz: use `filter` in the name here because we are not _getting_ we're filtering 😉

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+
+    // FIXME: COMPLETE THIS METHOD
+    /**
+     * Merge the provided route model into this route model, if possible
+     *
+     * @for RouteModel
+     * @method absorbRouteModel
+     * @param routeModel {RouteModel}
+     */
+    absorbRouteModel(/* routeModel */) {
```
n8rzz: this, once implemented, should likely return a _new instance_ of a route model.  that should make things much easier, logic-wise because all you need to do is figure out how the route strings should be combined.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+     * Return the `AirportModel` at whose runway this route will terminate
+     *
+     * @for RouteModel
+     * @method getArrivalRunwayAirportModel
+     * @return {AirportModel}
+     */
+    getArrivalRunwayAirportModel() {
+        const airportIcao = this.getArrivalRunwayAirportIcao();
+
+        if (airportIcao === null) {
```
n8rzz: two things here:
1) is `airportIcao` a string?
    1a) if yes, then it should never be `null`. it should be an empty string
2) if `null` is in fact an appropriate data type, this should instead check for `!airportIcao` instead of `null`

doing this:
```javascript
airportIcao === null
```
is not the same as
```javascript
!airportIcao
```

yay javascript!

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,902 @@
...
+     * Return the `RunwayModel` at which this route will terminate
+     *
+     * @for RouteModel
+     * @method getArrivalRunwayModel
+     * @return {RunwayModel}
+     */
+    getArrivalRunwayModel() {
+        const arrivalRunwayName = this.getArrivalRunwayName();
+
+        if (arrivalRunwayName === null) {
```
n8rzz: same thoughts here as `airportIcao` above

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -328,19 +327,20 @@ export default class AircraftCommander {
      * @return {array} [success of operation, readback]
      */
     runHold(aircraft, data) {
-        const turnDirection = data[0];
-        const legLength = data[1];
-        const holdFix = data[2];
```
n8rzz: we should keep these re-assignments because it adds additional context to otherwise anonymous variables.  `data[2]` means nothing whereas `holdFix` does.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -387,42 +387,9 @@ export default class AircraftCommander {
...
-     * to the `Pilot`.
-     *
-     * @for AircraftCommander
-     * @method _runSIDforSuffix
-     * @param  aircraft {AircraftModel}
-     * @param airportModel {AirportModel}
-     * @param sidId {strig}
-     * @return {array}  [success of operation, readback]
-     */
-    _runSIDforSuffix(aircraft, airportModel, sidId) {
```
n8rzz: I'm excited to see how we're handling suffixes now...

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -660,7 +660,7 @@ export default class AircraftController {
             positionModel: dynamicPositionModel,
             icao: aircraftTypeDefinition.icao,
             model: aircraftTypeDefinition,
-            route: spawnPatternModel.routeString,
+            routeString: spawnPatternModel.routeString,
```
n8rzz: mmm, much better name 👍

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -106,6 +106,7 @@ export default class AircraftModel {
          */
         this.id = _uniqueId('aircraft-');
 
+        // TODO: Does the aircraft itself really need this?
```
n8rzz: It did and I'd assume it still does.  But maybe not anymore?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -560,8 +561,8 @@ export default class AircraftModel {
 
     initFms(data) {
         const airport = AirportController.airport_get();
-        const initialRunway = airport.getActiveRunwayForCategory(this.category);
-        this.fms = new Fms(data, initialRunway, this.model, this._navigationLibrary);
+        // const initialRunway = airport.getActiveRunwayForCategory(this.category);
+        this.fms = new Fms(data, this.model, this._navigationLibrary);
```
n8rzz: To answer the `TODO` question about the `navigationLibrary`; yes, the `AircraftModel` does need a reference to it 😉

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -954,8 +951,9 @@ export default class AircraftModel {
      */
     isOnGround() {
         const errorAllowanceInFeet = 5;
-        const airport = AirportController.airport_get();
-        const nearRunwayAltitude = abs(this.altitude - this.fms.currentRunway.elevation) < errorAllowanceInFeet;
+        const airport = this.fms.arrivalAirportModel || this.fms.departureAirportModel;
+        const runwayModel = this.fms.arrivalRunwayModel || this.fms.departureRunwayModel;
```
n8rzz: these two lines make me nervous.  can we abstract this logic to the `fms` somehow?  How would we handle cases were both `arrivalRunwayModel` and `departureRunwayModel` are set in the `fms`?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -954,8 +951,9 @@ export default class AircraftModel {
...
+        const nearRunwayAltitude = abs(this.altitude - runwayModel.elevation) < errorAllowanceInFeet;
         const nearAirportAltitude = abs(this.altitude - airport.elevation) < errorAllowanceInFeet;
```
n8rzz: these two consts should be abstracted to class methods: `.isNearRunwayAltitude()` and `AirportModel.isAirportAltitude(this.altitude)`

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: seems I left a similar comment above just now.  we can discuss later...

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1714,19 +1698,21 @@ export default class AircraftModel {
...
-        const invalidTimerValue = -999;
-        const { hold } = this.fms.currentWaypoint;
-        const outboundHeading = radians_normalize(hold.inboundHeading + Math.PI);
-        const offset = getOffset(this, hold.fixPos, hold.inboundHeading);
-        const holdLegDurationInSeconds = hold.legLength * TIME.ONE_MINUTE_IN_SECONDS;
-        const bearingToHoldFix = vradial(vsub(hold.fixPos, this.relativePosition));
+     * @method _calculateTargetedHeadingHold
+     */
+    _calculateTargetedHeadingHold() {
+        const currentWaypoint = this.fms.currentWaypoint;
```
n8rzz: you can use destructuring here.

```javascript
const {
  holdParameters,
  relativePosition,
} = this.fms.currentWaypoint;
```

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1714,19 +1698,21 @@ export default class AircraftModel {
...
+        const holdParameters = currentWaypoint.holdParameters;
+        const waypointRelativePosition = currentWaypoint.relativePosition;
+        const outboundHeading = radians_normalize(holdParameters.inboundHeading + Math.PI);
+        const offset = getOffset(this, waypointRelativePosition, holdParameters.inboundHeading);
+        const holdLegDurationInMinutes = holdParameters.legLength.replace('min', '');
+        const holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
```
n8rzz: why not calculate these two on creation of a waypoint?  that way you can simply reference `holdParameters.legDurationInMinutes` and `holdParameters.legDurationInSeconds`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -79,41 +55,25 @@ export default class Fms {
...
-        *
-        * Used to generate a routeString for an entire, original, route
-        *
-        * @property _previousRouteSegments
-        * @type {array<string>}
-        * @default []
-        * @private
-        */
-        this._previousRouteSegments = [];
+        this.arrivalAirportModel = null;
```
n8rzz: I think we've talked about this before, but I am _very_ wary of including an `AirportModel` instance as part of the `fms`.  I'm sure you have good reasons, but I feel like this is dangerous.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * @method getTopAltitude
-     * @return {number}
+     * @for Fms
+     * @method _initializeArrivalAirport
+     * @param destinationIcao {string} ICAO identifier specified by spawn pattern
+     * @private
      */
-    getTopAltitude() {
-        const maxAltitudeFromLegs = _map(this.legCollection, (leg) => leg.getProcedureTopAltitude());
+    _initializeArrivalAirport(destinationIcao) {
```
n8rzz: `destinationIcao` is confusing here.  why not just call it `airportIco`?  The method name already provides enough context for what is going on so I feel this parameter can have a more generic name.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+     * @private
      */
-    getBottomAltitude() {
-        const valueToExclude = INVALID_NUMBER;
-        const minAltitudeFromLegs = _without(
-            _map(this.legCollection, (leg) => leg.getProcedureBottomAltitude()),
-            valueToExclude
-        );
+    _initializeArrivalRunway() {
+        if (this.arrivalAirportModel === null) {
```
n8rzz: assuming this is initialized as `null`, a simple `!this.arrivalAirportModel` should work here.  Using `=== null` is not doing what you think and can lead to errors in the future.  

I'd prefer `!this.arrivalAirportModel` instead of  `this.arrivalAirportModel === null` in this case (and any others using `=== null`).

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * Get the flight plan route string in dot notation
-     *
-     * @for Fms
-     * @method getFlightPlanRouteStringWithDots
-     * @return {string}
-     */
-    getFlightPlanRouteStringWithDots() {
-        const currentRouteSegments = _map(this.legCollection, (legModel) => legModel.routeString);
-        const previousAndCurrentRouteSegments = this._previousRouteSegments.concat(currentRouteSegments);
+        if (arrivalRunwayName === null) {
```
n8rzz: `arrivalRunwayName === null` should be `!arrivalRunwayName`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
      * @for Fms
-     * @method getFlightPlanRouteStringWithSpaces
-     * @return {string}
+     * @method _initializeDepartureAirport
+     * @param originIcao {string} ICAO identifier specified by spawn pattern
+     * @private
      */
-    getFlightPlanRouteStringWithSpaces() {
-        return this.getFlightPlanRouteStringWithDots().replace(REGEX.DOUBLE_OR_SINGLE_DOT, ' ');
+    _initializeDepartureAirport(originIcao) {
```
n8rzz: change `originIcao` to simply `airportIcao`.  Origin suggests additional context that may/or may not apply (in the future).

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+        if (originIcao === '') {
```
n8rzz: would `airportIcao` ever not be empty and invalid? would it ever not be a string?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * @method getNextWaypointModel
-     * @return waypointModel {WaypointModel}
+     * @method _initializeDepartureRunway
+     * @private
      */
-    getNextWaypointModel() {
-        if (!this.hasNextWaypoint()) {
-            return null;
+    _initializeDepartureRunway() {
+        if (this.departureAirportModel === null) {
```
n8rzz: `this.departureAirportModel === null` should be `!this.departureAirportModel`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+            return;
         }
 
-        let waypointModel = this.currentLeg.nextWaypoint;
+        const departureRunwayName = this._routeModel.getDepartureRunwayName();
 
-        if (!this.currentLeg.hasNextWaypoint()) {
-            waypointModel = this.legCollection[1].currentWaypoint;
+        if (departureRunwayName === null) {
```
n8rzz: `departureRunwayName === null` should be `!departureRunwayName`

_this is the last one I'll point out.  you get the idea 😉_

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * Get the position of the next waypoint in the flight plan
-     *
-     * Currently only used in `calculateTurnInitiaionDistance()` helper function
+     * Initialize #currentPhase as appropriate based on the spawn category
      *
      * @for Fms
-     * @method getNextWaypointPositionModel
-     * @return waypointPosition {StaticPositionModel}
+     * @method _initializeFlightPhaseForCategory
+     * @param category {string}
```
n8rzz: missing `return` type

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
     /**
-     * Return an array of waypoints in the flight plan that have speed restrictions
+     * Initialize #flightPlanAltitude
      *
      * @for Fms
-     * @method getSpeedRestrictedWaypoints
-     * @return {array<WaypointModel>}
+     * @method _initializeFlightPlanAltitude
+     * @param altitude {number}
+     * @param category {string} one of `FLIGHT_CATEGORY` enum
```
n8rzz: this can be written thusly:
```
@param category {FLIGHT_CATEGORY|string}
```

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+     * @param model {AircraftTypeDefinitionModel}
+     * @private
      */
-    getSpeedRestrictedWaypoints() {
-        return this.waypoints.filter((waypoint) => waypoint.hasSpeedRestriction);
+    _initializeFlightPlanAltitude(altitude, category, model) {
+        this.flightPlanAltitude = altitude;
+
+        if (category === FLIGHT_CATEGORY.DEPARTURE) {
```
n8rzz: reverse the logic here: 
```javascript
if (category !== FLIGHT_CATEGORY.DEPARTURE) {
  return;
}

this.flightPlanAltitude = model.ceiling;
```

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-        // TODO: this should be an `instanceof` check and should be implemented as part of (or after)
-        // https://github.com/openscope/openscope/issues/93
-        if (!_isObject(runwayModel)) {
-            throw new TypeError(`Expected instance of RunwayModel, but received ${runwayModel}`);
-        }
+    // ------------------------------ PUBLIC ------------------------------
 
-        if (this.arrivalRunwayModel && this.arrivalRunwayModel.name === runwayModel.name) {
-            return;
+    activateHoldForWaypointName(waypointName, holdParameters) {
```
n8rzz: missing docblock

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-            return new TypeError(`Expected known flight phase, but received '${phase}'`);
-        }
+    getAltitudeRestrictedWaypoints() {
+        return this._routeModel.getAltitudeRestrictedWaypoints();
+    }
 
-        if (this.currentPhase === phase) {
-            return;
-        }
+    /**
```
n8rzz: fix spacing here

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    *
+    * @for Fms
+    * @method getBottomAltitude
+    * @return {number}
+    */
+    getBottomAltitude() {
+        return this._routeModel.getBottomAltitude();
+    }
+
+    /**
```
n8rzz: fix spacing.  there are a couple of these in this class, take a spin through and see what you find.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    * @method getRestrictedWaypoints
+    * @return {array<WaypointModel>}
+    */
+    getRestrictedWaypoints() {
+        return this.waypoints.filter((waypoint) => waypoint.hasRestriction);
     }
 
     /**
-     * Add a new `LegModel` to the left side of the `#legCollection`
+     * Get the flight plan route string in dot notation
```
n8rzz: Why is it important to get dot notation?  meaning, what could this be used for?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    /**
+    *
+    *
+    * @for Fms
+    * @method isArrival
+    * @return {boolean}
+    */
+    isArrival() {
+        return this.currentPhase === FLIGHT_PHASE.CRUISE ||
+        this.currentPhase === FLIGHT_PHASE.DESCENT ||
```
n8rzz: fix spacing

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    /**
+    *
+    *
+    * @for Fms
+    * @method isDeparture
+    * @return {boolean}
+    */
+    isDeparture() {
+        return this.currentPhase === FLIGHT_PHASE.APRON ||
+        this.currentPhase === FLIGHT_PHASE.TAXI ||
```
n8rzz: fix spacing

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    * @param waypointName {string}
+    * @return {boolean}
+    */
+    hasWaypointName(waypointName) {
+        return this._routeModel.hasWaypointName(waypointName);
+    }
 
-        const legModel = this._createLegWithHoldWaypoint(waypointProps);
+    /**
+    *
```
n8rzz: missing description

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
+    isArrival() {
+        return this.currentPhase === FLIGHT_PHASE.CRUISE ||
+        this.currentPhase === FLIGHT_PHASE.DESCENT ||
+        this.currentPhase === FLIGHT_PHASE.APPROACH ||
+        this.currentPhase === FLIGHT_PHASE.LANDING;
+    }
 
-        this.prependLeg(legModel);
+    /**
+    *
```
n8rzz: missing description

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
 
-        return;
+    /**
+    * Move to the next possible waypoint
+    *
+    * This could be the next waypoint in the current leg,
+    * or the first waypoint in the next leg.
+    *
+    * @for LegModel
+    * @method moveToNextWaypoint
```
n8rzz: missing return type

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * the `legsColelction`. Then make that Leg active and `waypointName`
-     * the active waypoint for that Leg.
-     *
-     * @for Fms
-     * @method skipToWaypoint
-     * @param waypointName {string}
-     */
-    skipToWaypoint(waypointName) {
-        const { legIndex, waypointIndex } = this._findLegAndWaypointIndexForWaypointName(waypointName);
+        if (wasSuccessful) {
```
n8rzz: flip this logic.  have the if block here contain the fail state

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * @param runway {string}
-     */
-    replaceFlightPlanWithNewRoute(routeString, runway) {
-        // TODO: we may need to update the runway in this method
-        this._destroyLegCollection();
+        const wasSuccessful = this._routeModel.replaceDepartureProcedure(routeString);
 
-        this.legCollection = this._buildLegCollection(routeString);
-    }
+        if (wasSuccessful) {
```
n8rzz: flip this logic. have the if block here contain the fail state

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-        // flightPhase is unknown or unavailable so it must be extrapolated based on the `procedureId`.
-        if (flightPhase === '') {
-            flightPhase = this._translateProcedureNameToFlightPhase(routeStringModel.procedure);
-        }
-
-        // a `LegModel` already exists with this routeString
-        if (this.hasLegWithRouteString(routeStringModel.routeCode)) {
-            return true;
-        }
+        this._routeModel = nextRouteModel;
```
n8rzz: is there a reason this is assigned here and with this intermediate var?  why not just:
```javascript
try {
  this._routeModel = new RouteModel(this._navigationLibrary, routeString);
} catch (error) {
  // ...
}
```

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * @return {LegModel|undefined}
-     * @private
-     */
-    _findLegByRouteString(routeString) {
-        return _find(this.legCollection, { routeString: routeString.toLowerCase() });
+    * Fascade
+    *
+    * @for Fms
+    * @method skipToWaypointName
+    * @param waypointName {string}
```
n8rzz: missing return type

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-            // departure runway to throw. So instead will just return early.
-            return;
-        }
-
-        const sidLeg = this.legCollection[sidLegIndex];
-
-        this.replaceDepartureProcedure(sidLeg.routeString, this.departureRunwayModel);
-    }
+    _updateArrivalRunwayFromRoute() {
+        const arrivalRunwayModel = this._routeModel.getArrivalRunwayModel();
```
n8rzz: without having seen the `RouteModel`, I feel like this should be a simple getter.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-     * @param category {string}
-     * @private
+     * @method _updateArrivalRunwayFromRoute
      */
-    _setCurrentPhaseFromCategory(category) {
-        switch (category) {
-            case FLIGHT_CATEGORY.ARRIVAL:
-                this.setFlightPhase(FLIGHT_PHASE.CRUISE);
+    _updateDepartureRunwayFromRoute() {
+        const departureRunwayModel = this._routeModel.getDepartureRunwayModel();
```
n8rzz: without having seen the `RouteModel`, I feel like this should be a simple getter.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -304,1237 +234,653 @@ export default class Fms {
...
-        // correct order of legs.
-        const routeSegmentList = routeSegments.slice().reverse();
-
-        for (let i = 0; i < routeSegmentList.length; i++) {
-            const routeSegment = routeSegmentList[i];
-            const legModel = this._buildLegModelFromRouteSegment(routeSegment);
-
-            this.prependLeg(legModel);
+    _verifyRouteContainsMultipleWaypoints() {
+        if (this.waypoints.length < 2) {
```
n8rzz: reverse logic and return early for valid cases, then handle the hidden `else` afterwards with the `throw`

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
-         * Provide easy access to the parts of a procedureRouteString,
-         * when a leg is a procedure
+         * Type of leg from list of types defined in `LEG_TYPE`
          *
-         * @property _procedureRouteModel
-         * @type {RouteModel|null}
-         * @default null
+         * @for LegModel
+         * @property _legType
+         * @type {string}
```
n8rzz: you can docblock this thusly:
```
@type {LEG_TYPE|string}
```

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
-         * @type {PROCEDURE_TYPE}
-         * @default ''
-         */
-        this.procedureType = '';
+        this._previousWaypointCollection = [];
 
         /**
-         * String representation of the current routeSegment.
-         *
-         * A directRoute contains a single WaypointModel and is separated
```
n8rzz: I'd still like to keep this definition if it still makes sense.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
         /**
-         * List of `WaypointModel` objects defined within a `LegModel`.
+         * Array of `WaypointModel`s to follow, excluding any waypoints passed (or skipped)
          *
-         * If this leg represents a `directRoute`, there will only be one
-         * `WaypointModel` contained in `#waypointCollection`.
+         * Upon completion of a given `WaypointModel` in the `#_waypointCollection`, that
+         * waypoint will be moved to the `#_previousWaypointCollection`, and the aircraft will
+         * continue to the next `WaypointModel` in the `#_waypointCollection`.
          *
```
n8rzz: can we add a bit about the significance of the `0` index item in this collection?

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+     * Returns the active `WaypointModel`
      *
-     * @property procedureName
-     * @type {string}
-     * @return {string}
+     * Assumed to always be the first item in the `#waypointCollection`
+     *
+     * @for LegModel
+     * @property currentWaypoint
+     * @type {WaypointModel}
```
n8rzz: missing return type

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
     }
 
     /**
-     * Return the name of the procedure exit (airport) if this is a procedure leg
+     * Returns whether this leg is an airway leg
      *
-     * This is used only for arrival routes and only within the `AircraftStripView`
+     * @for LegModel
+     * @property isAirwayLeg
+     * @type {boolean}
```
n8rzz: missing return type

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+
+    /**
+     * Returns whether this leg is a direct leg
      *
-     * @property exitName
-     * @type {string}
-     * @return {string}
+     * @for LegModel
+     * @property isDirectLeg
+     * @type {boolean}
```
n8rzz: missing return type

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+        return this.isProcedureLeg && this._procedureModel.procedureType === PROCEDURE_TYPE.STAR;
+    }
+
+    /**
+     * Returns the type of this leg
      *
-     * @property procedureAndExitName
+     * @for LegModel
+     * @property legType
      * @type {string}
```
n8rzz: is this a `string` or a property of an enum?  If the latter, that should be the type instead of `string`.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
-        }
-
-        return `${this._procedureRouteModel.procedure}.${this._procedureRouteModel.exit}`;
+    get legType() {
+        return this._legType;
     }
 
     /**
-     * The active `WaypointModel`.
+     * Returns the `WaypointModel` immediately following the `#currentWaypoint`
```
n8rzz: what does _immediately following_ mean?  this may be more accurate if we said the item at `1` index.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+     * @chainable
      */
-    init(routeSegment, runway, flightPhase, holdWaypointProps) {
-        this.isProcedure = RouteModel.isProcedureRouteString(routeSegment);
-        // TODO: replace with constant
-        this._isHold = RouteModel.isHoldRouteString(routeSegment) || routeSegment === 'GPS';
+    init(navigationLibrary, routeString) {
+        this._routeString = routeString;
+
+        const [entryOrFixName, airwayOrProcedureName, exit] = routeString.split(PROCEDURE_OR_AIRWAY_SEGMENT_DIVIDER);
```
n8rzz: don't we have a helper function/method for this somewhere?  If not, this logic _should_ be abstracted so it can be tested.  This logic should not be tied to a particular method like this, even though it is pretty simple, it's also very important and could potentially be used elsewhere.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+
+        this._ensureRouteStringIsSingleSegment(routeString);
```
n8rzz: so what happens if a route is not a single segment?  this name suggests a branch in the logic where none appears to happen here.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
- *
- * RouteSegment Examples:
- *  - directRouteSegment: 'COWBY'
- *  - holdRouteSegment: '@COWBY'
- *  - procedureRouteSegment: 'KLAS.COWBY6.DRK'
+ * A portion of a navigation route containing one or more `WaypointModel` objects.
  *
  * @class LegModel
  */
 export default class LegModel {
```
n8rzz: can this extend `BaseModel`?

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+     * @return {array<WaypointModel>}
      */
-    hasWaypoint(waypointName) {
-        // using a for loop here instead of `_find()` because this operation could happen a lot. a for
-        // loop is going to be faster than `_find()` in most cases.
-        for (let i = 0; i < this.waypointCollection.length; i++) {
-            const waypoint = this.waypointCollection[i];
+    _generateWaypointCollection(entryOrFixName, exit) {
+        if (this._legType === LEG_TYPE.DIRECT) {
+            return [new WaypointModel(entryOrFixName)];
```
n8rzz: I know why we are returning an array like this, we should add a comment here so anybody else would know too.

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+
+    /**
+     * Return the ICAO identifier for the airport at which this leg will terminate (if
+     * it is in fact a STAR leg, of course).
+     *
+     * @for LegModel
+     * @method getArrivalRunwayAirportIcao
+     * @return {string}
+     */
+    getArrivalRunwayAirportIcao() {
```
n8rzz: as written, we aren't _getting_ an airport icao here we are extracting it.

if this logic is extracted, like I mentioned previously, we should have easier access to this airport icao

src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel.js
```diff
@@ -1,538 +1,780 @@
...
+     * @return {string}
      */
-    getProcedureBottomAltitude() {
-        const isMaximum = false;
+    getArrivalRunwayName() {
+        if (!this.isStarLeg) {
+            return null;
+        }
+
+        const airportAndRunway = this._routeString.split(PROCEDURE_OR_AIRWAY_SEGMENT_DIVIDER)[2];
```
n8rzz: this route string logic should be abstracted.  see previous comments

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: changed to use instance throughout everything now.

done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: that's what we've got:
- constructor & getters (no label)
- "lifecycle"
- "public"
- "private"

considering done.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: fixed, and verified this was the only such instance of that.

done

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: For _proper_ route merging, it's actually far more complicated than merging the route strings themselves. For example, leaving the route in the middle of an airway leg: `instead of exiting the airway at this point, exit early here and join this other STAR`, etc.

As this has now actually been implemented, it does not return a new instance, but rather does just what it says: it takes in the supplied `RouteModel` instance, examines it and `this` in order to see if they have any point(s) of continuity, and if so, it adjusts the original routes legs as necessary and inserts the provided legs at the appropriate places. So it takes the provided route, breaks it down, and adds it to `this` route model.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: 1. It originally comes from `._getArrivalRunwayAirportModel()` called on line 256 in the preview thing above and when that finds nothing, it returns null. So this if block is checking to see if we failed to find an airport ICAO identifier in any leg's route strings. So perhaps an odd way to be doing it, but the logic aligns correctly. Now changed to at least use `!airportIcao` and will note that for later 😄

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
n8rzz: Let’s go with a method

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js

src/assets/scripts/client/aircraft/AircraftCommander.js

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: short answer is that we aren't. Apparently the approach we were trying to use to keep the procedure definitions manageably short by design can't accommodate some of the ways some procedures apparently are (that I didn't know about at the time). So now, each "suffix" is defined as its own procedure.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: It apparently needs it for _instantiation_ of the FMS, but after passing it along, the aircraft itself _does not_ need it. Removed.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: this has been changed to `navigationLibrary` as a local variable passed down the initialization chain, after which it dies! :)

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: restructured:
```
isOnGround() {
    let airportModel = this.fms.departureAirportModel;
    let runwayModel = this.fms.departureRunwayModel;

    if (this.isArrival()) {
        airportModel = this.fms.arrivalAirportModel;
        runwayModel = this.fms.arrivalRunwayModel;
    }

    const errorAllowanceInFeet = 5;
    const nearRunwayAltitude = abs(this.altitude - runwayModel.elevation) < errorAllowanceInFeet;
    const nearAirportAltitude = abs(this.altitude - airportModel.elevation) < errorAllowanceInFeet;

    return nearRunwayAltitude || nearAirportAltitude;
}
```

Also, we definitely do NOT want a method for the `||`ing because that is a bad practice, we should always request either the arrival runway or departure runway. If we're asking for a runway assignment without that context, we're doing _something_ wrong.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
n8rzz: no, it should be `filter` 😉 

first, because we are actually using `filter`.  second, because in JS filter means you get something back.

`.filterAltitudeRestrictedWaypoints()` reads exactly as it's doing, which is what we're going for.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+    /**
+     * Return an array of all waypoints in all legs of the route
+     *
+     * @for RouteModel
+     * @property waypoints
+     * @type {array<WaypointModel>}
+     */
+    get waypoints() {
+        return _reduce(this._legCollection, (waypointList, legModel) => {
+            return waypointList.concat(legModel.waypoints);
```
n8rzz: pretty sure you could just do:
```javascript
return [
  ...waypointList,
  ...legModel.waypoints
];
```

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+     *
+     * @for RouteModel
+     * @method absorbRouteModel
+     * @param routeModel {RouteModel}
+     * @return {array} [success of operation, response]
+     */
+    absorbRouteModel(routeModel) {
+        const firstWaypointName = _first(routeModel.waypoints).name;
+        const lastWaypointName = _last(routeModel.waypoints).name;
+        const routesConverge = this.hasWaypointName(lastWaypointName);
```
n8rzz: these `const`s should be prepended by `is`:  `isConverging` and `isDiverging`.  Since these are both local to the method, we already know the context is that of the `route`, so I don't think that also needs to be included in the `const` name.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+    * @method getFullRouteStringWithoutAirportsWithSpaces
+    * @return {string}
+    */
+    getFullRouteStringWithoutAirportsWithSpaces() {
+        const pastAndPresentLegModels = this._getPastAndPresentLegModels();
+        const legRouteStringsWithoutAirports = _map(pastAndPresentLegModels, (legModel) => {
+            return legModel.getRouteStringWithoutAirports();
+        });
+
+        return this._combineRouteStrings(legRouteStringsWithoutAirports)
```
n8rzz: update spacing

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+    /**
+     * Whether the route has another waypoint after the current one
+     *
+     * This includes waypoints in the current and future legs
+     *
+     * @for RouteModel
+     * @method hasNextWaypoint
+     * @return {boolean}
+     */
+    hasNextWaypoint() {
```
n8rzz: this whole method can be simplified:
```javascript
return this.currentLeg.hasNextWaypoint() || this.hasNextLeg() || !_isNil(this.nextLeg.currentWaypoint);
```

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+     * @method updateSidLegForDepartureRunwayModel
+     * @param runwayModel {RunwayModel}
+     */
+    updateSidLegForDepartureRunwayModel(runwayModel) {
+        if (!this.hasSidLeg()) {
+            return;
+        }
+
+        const sidLegIndex = this._findSidLegIndex();
+
```
n8rzz: remove extra space

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -0,0 +1,1481 @@
...
+    * @method updateStarLegForArrivalRunwayModel
+    * @param runwayModel {RunwayModel}
+    */
+    updateStarLegForArrivalRunwayModel(runwayModel) {
+        if (!this.hasStarLeg()) {
+            return;
+        }
+
+        const starLegIndex = this._findStarLegIndex();
+
```
n8rzz: remove extra line

erikquinn:
**WORK TO BE CONTINUED ON A FOLLOW-UP PULL REQUEST.**

Merging this as-is for now, with the intent to continue working on this, while also providing the ability for airport developers to get their stuff updated with it.

PR #867, opened 2018-01-19
enhancement/866

### LINKED ISSUE #866: More customisation of callsigns
Currently, callsigns are generated based on the length of callsign and whether they should be alphanumeric. 

In #code-help on Slack, it has been suggested that more customisation of the structures of callsigns should be available.

For example: 
`['###']` would generate a callsign with three numbers only (BAW286)
`['##AA']` would generate a callsign with two numbers and two letters (RYR62LA)
`['###', '##AA']` would generate callsigns with three numbers and two numbers and two letters.

indianbhaji:
Resolves openscope/openscope#866 

The purpose of this pull request is to implement more customisation of callsign formats


src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -68,26 +68,15 @@ export default class AirlineModel extends BaseModel {
          */
         this.flightNumberGeneration = {
```
nategeslin: can you add a `TODO` here about abstracting the flightNumberGenerator out to it's own helper.  There is a growing amount of logic here and I feel it will likely (though maybe not now) need to be abstracted.  

The `AirlineModel` shouldn't really care about the _how_ only know about the rules.  It should be something like: _Hey, given these rules, build me a flight number'_

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -68,26 +68,15 @@ export default class AirlineModel extends BaseModel {
...
-             *
-             * Used to generate a North American style registration number, minus the `N`
-             *
-             * @memberof flightNumberGeneration
-             * @property alphaNumeric
-             * @type {boolean}
-             * @default false
-             */
-            alphaNumeric: false
+            callsignFormats: ['###']
```
nategeslin: please abstract this default value to a constant defined above the class or in a constants file:
```javascript
const DEFAULT_CALLSIGN_FORMAT = '###';
```

then here you can do:
```javascript
callsignFormats: [DEFAULT_CALLSIGN_FORMAT];
```

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -218,29 +206,23 @@ export default class AirlineModel extends BaseModel {
...
-        let list = '0123456789';
-
-        // Start with a number other than zero
-        flightNumber += choose(list.substr(1));
-
-        if (!this.flightNumberGeneration.alphaNumeric) {
-            for (let i = 1; i < this.flightNumberGeneration.length; i++) {
-                flightNumber += choose(list);
+        const numbersList = '0123456789';
+        const alphaList = 'abcdefghijklmnopqrstuvwxyz';
```
nategeslin: please abstract both `numberList` and `alphaList` to constants

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -218,29 +206,23 @@ export default class AirlineModel extends BaseModel {
...
+        const formats = this.flightNumberGeneration.callsignFormats;
+
+        const chosenFormat = choose(formats);
```
nategeslin: `chosenFormat` should be on the next line after `formats` with an empty line between that and the `for`

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -218,29 +206,23 @@ export default class AirlineModel extends BaseModel {
...
-
-        // Start with a number other than zero
-        flightNumber += choose(list.substr(1));
-
-        if (!this.flightNumberGeneration.alphaNumeric) {
-            for (let i = 1; i < this.flightNumberGeneration.length; i++) {
-                flightNumber += choose(list);
+        const numbersList = '0123456789';
+        const alphaList = 'abcdefghijklmnopqrstuvwxyz';
+        const formats = this.flightNumberGeneration.callsignFormats;
```
nategeslin: you can use destructuring here:
```javascript
const { callsignFormats } = this.flightNumberGeneration.callsignFormats;
```

src/assets/scripts/client/airline/AirlineModel.js
nategeslin: what is `.choose()`?

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -218,29 +206,23 @@ export default class AirlineModel extends BaseModel {
...
+
+        const chosenFormat = choose(formats);
+        for (let i = 0; i < chosenFormat.length; i++) {
+            if (chosenFormat[i] === '#') {
```
nategeslin: these nested `if`s and `else`s make me cringe.  there is a better, cleaner way to do this that I'll post here later.

test/airline/AirlineModel.spec.js
```diff
@@ -93,27 +93,6 @@ ava('.getRandomAircraftType() calls ._getRandomAircraftTypeFromFleet() fleet is
     t.true(_getRandomAircraftTypeFromFleetSpy.calledWithExactly(fleetNameMock));
 });
 
-ava('.generateFlightNumber() creates a flightNumber made up of numbers if !flightNumberGeneration.alphaNumeric', (t) => {
```
nategeslin: with all the added functionality here, we're going to need a _lot_ more tests.  be we can talk about the _how, when and where_ over the slacks.

src/assets/scripts/client/airline/AirlineModel.js
willfrd: Its acutually called the callsign, not the flight number :D

n8rzz: I believe we're using it for both here

willfrd: I think (depending on the airline), they are usually different, for example, the flight number never contains letters, the callsign can.

indianbhaji: @willfrd we don't display the flight number here, (ie FR6251) so it really doesn't matter what it's called

willfrd: Yeahhh, just saying :-)

test/airline/AirlineModel.spec.js
n8rzz: I'd add these two tests back in, but they will fail.  so for now you can do `ava.skip()` and they won't actually affect the rest of the test suite.

these two tests can now be tweaked to work with the new updates.

test/airline/AirlineModel.spec.js
```diff
@@ -93,27 +93,6 @@ ava('.getRandomAircraftType() calls ._getRandomAircraftTypeFromFleet() fleet is
...
-    model.flightNumberGeneration.alphaNumeric = true;
-
-    const result = model.generateFlightNumber();
-
-    t.true(result.length === model.flightNumberGeneration.length);
-    t.true(alphaNumericRegex.test(result));
-});
-
 ava.skip('.generateFlightNumber() creates a flightNumber with a prefix', (t) => {
     const expectedResult = 'N';
```
n8rzz: @indianbhaji are we handling this case still?

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,51 @@
+import { choose } from '../utilities/generalUtilities';
+
+const ALPHA = 'abcdefghijklmnopqrstuvwxyz';
+const NUMERIC = '0123456789';
+
+/**
+ * This picks a random number. If it is the first value within the callsign (ie. i === 0), then it picks a
+ * number between 1 and 9. Otherwise, it picks a number between 0 and 9.
+*/
+const generateNumericValue = (i) => {
```
n8rzz: this should be a private function: `_generateNumericValue`

be sure to also add:
```
@function
@param
@return
```
and applicable data types to the docblock

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,51 @@
...
+        return choose(NUMERIC.substr(1));
+    }
+
+    return choose(NUMERIC);
+};
+
+/**
+ * Self-explanatory; this function picks a random letter and returns it.
+*/
+const generateAlphaValue = () => {
```
n8rzz: this should be a private function: `_generateAlphaValue` and please remove the _self-explanatory_ part of the docblock.

be sure to also add:
```
@function
@param
@return
```
and applicable data types to the docblock

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -205,42 +194,11 @@ export default class AirlineModel extends BaseModel {
         return this._getRandomAircraftTypeFromFleet(fleet);
     }
 
-    // TODO: the logic here can be simplified.
-    /**
```
n8rzz: we still need this docblock.  If you could add it back in, I think we'll be good to go here.

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -207,40 +196,15 @@ export default class AirlineModel extends BaseModel {
 
     // TODO: the logic here can be simplified.
     /**
-     * Create a flight number/identifier
```
n8rzz: well we still need the content of the docblock here too unless it no longer applies

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -207,40 +196,20 @@ export default class AirlineModel extends BaseModel {
...
-        }
-
-        list = 'abcdefghijklmnopqrstuvwxyz';
+        const flightNumber = flightNumberBuilder(this.flightNumberGeneration.callsignFormats);
 
-        // the end of an `N` style registration: `N322WT`
-        for (let i = 0; i < 2; i++) {
-            flightNumber += choose(list);
+        if (this.activeFlightNumbers.indexOf(flightNumber) !== -1) {
+            return this.generateFlightNumber();
```
erikquinn: @n8rzz There seems to be two layers of recursion going on here. Here we start over if the `AirlineModel.activeFlightNumbers` contains the flight number, and we would also start over one step up in the call stack if `AirlineController.flightNumbers` contains the flight number. The latter is the one that matters, and come to think of it, there really is no reason we should even have a `AirlineModel.activeFlightNumbers` property. Most likely, it should be removed to avoid confusion, and then there is no need for the recursion at this line (212), as it is covered in the caller method.

Also important to note that the current behavior is that flight numbers are unique to all aircraft (which is to say we will not end up with EIN118 _AND_ BAW118, because `"118"`  is already "in use". Though not realistic, this enables an important feature of being able to issue commands by typing partial callsigns rather than clicking around.

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,56 @@
...
+const NUMERIC = '0123456789';
+
+/**
+ * This picks a random number. If it is the first value within the callsign (ie. i === 0), then it picks a
+ * number between 1 and 9. Otherwise, it picks a number between 0 and 9.
+ * @function _generateNumericValue
+ * @param i {number}
+ * @return NUMERIC {string}
+*/
+const _generateNumericValue = (i) => {
```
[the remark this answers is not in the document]
erikquinn: Also, we do not need this to be an arrow function, so the standard old school function declaration is preferred in this case:
```javascript
function _generateRandomDigit(i) {
    ...
}
```

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,56 @@
...
+
+    return choose(NUMERIC);
+};
+
+/**
+ * This function picks a random letter and returns it.
+ * @function _generateAlphaValue
+ * @return ALPHA {string}
+*/
+const _generateAlphaValue = () => {
```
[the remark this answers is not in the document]
erikquinn: Also, we do not need this to be an arrow function, so the standard old school function declaration is preferred in this case:
```javascript
function _generateRandomLetter(i) {
    ...
}
```

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Accepts a list of callsign formats, which are defined in the airline files. It randomly selects one of these
+ * formats and generates a callsign based off this format.
+ *
+ * @function flightNumberBuilder
+ * @param callsignFormats {array}
+ * @return {string}
+*/
+export const flightNumberBuilder = (callsignFormats) => {
```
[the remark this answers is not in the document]
erikquinn: Also, we do not need this to be an arrow function, so the standard old school function declaration is preferred in this case:
```javascript
function _buildFlightNumber(callsignFormats) {
    ...
}
```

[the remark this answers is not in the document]
indianbhaji: I did this for consistency though, seeing your other files have arrow functions

src/assets/scripts/client/airline/AirlineModel.js
indianbhaji: done

src/assets/scripts/client/airline/flightNumberBuilder.js
```diff
@@ -0,0 +1,56 @@
...
+    for (let i = 0; i < chosenFormat.length; i++) {
+        switch (chosenFormat[i]) {
+            case '#':
+                flightNumber += _generateNumericValue(i);
+                break;
+            case 'A':
+                flightNumber += _generateAlphaValue();
+                break;
+            default:
+                console.warn(`${this.icao} has an incorrect callsign format, it should only contain 'A' or '#'`);
```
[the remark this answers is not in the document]
indianbhaji: Good recommendation, done!

src/assets/scripts/client/airline/flightNumberBuilder.js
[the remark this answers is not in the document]
indianbhaji: done

src/assets/scripts/client/airline/flightNumberBuilder.js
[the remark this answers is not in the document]
indianbhaji: done

src/assets/scripts/client/airline/flightNumberBuilder.js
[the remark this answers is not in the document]
indianbhaji: done

src/assets/scripts/client/airline/AirlineModel.js
```diff
@@ -207,40 +196,20 @@ export default class AirlineModel extends BaseModel {
...
-            return flightNumber;
-        }
-
-        // TODO: why `this.flightNumberGeneration.length - 3`?  enumerate the magic number.
-        for (let i = 0; i < this.flightNumberGeneration.length - 3; i++) {
-            flightNumber += choose(list);
-        }
-
-        list = 'abcdefghijklmnopqrstuvwxyz';
+        const flightNumber = flightNumberBuilder(this.flightNumberGeneration.callsignFormats);
```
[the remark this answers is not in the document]
indianbhaji: @n8rzz your opinion?

src/assets/scripts/client/airline/flightNumberBuilder.js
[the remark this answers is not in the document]
erikquinn: > I did this for consistency though, seeing your other files have arrow functions for `export`s

Yep, we've been working on slowly changing all the ones back to normal functions which don't _need_ to be arrow functions.

src/assets/scripts/client/airline/flightNumberBuilder.js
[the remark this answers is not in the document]
erikquinn: wish github had a :bananadance: ...

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,60 @@
...
+    for (let i = 0; i < chosenFormat.length; i++) {
+        switch (chosenFormat[i]) {
+            case '#':
+                flightNumber += _generateRandomDigit(i);
+                break;
+            case '@':
+                flightNumber += _generateRandomLetter();
+                break;
+            default:
+                if (i === 0 && chosenFormat[i].charAt(0) === '0') {
```
n8rzz: this condition should be moved above the `switch`.  then this `default` can simply handle the assignment of `flightNumber`

src/assets/scripts/client/airline/AirlineModel.js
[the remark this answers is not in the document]
n8rzz: @erikquinn its in a file at my request.  

My reasoning there is that the generation of flight numbers _could_ contain a fair amount of logic/documentation.  Additionally, the `AirlineModel` as a class shouldn't really care _how_ a flight number is generated, only that one is needed based on rules supplied.

This gives us a way to break out that logic to keep the `AirlineModel` smaller with tighter concerns.

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+ * @function _validateCallsignFormats
+ * @param callsignFormat {Array}
+ * @return validatedFormats {Array}
+*/
+function _validateCallsignFormats(callsignFormats) {
+    const validatedFormats = [];
+
+    for (let i = 0; i < callsignFormats.length; i++) {
+        if (callsignFormats[i].charAt(0) === '0') {
+            console.warn(`Format ${callsignFormats[i]} is invalid as it has a leading zero!`);
```
n8rzz: I'd recommend using a `continue` here and removing the `else`.

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+    if (i === 0) {
+        return choose(NUMERIC.substr(1));
+    }
+
+    return choose(NUMERIC);
+}
+
+/**
+ * This function picks a random letter and returns it.
+ * @function _generateRandomLetter
```
n8rzz: missing empty line above `@function`

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+import { DEFAULT_CALLSIGN_FORMAT } from '../constants/airlineConstants';
+
+const ALPHA = 'abcdefghijklmnopqrstuvwxyz';
+const NUMERIC = '0123456789';
+const defaultCallsignFormats = [DEFAULT_CALLSIGN_FORMAT];
+
+/**
+ * This picks a random number. If it is the first value within the callsign (ie. i === 0), then it picks a
+ * number between 1 and 9. Otherwise, it picks a number between 0 and 9.
+ * @function _generateRandomDigit
```
n8rzz: missing empty line above `@function`

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+ * @return ALPHA {string}
+*/
+function _generateRandomLetter() {
+    return choose(ALPHA);
+}
+
+/**
+ * This function takes a list of callsignFormats, validates them (ie. ensures that no callsignFormat has a leading 0)
+ * then returns a list of validated callsignFormats or the default callsignFormat if there are no valid callsignFormats
+ * @function _validateCallsignFormats
```
n8rzz: missing empty line above `@function`

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+ * @param callsignFormat {Array}
```
n8rzz: `Array` should be lowercase.  

If you can, also add the type expected in the array: `{array<string>}` or `{array<string | number>}`

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+ * @return validatedFormats {Array}
+*/
+function _validateCallsignFormats(callsignFormats) {
+    const validatedFormats = [];
+
+    for (let i = 0; i < callsignFormats.length; i++) {
+        if (callsignFormats[i].charAt(0) === '0') {
```
n8rzz: meant to ask this earlier, what `charAt` and not just `callsignFormats[i]`?

n8rzz: ahh, never mind, I've got it.  makes sense.

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,85 @@
...
+*/
+export function buildFlightNumber(callsignFormats) {
+    let flightNumber = '';
+
+    const validatedFormats = _validateCallsignFormats(callsignFormats);
+    const chosenFormat = choose(validatedFormats);
+
+    for (let i = 0; i < chosenFormat.length; i++) {
+        switch (chosenFormat[i]) {
+            case '#':
```
n8rzz: can you create constants from these two symbols (`#` and `@`)?  I was going to put a note about adding documentation on what they mean, but I think moving them to constants would do enough to help.

src/assets/scripts/client/airline/AirlineModel.js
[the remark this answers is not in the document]
erikquinn: > This gives us a way to break out that logic to keep the AirlineModel smaller with tighter concerns.

> AirlineModel as a class shouldn't really care how a flight number is generated, only that one is needed based on rules supplied.

I would contend that the sole purpose of the `AirlineModel` would be to generate flight numbers for aircraft spawning. And since this class is definitely the only user of that logic, it seems strange, especially for this particular case that we would want to take the sole task of a simple class and move its logic elsewhere.

If this was at your request, I'll definitely go along with it, but I am trying to get a sense of why we would do this here when we haven't in _much_ larger and more complex classes.

src/assets/scripts/client/airline/buildFlightNumber.js
```diff
@@ -0,0 +1,93 @@
...
+ * @function _generateRandomLetter
+ * @return ALPHA {string}
+*/
+function _generateRandomLetter() {
+    return choose(ALPHA);
+}
+
+/**
+ * This function takes a list of callsignFormats, validates them (ie. ensures that no callsignFormat has a leading 0)
+ * then returns a list of validated callsignFormats or the default callsignFormat if there are no valid callsignFormats
```
erikquinn: Not worth changing, but just a note: `This function ...` doesn't really add much to the description, so to keep it concise, just come right out with it: `Validates and returns an array of validated cal...` or `Returns an array of validated cal...` etc. Shorter and clearer means faster and easier to read and understand 😄

PR #869, opened 2018-01-21
enhancement/837

### LINKED ISSUE #837: .hide() and show() within StripViewModel causes big performance hit
`StripViewModel` should be refactored to _not_ use jquery and especially not the `hide` and `show` methods.

n8rzz:
Resolves openscope/openscope#837 

* improves performance by not using `$.show()` and `$.hide()` every frame
* removes arrow functions in `StripViewModel`in favor of properly bound handler methods 
* adds `package-lock.json` to VC

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -293,6 +293,8 @@ export default class AircraftController {
     }
 
     /**
+     * Returns whether the specified aircraft model is in an area where they are controllable
+     *
      * @for AircraftController
      * @method isAircraftVisible
      * @param aircraft {AircraftModel}
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -182,7 +182,7 @@ export default class AircraftController {
      */
     enable() {
         this._eventBus.on(EVENT.ADD_AIRCRAFT, this.addItem);
-        this._eventBus.on(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickhandler);
+        this._eventBus.on(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickHandler);
         this._eventBus.on(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this.onSelectAircraftStrip);
         this._eventBus.on(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
         this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
@@ -198,7 +198,7 @@ export default class AircraftController {
      */
     disable() {
         this._eventBus.off(EVENT.ADD_AIRCRAFT, this.addItem);
-        this._eventBus.off(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickhandler);
+        this._eventBus.off(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickHandler);
         this._eventBus.off(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this._onSelectAircraftStrip);
         this._eventBus.off(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
         this._eventBus.off(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
@@ -300,7 +300,9 @@ export default class AircraftController {
      * @returns {boolean}
      */
     isAircraftVisible(aircraft, factor = 1) {
-        return vlen(aircraft.relativePosition) < AirportController.airport_get().ctr_radius * factor;
+        const visibleDistance = AirportController.airport_get().ctr_radius * factor;
+
+        return aircraft.distance < visibleDistance;
     }
 
     /**
@@ -799,11 +801,11 @@ export default class AircraftController {
      * then trigger another event for the `CanvasController`.
      *
      * @for AircraftController
-     * @method _onStripDoubleClickhandler
+     * @method _onStripDoubleClickHandler
      * @param callsign {string}
      * @private
      */
-    _onStripDoubleClickhandler = (callsign) => {
+    _onStripDoubleClickHandler = (callsign) => {
         const { relativePosition } = this.findAircraftByCallsign(callsign);
         const [x, y] = relativePosition;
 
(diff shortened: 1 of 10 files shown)


src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -303,11 +303,12 @@ export default class AircraftController {
...
      * @for AircraftController
-     * @method aircraft_visible
+     * @method isAircraftVisible
      * @param aircraft {AircraftModel}
      * @param factor {number}
+     * @returns {boolean}
      */
-    aircraft_visible(aircraft, factor = 1) {
+    isAircraftVisible(aircraft, factor = 1) {
         return vlen(aircraft.relativePosition) < AirportController.airport_get().ctr_radius * factor;
```
erikquinn: Like the new method name 😄 

Since we're here, let's change this to:
```js
const visibleDistance = AirportController.airport_get().ctr_radius * factor;

           // km              // km
return aircraft.distance < visibleDistance;
```
Also note that there are some comments and an enum `DEFAULT_CTR_RADIUS_NM` which are incorrectly labeled, as the `#distance` property is actually in kilometers. And we can save here (slightly) by using `#distance` over `vlen(aircraft.relativePosition)` since it is calculated every frame already.

src/assets/scripts/client/aircraft/StripView/StripViewController.js
```diff
@@ -132,14 +132,11 @@ export default class StripViewController {
             // is shown immediately. For arrivals, this does not work so well. We need to `$.detach() the
             // strip and re-add it to the list so it is at the end of the list.
             if (aircraftModel.inside_ctr && !stripViewModel.insideCenter) {
-                stripViewModel.$element.detach();
```
erikquinn: So is the above TODO resolved by removing the `.detach()`?

```js
// TODO: this should be looked at again
// an aircraft strip is created on instantiation, which works for departures where a strip
// is shown immediately. For arrivals, this does not work so well. We need to `$.detach() the
// strip and re-add it to the list so it is at the end of the list.
```

src/assets/scripts/client/aircraft/StripView/StripViewController.js
```diff
@@ -156,7 +153,10 @@ export default class StripViewController {
         const stripViewModel = new StripViewModel(aircraftModel, stripViewCid);
 
         this._collection.addItem(stripViewModel);
-        this._addViewToStripList(stripViewModel);
+
+        if (aircraftModel.isDeparture() || aircraftModel.inside_ctr) {
+            this._addViewToStripList(stripViewModel);
```
erikquinn: Could we get a description on the docblock for `#collection` to differentiate it from wherever we're "adding" this strip view model (apparently to the `.$stripViewList`)?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -599,38 +584,41 @@ export default class StripViewModel extends BaseModel {
...
-     * @method _onDoubleClickHandler
+     * @method _onDoubleClick
      * @param  event {jquery event}
      * @private
      */
-    _onDoubleClickHandler = (event) => {
+    _onDoubleClick(event) {
         event.stopPropagation();
 
         this._eventBus.trigger(EVENT.STRIP_DOUBLE_CLICK, this._callsign);
```
erikquinn: I'm not sure how long it's been like this, but it would seem that the event bus callback for this event does _NOT_ mark the canvas as dirty, so we get this weird behavior where the center point and aircraft are moved, but the canvas stays static on the screen (and thus, is misaligned). Smells like we should be `CanvasController._markDeepRender()`ing somewhere on an `EVENT.STRIP_DOUBLE_CLICK` occurrence.

src/assets/style/landmark/_stripView.less
```diff
@@ -26,15 +26,15 @@
     top: @spacing-2x;
     bottom: @spacing-2x;
     left: @spacing-2x;
+    display: flex;
+    flex-direction: column-reverse;
     overflow-y: scroll;
```
erikquinn: Speaking of this, the scroll direction seems to be reversed in the strip bay. Any idea why that would be, and if we are able to quickly swap it around? It hardly matters, but a few times I've found myself going the wrong way because it's a nonstandard scroll.

n8rzz: yes, it is _'reversed'_ 😉 

In fact, the way it was before was the real reversed version.  In order to achieve the first-item-displays-at-the-bottom, we had to get creative.  First we rotated the list 180 degress which means every strip is upside down.  Then we'd rotate every strip -180 degrees so then every strip was right side up again.

we don't need to do any of that craziness now that `flexbox` is much more supported.  The two changes above are all we need to achieve the same effect.

n8rzz:
@erikquinn If you think that performance is a big change, take a look at the difference in dropped frames.

src/assets/style/landmark/_stripView.less
erikquinn: Stupid alert: I'm just now seeing `flex-direction: column-reverse` literally one line above the one I commented on. I guess that's the one we want. If we can get that thing to scroll the standard way, just like a regular webpage, that would be peachy swell 😄

erikquinn:
😱

src/assets/style/landmark/_stripView.less
n8rzz: That's a little bit harder to do.  Unfortunately, not in the scope of this PR.  Before We refactored the strip view, I believe it did scroll in the same direction it is now.

erikquinn: 👍 No need to go after field mice anyway so disregard

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: done

src/assets/scripts/client/aircraft/StripView/StripViewController.js
n8rzz: done

src/assets/scripts/client/aircraft/StripView/StripViewController.js
n8rzz: yeah, one is a collection the other is a DOM list.  updated.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
n8rzz: This event is only triggering a handler in the `AircraftController` now.  Updated the docs about that.  Also updated `._onCenterPointInView()`, which is the eventual called method, to call `_markDeepRender()`.

Tested locally and didn't see any weirdness.

PR #875, opened 2018-02-06
refactor/529

### LINKED ISSUE #529: Abstract NavigationLibrary to static class and remove FixCollection references
These two classes should both be static and, as such, the `FixCollection` references can be removed from the `NavigationLibrary`

This one is going to be tricky (hence this issue), both classes are used all over the place.  The original intend was to pass references to the `NavigationLibrary` to the classes that needed it, and that worked, but it is getting harder to maintain.

n8rzz:
Resolves #529 

* Moves `NavigationLibrary` to a singleton
* Updates usages and removes passed instances of `NavigationLibrary`

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -4,6 +4,7 @@ import _round from 'lodash/round';
 import AirportController from '../airport/AirportController';
 import EventBus from '../lib/EventBus';
 import GameController from '../game/GameController';
+import NavigationLibrary from '../navigationLibrary/NavigationLibrary';
 import TimeKeeper from '../engine/TimeKeeper';
 import UiController from '../UiController';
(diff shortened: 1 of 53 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -13,7 +14,7 @@ import {
     FLIGHT_CATEGORY
 } from '../constants/aircraftConstants';
 import { EVENT } from '../constants/eventNames';
-import { PROCEDURE_TYPE } from '../constants/routeConstants';
+// import { PROCEDURE_TYPE } from '../constants/routeConstants';
```
erikquinn: can we remove this then?

erikquinn: ah forget it, who cares anyway :shipit: 😆

PR #897, opened 2018-02-23
refactor/874

### LINKED ISSUE #874: Follow-up to #809
Continue work on #809.
- get SID command able to accept `SID ANEKI2A` and have it update departure runway

erikquinn:
Resolves #874.

Continue and clean up following FMS / Route refactor


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1806,32 +1805,29 @@ export default class AircraftModel {
...
-                return this._calculateTargetedAltitudeVnavClimb(hardRestrictedWaypointModel);
+        if (!maximumAltitudeExists && !minimumAltitudeExists) {
+            return this.mcp.altitude;
+        }
 
-            case FLIGHT_PHASE.CRUISE: {
-                if (!this.shouldStartDescent()) {
-                    return;
-                }
+        if (maximumAltitudeExists && minimumAltitudeExists) {
```
n8rzz: just a few lines up you check for: 
```js
!maximumAltitudeExists && !minimumAltitudeExists
```

if that case is true, you return.  so the check here on this line seems repetitive.  Then again, you know how I feel about successive conditional blocks in the same method.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1806,32 +1805,29 @@ export default class AircraftModel {
...
+            if (indexOfMax < indexOfMin) {
+                return this._calculateTargetedAltitudeVnavDescent(nextAltitudeMaximumWaypoint);
             }
 
-            case FLIGHT_PHASE.DESCENT:
-                return this._calculateTargetedAltitudeVnavDescent();
-
-            default:
-                return;
+            return this._calculateTargetedAltitudeVnavClimb(nextAltitudeMinimumWaypoint);
```
n8rzz: I do like seeing this split up into methods, tho.  So thanks for that.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1945,16 +1923,29 @@ export default class AircraftModel {
      * @return {number} speed, in knots
      */
     _calculateTargetedSpeedVnav() {
-        const hardRestrictedWaypointModel = this.fms.nextHardSpeedRestrictedWaypoint;
+        const nextSpeedMaximumWaypoint = this.fms.getNextWaypointWithMaximumSpeedAtOrBelow(this.speed);
+        const nextSpeedMinimumWaypoint = this.fms.getNextWaypointWithMinimumSpeedAtOrAbove(this.speed);
+        const maximumSpeedExists = !_isNil(nextSpeedMaximumWaypoint);
+        const minimumSpeedExists = !_isNil(nextSpeedMinimumWaypoint);
```
n8rzz: these could be re-titled to `hasMaxSpeed` and `hasMinSpeed`, but that's a nit-pick.  just something to call out.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -514,6 +515,70 @@ export default class Fms {
...
+     * This is helpful to see only future waypoints for which a particular altitude is
+     * considered NOT to be compliant. This allows us to focus which waypoints the aircraft
+     * will need to adjust altitude for, even if they are many waypoints in the future.
+     *
+     * @for Fms
+     * @method getNextWaypointWithMaximumAltitudeAtOrBelow
+     * @param altitude {number}
+     * @return {WaypointModel}
+     */
+    getNextWaypointWithMaximumAltitudeAtOrBelow(altitude) {
```
n8rzz: you should use `find` in the name instead of `get`.  we aren't _getting_ here, we're _finding_ 😉

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -514,6 +515,70 @@ export default class Fms {
...
+     * This is helpful to see only future waypoints for which a particular altitude is
+     * considered NOT to be compliant. This allows us to focus which waypoints the aircraft
+     * will need to adjust altitude for, even if they are many waypoints in the future.
+     *
+     * @for Fms
+     * @method getNextWaypointWithMinimumAltitudeAtOrAbove
+     * @param altitude {number}
+     * @return {WaypointModel}
+     */
+    getNextWaypointWithMinimumAltitudeAtOrAbove(altitude) {
```
n8rzz: same here, use `find` in the name instead of `get`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -514,6 +515,70 @@ export default class Fms {
...
+     * This is helpful to see only future waypoints for which a particular speed is
+     * considered NOT to be compliant. This allows us to focus which waypoints the aircraft
+     * will need to adjust speed for, even if they are many waypoints in the future.
+     *
+     * @for Fms
+     * @method getNextWaypointWithMaximumSpeedAtOrBelow
+     * @param speed {number}
+     * @return {WaypointModel}
+     */
+    getNextWaypointWithMaximumSpeedAtOrBelow(speed) {
```
n8rzz: same here, use `find` in the name instead of `get`

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -514,6 +515,70 @@ export default class Fms {
...
+     * This is helpful to see only future waypoints for which a particular speed is
+     * considered NOT to be compliant. This allows us to focus which waypoints the aircraft
+     * will need to adjust speed for, even if they are many waypoints in the future.
+     *
+     * @for Fms
+     * @method getNextWaypointWithMinimumSpeedAtOrAbove
+     * @param speed {number}
+     * @return {WaypointModel}
+     */
+    getNextWaypointWithMinimumSpeedAtOrAbove(speed) {
```
n8rzz: same here, use `find` in the name instead of `get`

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -812,31 +808,55 @@ export default class RouteModel extends BaseModel {
 
     // ------------------------------ PRIVATE ------------------------------
 
-    _appendRouteModelBeginningAtWaypointName(waypointName, routeModel) {
-        const indexOfDivergentLeg = this._findIndexOfLegContainingWaypointName(waypointName);
+    /**
+     * Append a provided route model into this RouteModel
```
n8rzz: I think you mean:
> Append this `RouteModel` with a provided `RouteModel`

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -812,31 +808,55 @@ export default class RouteModel extends BaseModel {
...
+     *
+     * @for RouteModel
+     * @method _appendRouteModelOutOfAirwayLeg
+     * @param divergentWaypointName {string} name of waypoint at which the two routes have continuity
+     * @param routeModel {RouteModel} the RouteModel to be absorbed into this
+     * @return {array} [success of operation, readback]
+     */
     _appendRouteModelOutOfAirwayLeg(divergentWaypointName, routeModel) {
         const indexOfDivergentLeg = this._findIndexOfLegContainingWaypointName(divergentWaypointName);
         const amendedAirwayLeg = this._createAmendedAirwayLegUsingDifferentExitName(
```
n8rzz: a few lines down, I see `routeModel._legCollection`.  if `_legCollection` is intended to be private, it should not be accessed from outside the `routeModel`.  either make it public, or add a public getter for it.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
n8rzz: We should also add some notes in here about how this is going to mutate the existing `legCollection`

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -988,6 +1047,17 @@ export default class RouteModel extends BaseModel {
         return amendedAirwayLeg;
     }
 
+    /**
+     * Create an airway leg based on the provided one, except with the new specified exit
+     *
+     * NOTE: this assumes the exit fix provided has already been verified as valid for this airway
```
n8rzz: where would that verification happen?

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -1094,6 +1210,15 @@ export default class RouteModel extends BaseModel {
...
+     * Return a STAR leg based on the provided leg, except with the new specified exit
+     *
+     * @for RouteModel
+     * @method _createAmendedStarLegUsingDifferentExitName
+     * @param exitFixName {string} name of STAR exit to use for the new STAR leg
+     * @param legIndex {number} index of leg in the #_legCollection
+     * @return {LegModel}
+     */
     _createAmendedStarLegUsingDifferentExitName(exitFixName, legIndex) {
         const divergentLegModel = this._legCollection[legIndex];
```
n8rzz: is there ever a case where `divergentLegModel` would be `undefined`?

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -1094,6 +1210,15 @@ export default class RouteModel extends BaseModel {
...
         const procedureIcao = divergentLegModel.getProcedureIcao();
```
n8rzz: couple lines down, don't we have a helper somewhere that builds the procedure route string for us?  if not, we really should, cause this is used all over the place.

test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -1,21 +1,65 @@
 import ava from 'ava';
```
n8rzz: Glad you combined all these.  It was a pain in the ass to have to update each file.  We had good intentions, it's just become impractical

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: Yeah, I do know what you mean. I don't see a way to resolve it though, but as far as the branching...
- NEITHER minimum nor maximum
- BOTH minimum and maximum
- Minimum BUT NOT maximum
- Maximum BUT NOT minimum
...all of these require distinctly different actions, which are contained within their own methods which are called from here. Open to ideas to rearrange, but I don't see anything myself.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: eh. why not, right!?!?

done :smile:

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: done

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: Changed to

```
/**
 * Append a provided route model onto the end of this RouteModel
 *
 * This method only serves to call the method that contains the appropriate logic
 * based on the type of leg in which the divergent waypoint resides, since this
 * heavily weighs in to how the merging of the routes should be done. Note that
 * the #_legCollection will be mutated in this process.
```

done.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: Apparently I pulled this slimy trick more times than I realized. Tracked them all down, and replaced with the new (tested) getter, `RouteModel.legCollection`.

done.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: Added more detail:
```
/**
 * Create an airway leg based on the provided one, except with the new specified entry
 *
 * NOTE: this assumes the entry fix provided has already been verified as valid for this airway
 *
 * We know that `_createAmendedConvergentLeg()` and `_prependRouteModelIntoAirwayLeg()` both
 * are called only in situations where a requested route amendment ends at a fix that was already
 * included in the #_waypointCollection of an airway leg of the previous route. If this method is
 * called by either of them, we can be confident that the `entryFixName` is on the airway.
```

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: No, we will only be able to get here if the routes have continuity, which means _some_ leg is the one where the routes diverge. So we should be well-covered on that front.

test/aircraft/Pilot/Pilot.spec.js
erikquinn: yip! :+1:

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: Added (with test) and used here and elsewhere.

erikquinn:
Just stopping here as this WIP is not really "in progress". Should just get what's already done in, and other things can be addressed at a later time. Re-requesting review.

erikquinn:
Got some test stuff to fix.

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: fair enough.  sometimes the logic is gross because the logic _is_ gross

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -480,6 +481,70 @@ export default class Fms {
         return this._routeModel.absorbRouteModel(nextRouteModel);
     }
 
+    /**
+    * Return the next waypoint having an #altitudeMaximum equal to or less than the specified value
```
n8rzz: the spacing on all the docblocks in this file are off by one missing space
```
// do not
/**
*
*/

// do
/**
 *
 */
```

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -592,6 +600,86 @@ ava('.hasWaypointName() returns true when the specified waypoint is in the route
...
 });
 
+ava('.isRunwayModelValidForSid() returns false if passed runway is not an instance of RunwayModel', (t) => {
+    const model = new RouteModel(complexRouteStringMock);
+    const airportModel = createAirportModelFixture();
+    const runwayModel = airportModel.getRunway('00');
+    const expectedResult = false;
+    const result = model.isRunwayModelValidForSid(runwayModel);
+
+    t.true(result === expectedResult);
```
n8rzz: I am a fan of verbosity, but this is actually redundant and a little more confusing.  You're using `t.true()` with strict equals to verify a value is `false`.  

This matcher could be written:
```js
t.false(result);
// or
t.false(model.isRunwayModelValidForSid(runwayModel));
```

I'd be ok with either and feel free to apply the chosen one to all the other instances like this within this file

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: gah! done.

src/assets/scripts/client/utilities/navigationUtilities.js
```diff
@@ -0,0 +1,6 @@
+import { PROCEDURE_OR_AIRWAY_SEGMENT_DIVIDER } from '../constants/routeConstants';
+
+export function assembleProceduralRouteString(entryFix, procedureOrAirway, exitFix) {
```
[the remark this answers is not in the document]
erikquinn: done

test/aircraft/AircraftModel.spec.js
```diff
@@ -68,7 +68,8 @@ ava('.getViewModel() includes an altitude that has not been rounded to the neare
     t.true(result === 77.77123456700001);
 });
 
-ava('.updateTarget() causes arrivals to descend when the STAR includes only AT or ABOVE altitude restrictions', (t) => {
+// FIXME
+ava.skip('.updateTarget() causes arrivals to descend when the STAR includes only AT or ABOVE altitude restrictions', (t) => {
```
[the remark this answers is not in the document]
erikquinn: nope. thing is working as expected, updated expected result to pass.

done

test/aircraft/FlightManagementSystem/Fms.spec.js
```diff
@@ -386,6 +386,29 @@ ava('._initializePositionInRouteToBeginAtFixName() calls .skipToWaypointName() w
...
+    const routeStringToApply = 'BIKKR.PGS';
+    const absorbRouteModelSpy = sinon.spy(fms._routeModel, 'absorbRouteModel');
+    const expectedResult = [false, 'requested route of "BIKKR.PGS" is invalid'];
+    const result = fms.applyPartialRouteAmendment(routeStringToApply);
+
+    t.deepEqual(result, expectedResult);
+    t.true(absorbRouteModelSpy.notCalled);
+});
+
+ava('.applyPartialRouteAmendment() calls fms._routeModel.absorbRouteModel() when provided routestring is properly formatted', (t) => {
```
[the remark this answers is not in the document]
erikquinn: `calls #_routeModel.absorbRouteModel() when`

done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: yeah i guess that is a bit silly. what can i say... i'm a sucker for patterns

fixed!

erikquinn: also fixed for instances of 
```
const expectedResult = true;

t.true(result === expectedResult);
```
in favor of
```
const result = kljdhflgkjdhlfg

t.true(result);
```

PR #901, opened 2018-02-24
bugfix/899

### LINKED ISSUE #899: Route command doesn't work properly anymore
Browser: Chrome 64.0.3282.167

Airport: LSZH [https://openscope-dev-pr-839.herokuapp.com/](https://openscope-dev-pr-839.herokuapp.com/)

Approximate play time before issue: -

#### Issue Description
Incorrect routing

#### Steps to reproduce
f.e. this happens with aircraft with the following route: "LSZH28.DEGES2W.DEGES.Z1.SUDEN"
- enter "XXXXX route DEGES..MINGA" short after departure

Then you can see a really weird routing:
Screenshot
Screenshot2

Note that _RWY28KLO03DME _KLO253005 ZH552 _KLO232005 KLO MOMOL KOLUL ZH504 ZH525 are the fixes of the DEGES2W SID.
Actually the routing should be "DEGES..MINGA.Z1.SUDEN" 

- It looks like the aircraft is just go flying the SID normally. "DEGES2W DEGES"
- Then it turns westbound again, inbound "_RWY28KLO03DME" to fly the SID a second time (as mentioned above).
Screenshot3
- After that there's a small part which contains the correct routing "DEGES..MINGA.Z1.SUDEN"
- Last part just contains a part of the original routing "DEGES.Z1.SUDEN"

---

I think it could have something to do with the fact that both fixes DEGES and MINGA are already in the original route of the aircraft because when I'm using a fix that isn't in the route already it works.
f.e.: "route ZUE..MINGA"

erikquinn:
Resolves #899.

Fix improper route merge when waypoint at start/end of leg


src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -1020,10 +1020,18 @@ export default class RouteModel extends BaseModel {
         }
 
         if (convergentLegModel.isSidLeg) {
+            if (_first(convergentLegModel.waypoints).name === endWaypointName) {
```
n8rzz: you should pull this out to a `const` so that we can reuse it:
```js
const firstWaypointName = _first(convergentLegModel.waypoints).name;
```

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -1046,10 +1054,18 @@ export default class RouteModel extends BaseModel {
         }
 
         if (divergentLegModel.isSidLeg) {
+            if (_last(divergentLegModel.waypoints).name === startWaypointName) {
```
n8rzz: same here, pull this out to a `const` at the top of the method

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -1255,25 +1271,15 @@ export default class RouteModel extends BaseModel {
...
-            ...endingLegCollection
-        ];
-
         const indexOfDivergentLegModel = this._findIndexOfLegContainingWaypointName(startWaypointName);
         const indexOfConvergentLegModel = this._findIndexOfLegContainingWaypointName(endWaypointName);
         const amendedDivergentLegModels = this._createAmendedDivergentLeg(indexOfDivergentLegModel, startWaypointName);
         const amendedConvergentLegModels = this._createAmendedConvergentLeg(indexOfConvergentLegModel, endWaypointName);
+        const endingLegCollection = this._legCollection.splice(indexOfConvergentLegModel + 1);
+
+        this._legCollection.splice(indexOfDivergentLegModel);
```
n8rzz: I would suggest first creating a copy of the `_legCollection` at the top of this method:
```js
const legCollection = this._legCollection.slice(0);
```

Then you can mutate the `nextLegCollection` instead of directly mutating the `#_legCollection` instance.  This is a subtle change but it means that you only ever reassign `#_legCollection` with a finished update instead of part of the finished product.  Doing it this way makes the logic a little clearer too.

n8rzz: along those same lines, there should be a method on the `LegCollection` (or `BaseCollection`) class that returns a copy.  This logic should _never_ be done inline with a collection class.

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -953,6 +953,16 @@ ava('._createAmendedConvergentLeg() returns an empty array when leg is a direct
     t.deepEqual(result, []);
 });
 
+ava('._createAmendedConvergentLeg() returns the leg unmodified when convergent waypoint is the SID leg\'s first fix', (t) => {
+    const model = new RouteModel('KLAS07R.BOACH6.HEC');
+    const waypointName = 'JESJI';
+    const legIndex = 0;
+    const expectedResult = [model._legCollection[legIndex]];
```
n8rzz: You should specify the expected result inline here instead of relying on the model to produce it for you.  Doing it this way can lead to hard-to-track side affects and also makes the test easier to read, you can see exactly what is expected without having to inspect the result.

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -965,6 +975,16 @@ ava('._createAmendedConvergentLeg() calls ._createLegsFromSidWaypointsAfterWaypo
     t.deepEqual(result, expectedResult);
 });
 
+ava('._createAmendedConvergentLeg() returns the leg unmodified when convergent waypoint is the STAR leg\'s first fix', (t) => {
+    const model = new RouteModel('DVC.GRNPA1.KLAS07R');
+    const waypointName = 'DVC';
+    const legIndex = 0;
+    const expectedResult = [model._legCollection[legIndex]];
```
n8rzz: You should specify the expected result inline here instead of relying on the model to produce it for you.

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -1020,6 +1040,16 @@ ava('._createAmendedDivergentLeg() returns an empty array when leg is a direct l
     t.deepEqual(result, []);
 });
 
+ava('._createAmendedDivergentLeg() returns the leg unmodified when divergent waypoint is the SID leg\'s last fix', (t) => {
+    const model = new RouteModel('KLAS07R.BOACH6.HEC');
+    const waypointName = 'HEC';
+    const legIndex = 0;
+    const expectedResult = [model._legCollection[legIndex]];
```
n8rzz: You should specify the expected result inline here instead of relying on the model to produce it for you.

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -1032,6 +1062,16 @@ ava('._createAmendedDivergentLeg() calls ._createLegsFromSidWaypointsBeforeWaypo
     t.deepEqual(result, expectedResult);
 });
 
+ava('._createAmendedDivergentLeg() returns the leg unmodified when divergent waypoint is the STAR leg\'s last fix', (t) => {
+    const model = new RouteModel('BCE.GRNPA1.KLAS07R');
+    const waypointName = 'LEMNZ';
+    const legIndex = 0;
+    const expectedResult = [model._legCollection[legIndex]];
```
n8rzz: You should specify the expected result inline here instead of relying on the model to produce it for you.

test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -1154,7 +1194,19 @@ ava('._getPastAndPresentLegModels() returns #_previousLegCollection concatenated
...
 });
 
-ava.todo('._overwriteRouteBetweenWaypointNames()');
+ava('._overwriteRouteBetweenWaypointNames() adjusts divergent/convergent legs and replaces middle content correctly', (t) => {
+    const primaryModel = new RouteModel('KLAS07R.TRALR6.BCE.J11.DRK');
+    const otherModel = new RouteModel('BCE..NAVHO');
+    const expectedResult = [true, { log: 'rerouting to: KLAS07R TRALR6 BCE NAVHO J11 DRK', say: 'rerouting as requested' }];
+    const result = primaryModel._overwriteRouteBetweenWaypointNames('BCE', 'NAVHO', otherModel);
+
+    t.deepEqual(result, expectedResult);
```
n8rzz: this assertion belongs in it's own test, break this up into two separate test cases; one for the response and one for the updated `_legCollection`

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
erikquinn: first comment done, and for the second, the `_legCollection` is actually just a simple array currently, so can't do that at the moment. But added #925 to add a copy method to the base classes.

done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: Changed to explicitly specify the desired waypoints.

done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: done

test/aircraft/FlightManagementSystem/RouteModel.spec.js
erikquinn: Is it genuinely important to separate this? I definitely don't see this as a problem personally, and would find it redundant to repeat the test verbatim checking something different each time just to get more specificity from the failure of the test (I assume that's the reason to divide it up), and not worth the verbosity, especially since the test's failure message would point you to the issue. I agree in concept that we should want the tests to be pointed and specific, but I would implement that as being generally one test per logic branch, examining all resulting changes/return values/etc for each logical branch. That seems like what we would want, and that's what this test does here by examining the workings of only this private method.

PR #917, opened 2018-03-05
feature/884

### LINKED ISSUE #884: Add non-US registry tail number "airlines"
Currently, we have a `n.json` airline for all the american tail numbers, such as `N117KS`. But we do not have the ability to create any other non-airline tail numbers, which makes GA at european airports a little silly. I mean, when's the last time we saw a whole bunch of C172s with a US registry flying around Europe? Probably not recently!

Add `g.json` (UK), `ei.json` (Ireland), `d.json` (Germany), etc, so we can add some local flavor to airports that don't start with a K. 😃

Marcel510:
Resolves #884.

Adds non-US-registry general aviation airlines


Marcel510:
I wasn't sure if a changelog entry is needed since it doesn't really affect the sim itself only some files which aren't used yet but I'll add it.

assets/airlines/2.json
```diff
@@ -0,0 +1,54 @@
...
+            "####"
+        ]
+    },
+    "fleets": {
+        "default": [
+            ["BE36", 5],
+            ["C172", 5],
+            ["C182", 5],
+            ["C208", 1],
+            ["C208", 2],
```
erikquinn: C208, C510, C550, C750 are all duplicated in the default fleets of all of these airlines?

indianbhaji: Yeah, they're duplicated but I thought due to the number of files in this PR it wouldn't be necessary to go through all of them again

Marcel510: Oh yes, I messed that up, no problem I'll fix it

PR #941, opened 2018-04-26
refactor/927

### LINKED ISSUE #927: KDCA SID issues, points deduction
Browser: Google chrome

Airport: KDCA

Approximate play time before issue: any

#### Issue Description
Any of the SIDs at DCA that go over R-6608(a, b, and c) don't gain altitude before flying through the No-Fly zone. 

#### Steps to reproduce
- clear a plane for any of the SIDs that go over R-6608 i.e. CLTCH1.MAULS
- airplane goes through the R-6608 after leaving 7000, making me unable to redirect it.

Fechulo:
Resolves #927 
Removes PSA Airlines (JIA) from spawnPatterns flying over restricted areas 6608, due to their CRJs not reaching necessary altitude


erikquinn:
Lots of JIAs at DCA. Will need to figure this out before merge.   Hmmm

Fechulo:
I know it's not the best solution, but I'ts a quick fix. I'm not sure if the climb rate for the CRJs is realistic, I think it could be higher. If it isn't, then I guess they would file for the NATNL7 departure instead, a vectoring departure where there are no altitude restrictions.

indianbhaji:
I believe it’d be better to change the climb rates of those CRJs;   https://contentzone.eurocontrol.int/apm/#/apm/Home/BasicList?wsGroupDropDownList=12&wsNameTextField=&wsICAOTextField=Crj
  
  
>

Fechulo:
Yeah that makes more sense, I won't be able to do it until later tonight though, but should still be enough time to get it in this sprint

Fechulo:
@indianbhaji @erikquinn Should be fixed now! 😄

PR #954, opened 2018-05-02
hotfix/953

### LINKED ISSUE #953: Arrivals don't always descend to controllable altitudes
Browser: Chrome
Airport: KDCA

#### Issue Description
Aircraft do not always descend to within the vertical boundaries of our airspace. In cases where there is no STAR to descend via, the aircraft will descend to the top of the airspace. However, when there IS a star with descent information, but it doesn't force them down into _our_ airspace, they are stuck above where we can control. They should always be forced down to an altitude where we can do something with them, whether descending via STAR, or whether descending manually.

#### Steps to reproduce
- At KDCA, observe the arrivals from the south level off at 90 when our airspace tops out at 70.

erikquinn:
## https://openscope-dev-pr-954.herokuapp.com/

Resolves #953

Fix aircraft not descending into airspace when airspace ceiling is below STAR bottom altitude

Most notable at KDCA (see the southerly arrival stream)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 6.1.2 (May 2, 2018)
+### Hotfix
+- [#953](https://github.com/openscope/openscope/issues/953) - Fix aircraft not descending into airspace when airspace ceiling is below STAR bottom altitude
+
+
 # 6.1.1 (May 1, 2018)
 ### Hotfix
 - [#950](https://github.com/openscope/openscope/issues/950) - Fix left turn command `t l ###`
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "openscope",
-  "version": "6.1.1",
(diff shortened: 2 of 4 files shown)


f18bkr:
Seems this issue has been fixed as far as I can tell, but I had an issue the the MD82 on the southerly arrival route. It seems it couldn't descend in a timely manner to receive clearance to land, and only came below FL70 well north of the airport. 

Also, an the MD8X/717/CRJ/ERJ, it looks likes the descend rates aren't high enough to allow the aircraft to vectored properly as they only pass through FL70 close to the airport. Maybe these rates need to be adjusted?

Thanks

erikquinn:
Hi @f18bkr! Yes, those issues I also noticed while playing around with KDCA recently. All of those are caused by the aircraft not properly planning their descents due to some logic in the code that causes them to make incorrect decisions about when to start descending. Those bugs are being addressed in #959.

PR #973, opened 2018-05-21
enhancement/972

### LINKED ISSUE #972: Update fleets
#Browser: Chrome

Airport: KSTL

Approximate play time before issue: Any

#### Issue Description
SWA aircraft are still spawning with B733 and B735
Check airlines to see what fleets need updating

JPmAn24:
Resoves #972.

Update airline fleets to newest data


JPmAn24:
Note - need to remove a319 fleet from etd after removing it from airports if they are using it

assets/airlines/emb.json
```diff
@@ -11,10 +11,8 @@
     },
     "fleets": {
         "default": [
-            ["E50P", 5],
```
indianbhaji: Why have the `E50P`, `E55P`, `E545` aircraft types been removed? They should be readded.

assets/airlines/lan.json
```diff
@@ -2,30 +2,31 @@
     "icao": "lan",
     "name": "LATAM Airlines Chile",
     "callsign": {
-        "name": "LAN Chile",
+        "name": "LAN",
```
erikquinn: The callsign _is_ `Lan Chile`

Marcel510: Every source says something different... German wikipedia: LAN, English wikipedia: LAN Chile, planespotters.net: LAN. But youre an ATC so you probably know best, thanks

Fechulo: The problem is that the speech synthesiser would say that as "L - A - N Chile". While the callsign is technically `LAN Chile`, `Lan Chile` will get us the desired result

assets/airlines/klc.json
```diff
@@ -2,14 +2,15 @@
     "icao": "klc",
     "name": "KLM Cityhopper",
     "callsign": {
-        "name": "KLM",
+        "name": "City",
```
indianbhaji: Change this back to KLM. They never use the "City" callsign.

Marcel510: They never use it because they never fly with the `KLC` icao code, they are always operating for the normal `KLM`. This airline should't be used actually, it's more realistic to just use the "embraers" fleet of the normal KLM airline file.

indianbhaji: Yes, so the option is either change it, keep it, or delete it

Marcel510:
Note: EDDH needs to be updated to remove nly.

JPmAn24:
Got the EDDH update

JPmAn24:
I also need to get the A319 fleet out of OTHH

assets/airlines/sht.json
```diff
@@ -1,19 +0,0 @@
-{
```
indianbhaji: This file needs to be reinstated. What’s the reason for deletion? SHT is a commonly used callsign for British Airways domestic flights.

JPmAn24: Oh. Cancel that commit then.

erikquinn:
Updated with develop. @indianbhaji is that the only issue with it?

indianbhaji:
@erikquinn Cannot deny nor confirm; I’m laying in bed so can’t properly check until tomorrow
  

  
  

  
  
>

PR #977, opened 2018-05-27
enhancement/965

### LINKED ISSUE #965: Unrealistic aircraft speeds
#### Issue Description
The cruise/max speed of many aircrafts are true airspeeds eventhough they should be indicated airspeeds.
f.e. the current max speed of the A320 is 487kts eventhough a normal cruise indicated speed of an A320 is about 250-270kts. 487kts indicated speed at cruising level would be more than mach 1.00.

erikquinn:
Resolves #965.
__Blocked by #897.__

Add required data and math to get aircraft to fly appropriate speeds at any altitude.


Marcel510:
Maybe when this is implemented we should get rid of defining a speed for generated inbound traffic and just use the appropiate speed that is going to be defined in the files then. What do you think?

Issue #111, opened 2017-01-18
Draw STARs on canvas


indianbhaji:
@willfrd mentioned in zlsa/atc#781    that STARs aren't drawn.

@n8rzz said this is something that he'd like to work on, so here's the issue!

It would be nice to see the STAR lines just like how it is already done with the SID lines.

Related: #1093.

willfrd:
Just wondering, with this new repository the original zlsa/atc repository will be completely closed? Future projects will have to be uploaded into here? Do we know how much the "moving" will take? Thanks

erikquinn:
@willfrd Correct, all future work will be done here.

PR #980, opened 2018-05-29
enhancement/444

### LINKED ISSUE #444: Add video map toggle option
I can appreciate the realism that video maps provide to experienced players and actual controllers, but from a layman's perspective, they are really mystifying and detract from/clutter the experience.  It would be nice to have an option of hiding them like the other overlay components.

n8rzz:
Resolves openscope/openscope#444 

* Adds button to toggle video map display
* updates method names in `UiController` to use `on` prefix when dealing with event triggers
* also wires up logic and image for future draw stars functionality (#111)


n8rzz:
@erikquinn I was going to say it was tagged as an enhancement hence the branch name.  But, it's not 😦 

At any rate, changing the branch name will F up my commit history and I'd rather get this in than worry about the branch name and where it lives in the changelog.

Issue #1025, opened 2018-07-08
Empty space in airport selection dialog


erikquinn:
Browser: Chrome

#### Issue Description
Airport selection dialog has a big empty space at the bottom. Should be adjusted so the size of `#airport-switch` will be the size of its contents, keeping its current `max-height` values.

![image](https://user-images.githubusercontent.com/5103735/42422247-73e28afc-82b0-11e8-89b0-8c4b47ad5f68.png)

PR #1026, opened 2018-07-08
enhancement/1000 - Increase max width of airport selection dialog

### LINKED ISSUE #1000: Ugly newline within the airport selection menu
When there's not enough space for all airports and the scrollbar gets activated, the name of CYOW is too long for the line and therefore a ugly newline gets added which doesn't look really professional in my opinion. I think we should either shorten the name of CYOW or make the menu a bit wider.

erikquinn:
Resolves #1000.

Increase max width of airport selection dialog to permit longer airport names


src/assets/style/module/airport/_airport.less
```diff
@@ -1,11 +1,11 @@
 /* TODO: this rule should be moved to a class instead of ID */
 /* TODO: this rule should be broken up into several other rules */
 #airport-switch {
+    width: fit-content;
```
n8rzz: this is unnecessary and unsupported (https://caniuse.com/#search=fit-content)

erikquinn: Without it, it doesn't achieve the desired effect, so I'll wait until we can find another way to do it. I stink at CSS, and am surprised I could even get this to work at all :wink:

The idea is for the width to be only as wide as the content inside needs it to be, and allows it to expand its width up to 600px.

n8rzz: yeah, it does that already even without the `fit-content`.  what, exactly, is the desired effect?

also, how big is your screen?

erikquinn: The desired effect is for `#airport-switch` to be no wider than its content necessitates, and for it to expand as needed to fit longer airport names, with a maximum width of 600px. Currently, it expands and tops out at 600px, but does not shrink beyond a certain width, even if the airport names are very short. (Note that the bottom note text can be ignored, as it will be removed soon)

image

erikquinn: And I was originally on my big monitor, 2560x1440, now on the laptop (where the screenshot was taken), 1920x1080.

erikquinn: @n8rzz Open to other ways to resolve this, or even just rolling that into another issue, #1025 (basically the same problem in the other axis), and having this PR discard these CSS changes, and simply increase the max-width.

PR #1036, opened 2018-07-26
feature/1032 - Add range ring options and toggle button

### LINKED ISSUE #1032: Add range ring options and toggle button
Add range ring options that allow the user to specify the default location and radius of range rings for a given airport, or disable them by default. Users should be able to turn the rings on/off, regardless of the default, and change the radius in increments of 1-2-5-10-15-20 nautical miles.

Inspired by #937.

erikquinn:
Resolves #1032.

Add range ring options and toggle button


JPmAn24:
@erikquinn Could I take this over?

JPmAn24:
@felixscheffer Thanks for taking over for the time being.

felixscheffer:
Needs review. I tried to maintain backward-compatibility for the airport file.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -358,8 +356,24 @@ export default class AirportModel {
...
         this.ctr_ceiling = _get(data, 'ctr_ceiling', DEFAULT_CTR_CEILING_FT);
         this.initial_alt = _get(data, 'initial_alt', DEFAULT_INITIAL_ALTITUDE_FT);
-        this.rr_radius_nm = _get(data, 'rr_radius_nm');
-        this.rr_center = _get(data, 'rr_center');
+        this.rangeRings = _get(data, 'rangeRings');
+
+        if (!this.rangeRings) {
+            // fallback code, so we don't break existing airport files
+            const fallbackRadius = _get(data, 'rr_radius_nm');
+            const fallbackCenter = _get(data, 'rr_center');
```
erikquinn: Remove all this, just update all the airports in this PR-- it's a small change so it shouldn't be too bad.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -151,5 +152,47 @@ export const GAME_OPTION_VALUES = [
                 value: 'no'
             }
         ]
+    },
+    {
+        name: GAME_OPTION_NAMES.RANGE_RINGS,
+        defaultValue: 'default',
+        description: 'Range rings',
+        help: '@erikquinn your help text here',
```
erikquinn: `Radius of range rings, in nautical miles`

src/assets/scripts/client/UiController.js
```diff
@@ -309,13 +309,12 @@ class UiController {
         const parsedSelectedOption = !_isNaN(parseFloat(selectedOption))
             ? parseFloat(selectedOption)
             : selectedOption;
-        let optionSelectTempalate = `<option value="${optionData.value}">${optionData.displayLabel}</option>`;
 
         if (optionData.value === parsedSelectedOption) {
-            optionSelectTempalate = `<option value="${optionData.value}" selected>${optionData.displayLabel}</option>`;
```
n8rzz: hmm, why the change to a return here?

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -27,6 +27,11 @@ import { STORAGE_KEY } from '../constants/storageKeys';
 const DEFAULT_CTR_RADIUS_KM = 80;
 const DEFAULT_CTR_CEILING_FT = 10000;
 const DEFAULT_INITIAL_ALTITUDE_FT = 5000;
+const DEFAULT_RANGE_RINGS = {
+    enabled: false,
```
n8rzz: if this is a constant, the child keys should be `UPPERCASE` as well

src/assets/scripts/client/airport/AirportModel.js
n8rzz: agreed.  this entire block should go away

felixscheffer: No offense, but sometimes you guys have a very demanding attitude. I am not going to change 86 airport files everytime there is a change to the file format. If you think it's the right thing to do, feel free to go ahead and make the change, but I won't.

src/assets/scripts/client/constants/gameOptionConstants.js
felixscheffer: done

src/assets/scripts/client/UiController.js
felixscheffer: So we don't do the string templating twice.

src/assets/scripts/client/airport/AirportModel.js
n8rzz: No offense taken 😉 , you're exactly right we are demanding.  Please don't take that personally, we are demanding of everybody including ourselves.

Our goal is simply to maintain a maintainable app.  as such, every change matters every time.

As far as this part goes, if you don't want to make the airport file changes yourself, that's ok.  You don't need to.  That leaves us with a few options here:

##### we leave this as-is (cause it *is* backwards compatible)
- add a comment about it needing to be removed later
- open a new issue to update the existing airport files

#### we ask for help from someone else to finish up the airport file updates
- this would include updating the airport documentations if that hasn't happened already

Never be afraid to push back!  We will ask questions and suggest changes (or possibly even make demands), but never ever be afraid to say no or suggest a different way.  This is a group project, not just the work of a few people.  Everybody participates and everybody has a voice, don't be afraid to use it!

erikquinn: Yeah, sometimes we are a bit blunt with feedback. I think @n8rzz and I exhaust each other sometimes with our requests for changes on _everything_, and then just lengthen the to-do list on the next review. As Nate said, we really just want to get things right. We've been on a cleanup mission for a long time, and things have gotten a lot better, so that may be the reason for excessive pickiness and/or OCD tendencies.

I'm happy to go through and make the changes to the airport files myself. We're all volunteers, so if there's something you don't want to do, that's always going to be 100% cool with us. Whatever you didn't want to do, you can always just wrap up and leave a comment on what's left to do and mark is as orphaned so somebody else can finish up the gross parts. :+1:

Fechulo: I'm happy to help, what is it that needs updating?

erikquinn: Just the format of the range rings is different

felixscheffer: 1. This code is not meant to be a permanent solution. I thought that was obvious but maybe it wasn't. The idea was to provide a temporary solution until we have changed all the airport files over time. I probably should have added a comment to make that more clear.
2. I am trying not to take things personally, but honestly, sometimes it's hard not to. Just because I spent a LOT of time on this project lately, doesn't mean I have time to waste on useless polishing (and that's what it is IMHO). 
3. I am okay with being demanding. This project is being managed better than most 7-figures projects I have worked on over past ten years and that's great. But man power is a very limited resource around here. At some point you have to make compromises and spend your resources wisely. Otherwise this project won't go anywhere anytime soon.

n8rzz: I'm going to skip straight to #3

First, thank you.  We've tried really hard to make it as well run as possible with volunteer effort.  I'm curious, how would you suggest we solve the second point there because you're right resources are limited.

The counterpoint there is that is one reason why this is what you call a well run project, we take our time to do things 'right' the first time.  We don't make too many compromises on quality because we don't have to.  We're our own product owners.

Either way I think we need to strike a balance between intricate feedback and broad PR acceptance.  At present, we may be too far to the intricate feedback side.

src/assets/scripts/client/airport/AirportModel.js
erikquinn: It's actually a default object which is directly assigned to a property during initialization, so the keys have to match the casing of the object. Is this a bad practice, or okay to do it like this?

felixscheffer: Now that you mention it, I think we should create a copy of the object during initialization, so we don't (accidentally) modify the constant.

erikquinn:
@JPmAn24 Did you get any airports updated to the new range ring format? If not, I can maybe pick it up. Would like to try to get this one out this month.

felixscheffer:
👍

erikquinn:
Awaiting #1285.

erikquinn:
#1285 merged. Unblocking.

PR #1038, opened 2018-07-26
bugfix/1034 - fix waypoint time-to-turn calculations

### LINKED ISSUE #935: Aircrafts are skipping a fix
Airport: OTHH (#857)

#### Issue Description
Screenshot
The route of the aircraft should be ELEDA-MEBTI-BOSUP-... 
But after ELEDA the aircraft skipping MEBTI and turns inbound BOSUP.

#### Steps to reproduce
That's applicable to the following SIDs: BATHA1E, BUNDU1E, NAMLA1E, SALWA1E
(At the current version of the airport MEBTI defined as a fly-over fix in order to prevent this)

### LINKED ISSUE #1034: Waypoint time-to-turn calculations are slightly off
Discovered during the work on #937, the calculations (which could never be figured out because they're old and without labels/explanations) turned out to actually be slightly off. They should be reworked and better documented so we can clearly see how the calculations work.

erikquinn:
Resolves #1034
Resolves #935

Fix waypoint time-to-turn calculations.

The errors were fairly minimal, but became quite noticeable at very high ground speeds from the en route demo. They were constantly undershooting the turn, whereas with this they will cut the corner properly and end up right on the targeted track.

### DIFF
--- a/src/assets/scripts/client/math/flightMath.js
+++ b/src/assets/scripts/client/math/flightMath.js
@@ -25,27 +25,34 @@ import {
 import { PHYSICS_CONSTANTS } from '../constants/globalConstants';
 
 /**
+ * Calculate the radius of turn of the aircraft, given its groundspeed
+ * 
+ * Reference:
+ * http://www.flightlearnings.com/2009/08/26/radius-of-turn/
+ * 
+ * Using constant of 68416 (to yield nm) in lieu of 11.26 (yields ft)
+ * 
  * @function calcTurnRadius
- * @param speed {number} currentSpeed of an aircraft
- * @param bankAngle {number} bank angle of an aircraft
- * @return {number}
+ * @param speed {number} aircraft groundspeed, in knots
+ * @param bankAngle {number} bank angle to use, in radians
+ * @return {number} radius of turn, in nautical miles
  */
-export const calcTurnRadius = (speed, bankAngle) => {
-    return (speed * speed) / (PHYSICS_CONSTANTS.GRAVITATIONAL_MAGNITUDE * tan(bankAngle));
-};
+export function calcTurnRadius(speed, bankAngle) {
+    return (speed * speed) / (68416 * Math.tan(bankAngle));
+}
 
 /**
- * @function calcTurnInitiationDistance
+ * @function calcTurnInitiationDistanceNm
  * @param speed {number}            currentSpeed of an aircraft
  * @param bankAngle {number}        bank angle of an aircraft
- * @param courseChange {number}
+ * @param courseChange {number}     angular difference, in radians
  * @return {number}
  */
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/math/flightMath.js
```diff
@@ -25,27 +25,34 @@ import {
...
- * @return {number}
+ * @param speed {number} aircraft groundspeed, in knots
+ * @param bankAngle {number} bank angle to use, in radians
+ * @return {number} radius of turn, in nautical miles
  */
-export const calcTurnRadius = (speed, bankAngle) => {
-    return (speed * speed) / (PHYSICS_CONSTANTS.GRAVITATIONAL_MAGNITUDE * tan(bankAngle));
-};
+export function calcTurnRadius(speed, bankAngle) {
+    return (speed * speed) / (68416 * Math.tan(bankAngle));
```
n8rzz: Does this 68416 have a name we can use for a constant?

src/assets/scripts/client/math/flightMath.js
```diff
@@ -25,27 +25,34 @@ import {
...
  * @return {number}
  */
-export const calcTurnInitiationDistance = (speed, bankAngle, courseChange) => {
-    const turnRadius = calcTurnRadius(speed, bankAngle);
+export function calcTurnInitiationDistanceNm(speed, bankAngle, courseChange) {
+    const turnRadiusNm = calcTurnRadius(speed, bankAngle);
 
-    return turnRadius * tan(courseChange / 2) + speed;
-};
+    return turnRadiusNm * (courseChange / 2) / (Math.PI / 2);
```
n8rzz: Break out these `/ 2` items to variables

src/assets/scripts/client/math/flightMath.js
erikquinn: it's kinda just a constant. we can call it `const conversionFactor = 68416;` if you like.

erikquinn: done

src/assets/scripts/client/math/flightMath.js
erikquinn: done

felixscheffer:
Update:

1.  calculate the turn initiation distances based on turnRadius * tan(angle*0.5)
2.  calculate the turn radius based on turn rate (like we do, when we actually turn the aircraft) instead of using bank angle.
3.  wind is constantly pushing the aircraft off the track. I think there is a ticket for this issue but if there is none, we should definitely create one.
4. calculate the turn initiation distances based on the aircraft's target heading instead of its actual heading to prevent the aircraft from skipping fixes.

@erikquinn  Is there an easy way to see if these changes work for the enroute demo (#937)?

erikquinn:
Approved!

PR #1056, opened 2018-08-21
fixed altitude after aircraft vectored off ILS

### LINKED ISSUE #994: Unnecessary precision in flight strip
Browser: Chrome 66.0.3359.181 (Official Build) (64-bit)

Airport: all

Approximate play time before issue: after an aircraft is cleared for an ILS approach and starts descending, but is vectored away from the ILS

#### Issue Description
image

The precision is absolutely unnecessary

#### Steps to reproduce
- clear an aircraft for the ILS and allow it to start descending on the glideslope and then vector it away

felixscheffer:
Resolves #994

Floor altitude to a 100 ft increment after aircraft vectored off ILS

original PR #1050

### DIFF
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -357,7 +357,8 @@ export default class Pilot {
         }
 
         const airport = AirportController.airport_get();
-        const descentAltitude = Math.min(aircraftModel.altitude, this._mcp.altitude);
+        const currentAltitude = _floor(aircraftModel.altitude, -2);
+        const descentAltitude = Math.min(currentAltitude, this._mcp.altitude);
         const altitudeToMaintain = Math.max(descentAltitude, airport.minAssignableAltitude);
 
         this._mcp.setAltitudeFieldValue(altitudeToMaintain);


src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -357,7 +357,8 @@ export default class Pilot {
         }
 
         const airport = AirportController.airport_get();
-        const descentAltitude = Math.min(aircraftModel.altitude, this._mcp.altitude);
+        const currentAltitude = _floor(aircraftModel.altitude, -2);
```
n8rzz: Why floor here instead of `Math.min`?

felixscheffer: The Math.min part is still there:  `const descentAltitude = Math.min(currentAltitude, this._mcp.altitude);`
The aircraft's altitude is just floored before comparing it with mcp.altitude.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 
 ### Bugfixes
-
+- [#994](https://github.com/openscope/openscope/issues/994) - fix "Aircraft vectored off ILS maintain over-precise altitude"
```
erikquinn: @felixscheffer try to keep changelog entries as added lines instead of edited lines, makes things easier at merge time.

PR #1058, opened 2018-08-21
bugfix/1044 - Prevent aircraft from deviating off centerline on short final

### LINKED ISSUE #1044: Strange behaviour after landing
Browser: Chrome

Airport: KBOS

#### Issue Description
After landing aircraft will decelerate very slowly causing them to overrun the runway. They also seem to be touching down well above the runway height.
runway overrun
In the picture above, UAL6608 landed on runway 27, but we can see it far past the terminals still going at 110kts and at 50ft (despite airport elevation being 19ft)
VIR123 landed on runway 22L and also overran the runway, currently in the water at 70kts

#### Steps to reproduce
- Land an aircraft and observe touchdown

felixscheffer:
Prevent aircraft from deviating off centerline on short final

Resolves openscope/openscope#1044

The target heading calculated by _calculateTargetedHeadingDuringLanding() is overridden in overrideTarget() causing aircrafts to drift off the ILS localizer.

original PR openscope/openscope#1052

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -1367,8 +1367,6 @@ export default class AircraftModel {
                 break;
 
             case FLIGHT_PHASE.LANDING: {
-                this.target.heading = this.mcp.course;
-
                 if (this.altitude <= this.mcp.nav1Datum.elevation) {
                     this.altitude = this.mcp.nav1Datum.elevation;
                     this.target.speed = 0;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Bugfixes
 - [#994](https://github.com/openscope/openscope/issues/994) - fix "Aircraft vectored off ILS maintain over-precise altitude"
 - [#1048](https://github.com/openscope/openscope/issues/1048) - fix "Aircrafts do not land after ILS clearance" because runway is not part of their STAR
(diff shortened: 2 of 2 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1367,8 +1367,6 @@ export default class AircraftModel {
                 break;
 
             case FLIGHT_PHASE.LANDING: {
-                this.target.heading = this.mcp.course;
```
n8rzz: @felixscheffer could you elaborate on _why_ this one little change fixes this issue?  I think I know where you're going with this, but it's been probably over a year since I was in this part of the code.  So I'm a little fuzzy on what is happening here.

@erikquinn do you have any thoughts on this one?

felixscheffer: `updateTarget( `) calls  `_calculateTargetedHeadingDuringLanding() ` during  `FLIGHT_PHASE.LANDING ` to calculate a course correction to steer the aircraft towards the runway (which it does). Then,  `overrideTarget() ` overrides the new target heading with `mcp.course`. 

So in the end, `_calculateTargetedHeadingDuringLanding()` has no effect and the ` target.heading ` is always  `mcp.course` 

Maybe there is a scenario where the target heading should be `mcp.course` but IMHO it should be addressed in `_calculateTargetedHeadingDuringLanding()`

erikquinn: Hm. Must have always been an issue then. A recent change made the phase change to `LANDING` occur further out, and that's why we're seeing an increase in the severity of them missing the runway. I concur with the removal :+1:

Issue #10, opened 2017-01-02
Command for crossing restrictions


erikquinn:
Issuing descent/speed instructions relative to a fix on an aircraft's flightplan is commonplace. For example:

>"AAL482, maintain 190 knots to FINKA"
>"UAL3185, cross LENDY at Flight Level 190 at 250 knots"

Adding a command that would allow for these types of instructions would do a lot for the enroute side of the simulation, whenever that comes along, and will still be very useful in the current version of the sim (especially in very large "super-TRACONs").

Closely related to #11.
Originally reported in zlsa/atc#315.

### Altitude crossing restrictions already completed in #1080.

n8rzz:
this feels like we can simply update a WaypointModel's `_altitudeRestriction` or `_speedResctriction` property.  we would have add some new commands, but based on the examples posted here, that should be fairly trivial.

erikquinn:
Correct. We would enter a new altitude/speed restriction at the appropriate waypoint, and then set the mcp altitude to the procedure's bottom altitude, and activate VNAV. All stuff that _magically works because we've set things up to emulate reality_. Awesome stuff!!

indianbhaji:
From what I've witnessed before, cross doesn't imply direct - if you want someone to direct, you say something along the lines of "DAL36, direct LENDY, descend to reach FL190 at LENDY"

erikquinn:
Not direct. Stay on the route and be at that altitude by the time you pass the specified fix.

JPmAn24:
I think I could make this work. An idea I had was to split the `cross`/`cr` command in two:
`cra`: Cross at a specific altitude
`crs`: Cross at a specific speed
Would this be okay with everyone?

erikquinn:
No, they should be parsed from the arguments of the same command (cross/cr). We should also add `X` to the aliases (stealing from taxi).

```
x hoggs a180
x irokt a90
x essso a100 s250
x dylin s210
```

j-651:
Would there be any conflict with x being used for "expedite"?

erikquinn:
@j-651 Should be fine. The command parser pays attention to the context of where the x is within the user's entry, and it should actually be able to tell the difference easily iirc.

erikquinn:
Speeds implemented, completing this issue:
- secondary issue created for reference number for the second half of this work: #1624.
- work originally from PR #1603.
- moved that PR to a new one of the right branch: #1625.
- merged to develop.

Closing this issue as complete.

PR #1063, opened 2018-08-22
bugfix/1045 - Arrivals ignoring altitude restrictions

### LINKED ISSUE #1045: Arrivals ignoring altitude restrictions
Browser: Chrome

Airport: KBOS

#### Issue Description
Despite the altitude restrictions on the STARs, aircraft are descending to the final altitude, well outside the airspace
low altitude
In the picture, we can see JBU3864 at 5,000ft aproaching `PROVI` which has a restriction of at or above 11,000ft

felixscheffer:
Resolves #1045

I tried to come up with a simple solution for openscope/openscope#1045 and ended up completely rewriting `_calculateTargetedAltitudeVnav()`

Now, before I start writing tests, i was hoping that someone could provide some feedback ;-)
@n8rzz  @erikquinn 

There are basically two issues that i am trying to solve here:

1. Aircrafts should not descend to `mcp.altitude` until VNAV minimumal altitude allows it.
2. Some aircrafts are spawned below VNAV minimal altitude. For example aircrafts are spawned at 18,000ft at KBOS and send on a route with VNAV minimal altitude at 21,000ft. Now there is nothing wrong with this (=takeoff from a nearby airport still climbing) but the current implementation ends up sending aircrafts down to 5,000ft (`mcp.altitude`) instead of climbing them to 21,000ft.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -1,15 +1,12 @@
 import _ceil from 'lodash/ceil';
 import _defaultTo from 'lodash/defaultTo';
-import _filter from 'lodash/filter';
 import _findIndex from 'lodash/findIndex';
 import _floor from 'lodash/floor';
 import _forEach from 'lodash/forEach';
 import _get from 'lodash/get';
-import _head from 'lodash/head';
 import _isEmpty from 'lodash/isEmpty';
 import _isEqual from 'lodash/isEqual';
 import _isNil from 'lodash/isNil';
-import _last from 'lodash/last';
 import _uniqueId from 'lodash/uniqueId';
 import AircraftTypeDefinitionModel from './AircraftTypeDefinitionModel';
 import AirportController from '../airport/AirportController';
@@ -841,26 +838,12 @@ export default class AircraftModel {
      *
      * @for AircraftModel
      * @method isBeyondTopOfDescentForWaypointModel
+     * @param waypointModel {WaypointModel} the waypoint at which to comply with the restriction
+     * @param targetAltitude {number} the altitude to comnply with
      * @return {boolean}
      */
-    isBeyondTopOfDescentForWaypointModel() {
-        const currentAltitude = this.altitude;
-        const altitudeRestrictedWaypoints = this.fms.getAltitudeRestrictedWaypoints();
-        const waypointsWithRelevantCeiling = _filter(altitudeRestrictedWaypoints,
-            (waypoint) => waypoint.hasMaximumAltitudeAtOrBelow(currentAltitude));
-
-        if (_isEmpty(altitudeRestrictedWaypoints)) {
-            return;
-        }
-
-        let targetAltitude = this.mcp.altitude;
-        let targetPosition = _last(altitudeRestrictedWaypoints).positionModel;
-
-        if (!_isEmpty(waypointsWithRelevantCeiling)) {
-            targetAltitude = _head(waypointsWithRelevantCeiling).altitudeMaximum;
-            targetPosition = _head(waypointsWithRelevantCeiling).positionModel;
-        }
-
+    isBeyondTopOfDescentForWaypointModel(waypointModel, targetAltitude) {
+        const targetPosition = waypointModel.positionModel;
         const waypointDistance = this.positionModel.distanceToPosition(targetPosition);
         const altitudeChange = targetAltitude - this.altitude;
         const descentRate = -this.model.rate.descent * PERFORMANCE.TYPICAL_DESCENT_FACTOR;
@@ -1845,8 +1828,8 @@ export default class AircraftModel {
 
         if (shouldMoveToNextFix) {
             if (!this.fms.hasNextWaypoint()) {
-                // we've hit this block becuase and aircraft is about to fly over the last waypoint in its flightPlan
-                this.pilot.maintainPresentHeading(this.heading);
+                // we've hit this block because and aircraft is about to fly over the last waypoint in its flightPlan
+                this.pilot.maintainPresentHeading(headingToWaypoint);
 
                 return headingToWaypoint;
             }
@@ -1957,27 +1940,27 @@ export default class AircraftModel {
      * @return {number}
      */
     _calculateTargetedAltitudeVnav() {
-        const nextAltitudeMaximumWaypoint = this.fms.findNextWaypointWithMaximumAltitudeRestriction();
-        const nextAltitudeMinimumWaypoint = this.fms.findNextWaypointWithMinimumAltitudeRestriction();
-        const maximumAltitudeExists = !_isNil(nextAltitudeMaximumWaypoint);
-        const minimumAltitudeExists = !_isNil(nextAltitudeMinimumWaypoint);
+        const altitudeMaximumWaypoint = this.fms.findNextWaypointWithMaximumAltitudeRestriction();
+        const altitudeMinimumWaypoint = this.fms.findNextWaypointWithMinimumAltitudeRestriction();
+        const maximumAltitudeExists = !_isNil(altitudeMaximumWaypoint);
+        const minimumAltitudeExists = !_isNil(altitudeMinimumWaypoint);
 
         if (this.mcp.altitude < this.altitude) {
             // we want to descend...
-            if (!minimumAltitudeExists || nextAltitudeMinimumWaypoint.altitudeMinimum <= this.mcp.altitude) {
+            if (!minimumAltitudeExists || altitudeMinimumWaypoint.altitudeMinimum <= this.mcp.altitude) {
                 // ... and there is nothing that can stop us.
                 return this.mcp.altitude;
             }
 
-            const altitudeMinimum = nextAltitudeMinimumWaypoint.altitudeMinimum;
+            const altitudeMinimum = altitudeMinimumWaypoint.altitudeMinimum;
 
             if (this.altitude < altitudeMinimum) {
                 // ... but we are too low and we have to comply with VNAV restriction
-                return this._calculateTargetedAltitudeVnavClimb(nextAltitudeMinimumWaypoint);
+                return this._calculateTargetedAltitudeVnavClimb(altitudeMinimumWaypoint);
             }
 
             if (maximumAltitudeExists) {
-                const altitudeMaximum = nextAltitudeMaximumWaypoint.altitudeMaximum;
+                const altitudeMaximum = altitudeMaximumWaypoint.altitudeMaximum;
 
                 if (this.mcp.altitude > altitudeMaximum) {
                     // we are too high but we are prioritizing clearance over VNAV restriction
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1862,32 +1862,102 @@ export default class AircraftModel {
...
+                            return nextAltitudeMinimumWaypoint.altitudeMinimum;
+                        }
+                    }
+
+                    // ...so descend to comply with VNAV restriction
+                    return this._calculateTargetedAltitudeVnavDescent(nextAltitudeMaximumWaypoint);
+                }
+            }
+        }
+        else {
```
n8rzz: `else` should be on the same line as the closing `}`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1862,32 +1862,102 @@ export default class AircraftModel {
...
-        const nextAltitudeMinimumWaypoint = this.fms.findNextWaypointWithMinimumAltitudeAtOrAbove(this.altitude);
+        const nextAltitudeMaximumWaypoint = this.fms.findNextWaypointWithMaximumAltitudeRestriction();
+        const nextAltitudeMinimumWaypoint = this.fms.findNextWaypointWithMinimumAltitudeRestriction();
         const maximumAltitudeExists = !_isNil(nextAltitudeMaximumWaypoint);
         const minimumAltitudeExists = !_isNil(nextAltitudeMinimumWaypoint);
 
-        if (!maximumAltitudeExists && !minimumAltitudeExists) {
-            return this.mcp.altitude;
-        }
+        if (this.mcp.altitude < this.altitude) {
```
n8rzz: holy nested `if` logic!  I'd *_really_* like to see this broken up into several other methods, all with as little conditional logic as possible.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1862,32 +1862,102 @@ export default class AircraftModel {
...
+     * @method _findFirstWaypoint
+     * @param waypoints
+     * @param waypointA {WaypointModel}
+     * @param waypointB {WaypointModel}
+     * @return waypointA or waypointB {WaypointModel}
+     */
+    _findFirstWaypoint(waypoints, waypointA, waypointB) {
+        const indexOfA = _findIndex(waypoints, (waypoint) => waypoint.name === waypointA.name);
+        const indexOfB = _findIndex(waypoints, (waypoint) => waypoint.name === waypointB.name);
+        return indexOfA < indexOfB ? waypointA : waypointB;
```
n8rzz: missing empty line above `return`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1862,32 +1862,102 @@ export default class AircraftModel {
...
+     * Takes two waypoints and returns the waypoint that comes first in the list of waypoints
+     *
+     * @for AircraftModel
+     * @method _findFirstWaypoint
+     * @param waypoints
+     * @param waypointA {WaypointModel}
+     * @param waypointB {WaypointModel}
+     * @return waypointA or waypointB {WaypointModel}
+     */
+    _findFirstWaypoint(waypoints, waypointA, waypointB) {
```
n8rzz: may have missed where this is getting called, but I feel like this logic should probably live within the `fms` instead

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1899,7 +1969,7 @@ export default class AircraftModel {
     _calculateTargetedAltitudeVnavClimb(waypointWithMinimumAltitudeRestriction) {
         const waypointMinimumAltitude = waypointWithMinimumAltitudeRestriction.altitudeMinimum;
```
n8rzz: if we're just returning a re-assigned property here, we can probably remove this method entirely and just use `waypointWithMinimumAltitudeRestriction.altitudeMinimum` in it's place.

src/assets/scripts/client/aircraft/AircraftModel.js

src/assets/scripts/client/aircraft/AircraftModel.js

src/assets/scripts/client/aircraft/AircraftModel.js
felixscheffer: called in `_calculateTargetedAltitudeVnav()`

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: The branches are all needed, and it's broken up really cleanly and commented well. I think it likely makes the most _sense_ in terms of readding and figuring out what the heck is going on as it is here. That said, we typically would avoid nested ifs like these as if they were the plague.

My two cents is that we roll with it for now.

felixscheffer:
I am trying to fix the `'.updateTarget() causes arrivals to descend when the STAR includes only AT or ABOVE altitude restrictions'` test and ran into a new issue. 

The STAR we are using for testing is `GRNPA2 `: `["KSINO", "A170+"], ["LUXOR", "A120+|S250"], ["GRNPA", "A110+"], ["DUBLX", "A90+"]`.

These altitude restrictions are minimal restrictions, so an arrival at 28000ft will stay at that altitude forever and will never descend because it's already above the required minimal altitude.

any thoughts?

n8rzz:
This is once place where things get a little tricky.  In order to accurately test the `AircaftModel`, we need to instantiate several fixtures like the `NavigationLibrary` and a few others.  `KLAS` was chosen mostly at random just cause I liked it and it was handy at the time.  We aren't really tied to testing with that specific STAR, it could really be any STAR.  However the assertions will need to be tailored to the procedure, obviously.

In the future, I would love to be able to be a little more flexible with how all those things are setup within tests.  Things like being able to configure a specific airport and a specific procedure within a test would be a good thing.  Not an easy thing to do at the moment, however 😦

felixscheffer:
@n8rzz  Okay, I got it now. Thanks for clarifying.

@erikquinn Can you deploy this branch?

I added some tests, but we definitely need more.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1862,32 +1862,102 @@ export default class AircraftModel {
...
+        const nextAltitudeMinimumWaypoint = this.fms.findNextWaypointWithMinimumAltitudeRestriction();
         const maximumAltitudeExists = !_isNil(nextAltitudeMaximumWaypoint);
         const minimumAltitudeExists = !_isNil(nextAltitudeMinimumWaypoint);
 
-        if (!maximumAltitudeExists && !minimumAltitudeExists) {
-            return this.mcp.altitude;
-        }
+        if (this.mcp.altitude < this.altitude) {
+            // we want to descend...
+            if (!minimumAltitudeExists || nextAltitudeMinimumWaypoint.altitudeMinimum <= this.mcp.altitude) {
```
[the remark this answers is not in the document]

test/aircraft/AircraftModel.spec.js
```diff
@@ -408,26 +409,92 @@ ava('.matchCallsign() returns true when passed a mixed case callsign that matche
...
+    moveAircraftToFix(model, "WILLW");
+    model.mcp.enable();
+    model.pilot.climbViaSid(model, 31000);
     model.updateTarget();
 
-    t.true(model.target.altitude === 8000);
+    t.true(model.target.altitude <= 14000);
 });
 
+function moveAircraftToFix(aircraft, fixName) {
```
felixscheffer: This method should probably live somewhere else...any suggestions?

n8rzz:
@felixscheffer I believe there is a testUtilities folder that could live in.  If that doesn't exist already, feel free to create one.  do you foresee this being used outside of the `AircraftModel` tests?  If not, then I'm ok with it living where it is for now.  

I'll get the deploy spinning right now...

felixscheffer:
@n8rzz Thanks. Ok, then the method will stay with the `AircraftModel` tests for now.

n8rzz:
you can always add a `TODO` comment about possible future abstraction so at least there is a breadcrumb for the future.

felixscheffer:
No, needs more tests.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -841,26 +838,12 @@ export default class AircraftModel {
...
-
-        let targetAltitude = this.mcp.altitude;
-        let targetPosition = _last(altitudeRestrictedWaypoints).positionModel;
-
-        if (!_isEmpty(waypointsWithRelevantCeiling)) {
-            targetAltitude = _head(waypointsWithRelevantCeiling).altitudeMaximum;
-            targetPosition = _head(waypointsWithRelevantCeiling).positionModel;
-        }
-
+    isBeyondTopOfDescentForWaypointModel(waypointModel, targetAltitude) {
```
felixscheffer: Hopefully, someday we can do something like "cross fix at 7000 ft or above"...

felixscheffer:
I added more tests...

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1845,8 +1828,8 @@ export default class AircraftModel {
 
         if (shouldMoveToNextFix) {
             if (!this.fms.hasNextWaypoint()) {
-                // we've hit this block becuase and aircraft is about to fly over the last waypoint in its flightPlan
-                this.pilot.maintainPresentHeading(this.heading);
+                // we've hit this block because and aircraft is about to fly over the last waypoint in its flightPlan
+                this.pilot.maintainPresentHeading(headingToWaypoint);
```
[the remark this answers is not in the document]
erikquinn: that's not the right value though, we definitely can't do this. Not sure what the test issue was, but this will need to be `this.heading`.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: #1178 & #10! One day maybe soon! :smiley:

src/assets/scripts/client/aircraft/AircraftModel.js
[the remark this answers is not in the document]
erikquinn: Also, in the app, it seems like the headings are not being assigned properly, because if I ask an airplane who has passed their last fix what heading they are on, they say `assigned heading NaN`. Not sure how they ended up with that, but they do.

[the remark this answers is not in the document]
erikquinn: Looks like `Pilot.maintainPresentHeading()` takes the aircraftModel, which it inspects to find the aircraft's current heading, rather than taking the heading. Sounds dumb (and kinda is), but the Pilot so frequently has to inspect the aircraft to execute an action that doing this has become a relatively common pattern with it. Hence why the Pilot needs to die (#1177).

Thus, this should be changed to `this.pilot.maintainPresentHeading(this);`. :scream: :grimacing:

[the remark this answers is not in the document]
erikquinn: done

PR #1098, opened 2018-09-29
feature/1080 - Sort flight strips list by drag and drop

### LINKED ISSUE #1080: Sort flight strip lists by drag and drop
Update the flight strip lists to be sortable using drag and drop. This would improve arrival and departure management of aircraft.

tdawe:
Resolves #1080 

The purpose of this pull request is to sort flight strips list by drag and drop.

### DIFF
--- a/src/templates/_stripView.hbs
+++ b/src/templates/_stripView.hbs
@@ -1,7 +1,20 @@
+<script src="assets/scripts/client/jquery.min.js"></script>
+<script src="assets/scripts/client/jquery-ui.min.js"></script>
+  <script>
+  $( function() {
+    $( ".stripView-list" ).sortable();
+    $( ".stripView-list" ).disableSelection();
+  } );
+  </script>
+
 <div id="sidebar" class="stripView js-stripView mix-stripView_isHidden">
     <div class="stripView-trigger js-stripView-trigger"></div>
     <div class="stripTitle">Arrivals</div>
-    <ul class='stripView-list nice-scrollbar js-stripViewArrivals-list'></ul>
+    <div class="stripViewContainer">
+        <ul class='stripView-list nice-scrollbar js-stripViewArrivals-list'></ul>
+    </div>
     <div class="stripTitle">Departures</div>
-    <ul class='stripView-list nice-scrollbar js-stripViewDepartures-list'></ul>
+    <div class="stripViewContainer">    
+        <ul class='stripView-list nice-scrollbar js-stripViewDepartures-list'></ul>
+    </div>
 </div>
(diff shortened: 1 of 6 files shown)


src/templates/_stripView.hbs
```diff
@@ -1,7 +1,20 @@
+<script src="assets/scripts/client/jquery.min.js"></script>
```
n8rzz: This is a big no-no in my book.  Any scripts that need to be included should be included in the main bundle.  If, for some reason, that is not possible then a script should be included before the bundle at the foot of the `_layout.hbs` template.

Additionally, `jquery` is already included in the project as part of the bundle, so including it here like this will cause problems with eventual version conflicts.

src/templates/_stripView.hbs
```diff
@@ -1,7 +1,20 @@
...
+<script src="assets/scripts/client/jquery-ui.min.js"></script>
```
n8rzz: I really question why you chose `jquery-ui`.  well, I can guess why, it used to be super common and its relatively quick to add drag and drop.

There are a few problems with that, though:
1) `jquery-ui` hasn't had a commit since September 2016, which suggests it is no longer maintained
1) all the browsers we support also support *native* drag and drop, no libraries needed (see: https://medium.com/quick-code/simple-javascript-drag-drop-d044d8c5bed5 or https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API)
1) we are slowly moving away from `jquery` entirely.  You just don't need it anymore and it adds additional weight to the code that we can live mostly without.

src/templates/_stripView.hbs
```diff
@@ -1,7 +1,20 @@
...
+  <script>
```
n8rzz: This is another big no-no.  Absolutely no inline javascript.

Any js should live within the existing codebase and *never* in our templates.  The only exception here is our google analytics snippet.

tools/tasks/scriptTasks.js
```diff
@@ -36,6 +36,15 @@ module.exports = function(gulp, config) {
             .pipe(gulp.dest(OPTIONS.DIR.DIST_SCRIPTS_CLIENT));
     }
 
+    gulp.task('jquery', function() {
```
n8rzz: remove this task

n8rzz: Actually, I'm a little confused, why would you include a manual script tag if you've also included `jquery-ui` in the `package.json`?

felixscheffer:
I would like to suggest that we close this PR as invalid and create a new PR if necessary.

n8rzz:
I don’t think that’s necessary. Work can continue here without issue.

tdawe:
Thanks for the feedback. I used jquery-ui since it was the last library I used for drag and drop, but that was years ago when it was still actively maintained. I'll get this working with native drag and drop.

PR #1106, opened 2018-10-04
bugfix/1104 - Fix flight strip console warning

### LINKED ISSUE #1104: Console warnings for removing StripViewModel which doesn't exist
Console warnings appear regarding attempts to remove a `StripViewModel` from the `StripViewCollection`.

`Attempted to remove a StripViewModel for QXE7954 that does not exist. This is likely not a fatal problem, but if you are seeing this, please let somebody know.`

Comes from `StripViewController.removeStripView()`.

image

felixscheffer:
Resolves #1104.

Fix console warnings for removing `StripViewModel` which doesn't exist.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -321,8 +321,6 @@ export default class AircraftController {
         for (let i = 0; i < this.aircraft.list.length; i++) {
             this.aircraft_remove(this.aircraft.list[i]);
         }
-
-        this.aircraft.list = [];
     }
 
     /**
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -319,7 +319,11 @@ export default class AircraftController {
      */
     aircraft_remove_all() {
         for (let i = 0; i < this.aircraft.list.length; i++) {
-            this.removeStripView(this.aircraft.list[i]);
+            const aircraft = this.aircraft.list[i];
+
+            if (aircraft.isControllable) {
+                this.removeStripView(aircraft);
+            }
         }
 
         this.aircraft.list = [];
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -319,11 +319,7 @@ export default class AircraftController {
      */
     aircraft_remove_all() {
         for (let i = 0; i < this.aircraft.list.length; i++) {
-            const aircraft = this.aircraft.list[i];
-
-            if (aircraft.isControllable) {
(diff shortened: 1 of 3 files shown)


felixscheffer:
We are calling `AircraftController aircraft_remove_all()` and then `removeStripView` is called for every aircraft.

felixscheffer:
@erikquinn I moved the guard.

erikquinn:
Actually I'm not sure what I was thinking. I was assuming you moved the guard to the method that removes a single aircraft. As you mentioned in the issue for this PR, they are often removed individually and not by doing all of them. It should be moved one more time.

felixscheffer:
The warning appears when we remove a single aircraft AND when we remove all aircrafts (before switching to another airport), so I added another guard to `aircraft_remove()`. If you can come up with a better solution feel free to take over.

felixscheffer:
The more I look at it, the more I favor the idea of removing the warning (and the guards) and simply do nothing if the strip does not exist. None of the warnings actually indicate a real problem and the more guards we add, the more over-engineered the code feels.

erikquinn:
```
/**
 * @for AircraftController
 * @method aircraft_remove_all
 */
aircraft_remove_all() {
    for (let i = 0; i < this.aircraft.list.length; i++) {
        this.removeStripView(this.aircraft.list[i]);
    }

    this.aircraft.list = [];
}
```

I think I'm finally seeing the reason for the confusion. I assumed `aircraft_remove_all()` would be _calling_ `aircraft_remove()` for each aircraft, but it does not. It simply overwrites the aircraft collection with a new empty array. Frankly, I'm shocked we haven't seen problems with that-- there are reasons certain things need to be done when aircraft are removed, and to remove all of them in a single call should be the same as removing all of them one at a time.

We need to update the remove-all method to call remove() on all individually. And have `aircraft_remove()` only attempt to remove a strip _if the aircraft has one_.

It might make sense to remove the warning? I'm not sure, I'd defer to @n8rzz since he did the work on the strips and would know best whether or not that is a good idea.

felixscheffer:
`aircraft_remove_all` now calls `aircraft_remove ` for each aircraft.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -319,7 +319,7 @@ export default class AircraftController {
      */
     aircraft_remove_all() {
         for (let i = 0; i < this.aircraft.list.length; i++) {
-            this.removeStripView(this.aircraft.list[i]);
+            this.aircraft_remove(this.aircraft.list[i]);
         }
 
         this.aircraft.list = [];
```
erikquinn: This should also be removed from remove_all(), because remove() will take the aircraft out of the aircraft list (via `.removeAircraftModelFromList()`)

PR #1107, opened 2018-10-04
bugfix/1097 - Fix speed restrictions

### LINKED ISSUE #1097: Speed restrictions don't work correctly
Airport: LTBA

#### Issue Description
At LTBA, every STAR has a speed restriction of 220 kts at the first downwind waypoint.
Let's take `ENESU1A` as an example eventhough this applies to every STAR.
       
        "ENESU1A": {
            "icao": "ENESU1A",
            "name": "Enesu One Alpha",
            "entryPoints": {
                "ENESU": []
            },
            "body": ["ENESU", ["CRL", "S250-"], ["BA611", "A120+"], "BA612", "ERLEB", ["BA791", "S220"], ["BA792", "A40+|A60-"], "BA793", "BA794", "BA795", "BA796", "BA797", "BA798", "#234"],
            "rwy": {
                "LTBA05": []
            },
            "draw": [
                ["ENESU", "CRL", "BA611", "BA612", "ERLEB", "BA791", "BA792", "BA793", "BA794", "BA795", "BA796", "BA797", "BA798"]
            ]
        }
220 kts at BA791, but even after passing the waypoint, the aircraft still maintain 250kts:
screenshot_129

#### Steps to reproduce
- Wait for an aircraft (with ENESU1A routing) to pass BA791
- Check indicated airspeed
---
#### Additional information
I thought this slight shortcut could have something to do with that:
screenshot_128
And therefore I checked if #1038 would change anything: Same slight shortcut and also no speed reduction.

felixscheffer:
Resolves #1097.

Fix speed restrictions.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -827,7 +827,7 @@ export default class AircraftModel {
             return false;
         }
 
-        if (this._isReadyToMoveToWaypoint(waypointModel)) {
+        if (this._isPassingWaypoint(waypointModel)) {
             return true;
         }
 
@@ -1814,7 +1814,7 @@ export default class AircraftModel {
         const waypointPosition = this.fms.currentWaypoint.positionModel;
         const headingToWaypoint = this.positionModel.bearingToPosition(waypointPosition);
 
-        if (this._isReadyToMoveToWaypoint(this.fms.currentWaypoint)) {
+        if (this._isPassingWaypoint(this.fms.currentWaypoint)) {
             if (!this.fms.hasNextWaypoint()) {
                 // we've hit this block because an aircraft is about to fly over the last waypoint in its flightPlan
                 this.pilot.maintainPresentHeading(this.heading);
@@ -1848,11 +1848,11 @@ export default class AircraftModel {
      * Determine if it is time to move to the next fix.
      *
      * @for AircraftModel
-     * @method _isReadyToMoveToWaypoint
+     * @method _isPassingWaypoint
      * @param waypoint {WaypointModel}
      * @return {boolean}
      */
-    _isReadyToMoveToWaypoint(waypoint) {
+    _isPassingWaypoint(waypoint) {
         const waypointPosition = waypoint.positionModel;
         const distanceToWaypoint = this.positionModel.distanceToPosition(waypointPosition);
         const closeToBeingOverFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_PASS_WAYPOINT_NM;
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -827,6 +827,10 @@ export default class AircraftModel {
             return false;
         }
 
+        if (this._isReadyToMoveToNextFix()) {
+            return true;
+        }
+
         const waypointSpeed = waypointModel.speedMaximum;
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1850,6 +1844,31 @@ export default class AircraftModel {
...
 
+    /**
+     * Determine if it is time to move to the next fix.
+     *
+     * @for AircraftModel
+     * @method _isReadyToMoveToWaypoint
+     * @param waypoint {WaypointModel}
+     * @return {boolean}
+     */
+    _isReadyToMoveToWaypoint(waypoint) {
```
erikquinn: Is this checking whether or not it is ready to move _past_ the provided waypoint?

felixscheffer: It checks whether the aircraft is close enough to the waypoint to consider it reached. So maybe `hasReachedWaypoint` would be a better name.

erikquinn: Better, though I would say that `.hasReachedWaypoint()` has the implication that it would return true for all waypoints which have been passed (which it won't do). Maybe like `.isPassingWaypoint()` or something? That's not great, but is in the right direction I think.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1807,22 +1812,11 @@ export default class AircraftModel {
...
-        const closeEnoughToFlyByFix = distanceToWaypoint < PERFORMANCE.MAXIMUM_DISTANCE_TO_FLY_BY_WAYPOINT_NM;
-        const shouldFlyByFix = closeEnoughToFlyByFix && isTimeToStartTurning;
-        let shouldMoveToNextFix = closeToBeingOverFix;
 
-        if (!this.fms.currentWaypoint.isFlyOverWaypoint) {
-            shouldMoveToNextFix = closeToBeingOverFix || shouldFlyByFix;
-        }
-
-        if (shouldMoveToNextFix) {
+        if (this._isPassingWaypoint(this.fms.currentWaypoint)) {
```
erikquinn: This branch should be reversed.
```
if (!this._isPassingWaypoint()) {
    return headingToWaypoint;
}
```

felixscheffer:
True. I think this is basically like #1063.

We have to figure out if we are on a SID or a STAR and then apply different logics.

erikquinn:
Hm, right.... :thinking: 

@felixscheffer So what do you think is best to do? Can we combine them? Should we merge both individually and then clean up? I'm not wholly opposed to fixing one thing to break another if it's moving in the right direction, provided we come back to fix what we broke...

Open to input on what to do, because we want to see these things fixed, but since you are doing the work on both, you'd probably know best how to proceed. So what do you suggest?

felixscheffer:
@erikquinn I would prefer to finish #1063 first, so we can ship it with the next release and then focus on the speed restrictions. Assuming that we have tests, cleaning things up later shouldn't be an issue (hopefully).

I haven't had the time to think about a solution for #1097 yet, so if you have any ideas, let me know.

erikquinn:
Moving to next sprint.

@felixscheffer what is the status of this one?

felixscheffer:
@erikquinn I don't know. Currently, I think there are probably multiple speed related issues in the code. I stopped working on this PR to finish #1063. Right now it's definitely not my top priority.

erikquinn:
@felixscheffer Gotcha, no problem. I'll mark it orphaned for now, and leave it around in case it proves to be useful when somebody comes back to fix the same issue. :+1:

PR #1122, opened 2018-10-15
bugfix/1103 - Fix pressing ESC doesn't fully clear the command bar

### LINKED ISSUE #1103: Pressing ESC doesn't fully clear the command bar
Browser: Chrome

#### Issue Description
When the command  bar has a callsign and some stuff typed in, pressing escape should remove all but the callsign. When the command bar has no callsign but just other stuff typed in, pressing escape should clear everything. For the latter, the current behavior actually sets the command bar content with a single space, which if you ESC again, you can remove. It should have been removed with the first ESC press.

#### Steps to reproduce
1. Type `timewarp 10` into the command bar (do not press enter).
1. Press the ESC key.
1. Observe that the command bar is cleared except for a space.
1. Press the ESC key.
1. Observe the command bar fully and properly cleared.

Screenshots of the progression:

image
image
image
image

TheBear44:
Resolves openscope/openscope#1103 

I add the condition to check if the input callsign is empty before fully clear the command bar.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -450,9 +450,9 @@ export default class InputController {
             case KEY_CODES.ESCAPE: {
                 this.closeAllDialogs();
 
-                const hasCallsign = !_includes(currentCommandInputValue, this.input.callsign);
+                const hasCallsign = _includes(currentCommandInputValue, this.input.callsign);
                 const hasOnlyCallsign = currentCommandInputValue.trim() === this.input.callsign;
-                const hasSelectedCallsign = this.input.callsign === '';
+                const hasSelectedCallsign = this.input.callsign !== '';
 
                 if (!hasCallsign || hasOnlyCallsign || !hasSelectedCallsign) {
                     this.deselectAircraft();
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -450,10 +450,11 @@ export default class InputController {
             case KEY_CODES.ESCAPE: {
                 this.closeAllDialogs();
 
-                if (!_includes(currentCommandInputValue, this.input.callsign) ||
-                    currentCommandInputValue.trim() === this.input.callsign ||
-                    this.input.callsign === ''
-                ) {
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/InputController.js
```diff
@@ -451,7 +451,8 @@ export default class InputController {
                 this.closeAllDialogs();
 
                 if (!_includes(currentCommandInputValue, this.input.callsign) ||
-                    currentCommandInputValue.trim() === this.input.callsign
+                    currentCommandInputValue.trim() === this.input.callsign ||
+                    this.input.callsign === ''
```
erikquinn: For readability, could we break these three conditions out into clearly named consts, and then do the if on the consts, instead of inline?

TheBear44: Done in 74d768a

src/assets/scripts/client/InputController.js
```diff
@@ -450,9 +450,11 @@ export default class InputController {
...
                 this.closeAllDialogs();
 
-                if (!_includes(currentCommandInputValue, this.input.callsign) ||
-                    currentCommandInputValue.trim() === this.input.callsign
-                ) {
+                const noCallsign = !_includes(currentCommandInputValue, this.input.callsign);
+                const hasOnlyCallsign = currentCommandInputValue.trim() === this.input.callsign;
+                const noSelectedCallsign = this.input.callsign === "";
+
+                if (noCallsign || hasOnlyCallsign || noSelectedCallsign) {
```
n8rzz: I'd like for you to rename two of these _(also note the single quotes)_:
```js
const hasCallsign = _includes(currentCommandInputValue, this.input.callsign);
const hasSelectedCallsign = this.input.callsign === '';
```

Then in your if, just reverse the logic for those two thusly:
```js
 if (!hasCallsign || hasOnlyCallsign || !hasSelectedCallsign) {
  // ...
}
```

TheBear44: Done by 61a95aa e38f014

TheBear44:
It won't if you select the callsign from the strip. But if you type the aircraft callsign explicitly into the command bar, it will fully clear no matter what you type in.

PR #1125, opened 2018-10-16
feature/1003 - Add assigned runway to the aircrafts' strips

### LINKED ISSUE #1003: Add assigned runway to the aircrafts' strips
As already said on Slack, it would be good if it would be possible to see to which runway the aircraft is currently added.
"Would make playing much easier for players who aren't familiar with the airport. I mean, after all it's just a simulator that people play for fun and probably don't want to memorize every departure route with the corresponding runway."

felixscheffer:
Resolves #1003

### DIFF
--- a/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
+++ b/src/assets/scripts/client/aircraft/StripView/StripViewModel.js
@@ -23,7 +23,8 @@ const STRIP_VIEW_SELECTORS = {
     ARRIVAL_AIRPORT_ID: '.js-stripView-arrivalAirportId',
     ALTERNATE_AIRPORT_ID: '.js-stripView-alternateAirportId',
     FLIGHT_PLAN: '.js-stripView-flightPlan',
-    REMARKS: '.js-stripView-remarks'
+    REMARKS: '.js-stripView-remarks',
+    RUNWAY: '.js-stripView-runway'
 };
 
 /**
@@ -320,11 +321,49 @@ export default class StripViewModel extends BaseModel {
         /**
          * HTML Element that holds the `#_flightPlan` value
          *
-         * @property $_flightPlanView
+         * @property $flightPlanView
          * @type {JQuery Element}
          * @default null
          */
         this.$flightPlanView = null;
+
+        /**
+         * Value of the remarks field
+         *
+         * @property _remarks
+         * @type {string}
+         * @default ''
+         * @private
+         */
+        this._remarks = '';
+
+        /**
+         * HTML Element that holds the `#remarks` value
+         *
+         * @property $remarks
+         * @type {JQuery Element}
+         * @default null
+         */
+        this.$remarks = null;
+
+        /**
+         * The expected runway
+         *
+         * @property _runway
+         * @type {string}
+         * @default ''
+         * @private
+         */
+        this._runway = '';
+
+        /**
+         * HTML Element that holds the `#runway` value
+         *
+         * @property $runway
+         * @type {JQuery Element}
+         * @default null
+         */
+        this.$runway = null;
 
         return this._init(aircraftModel)
             ._createChildren()
@@ -381,6 +420,8 @@ export default class StripViewModel extends BaseModel {
         this._flightPlan = flightPlan;
         this._categoryClassName = this._buildClassnameForFlightCategory(aircraftModel);
         this.isDeparture = aircraftModel.isDeparture();
+        // TODO: remarks
+        this._runway = this._buildRunway(aircraftModel);
 
         return this;
     }
@@ -406,6 +447,8 @@ export default class StripViewModel extends BaseModel {
         this.$departureAirportView = this.$element.find(STRIP_VIEW_SELECTORS.DEPARTURE_AIRPORT_ID);
         this.$alternateAirportView = this.$element.find(STRIP_VIEW_SELECTORS.ALTERNATE_AIRPORT_ID);
         this.$flightPlanView = this.$element.find(STRIP_VIEW_SELECTORS.FLIGHT_PLAN);
+        this.$remarks = this.$element.find(STRIP_VIEW_SELECTORS.REMARKS);
+        this.$runway = this.$element.find(STRIP_VIEW_SELECTORS.RUNWAY);
 
         return this;
     }
@@ -483,6 +526,8 @@ export default class StripViewModel extends BaseModel {
         this.$arrivalAirportView.text(this._arrivalAirport);
         this.$alternateAirportView.text(this._alternateAirport);
         this.$flightPlanView.text(this._flightPlan);
+        this.$remarks.text(this._remarks);
+        this.$runway.text(this._runway);
 
         return this;
     }
@@ -537,6 +582,10 @@ export default class StripViewModel extends BaseModel {
         this.$alternateAirportView = null;
         this._flightPlan = '';
         this.$flightPlanView = null;
+        this._remarks = '';
+        this.$remarks = null;
+        this._runway = '';
+        this.$runway = null;
 
         return this;
     }
@@ -594,6 +643,27 @@ export default class StripViewModel extends BaseModel {
         }
 
         return SELECTORS.CLASSNAMES.ARRIVAL;
+    }
+
+    /**
+     * Return the assigned runway based on whether an aircraft is a `departure` or an `arrival`
+     *
+     * @for AircraftStripView
+     * @method _buildRunway
+     * @return {string}
+     */
+    _buildRunway(aircraftModel) {
+        const fms = aircraftModel.fms;
+
+        if (fms.departureRunwayModel != null) {
+            return fms.departureRunwayModel.name;
+        }
+
+        if (fms.arrivalRunwayModel != null) {
+            return fms.arrivalRunwayModel.name;
+        }
+
+        return '';
     }
 
     /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 6.7.0 (November 1, 2018)
 ### New Features
 - [#1033](https://github.com/openscope/openscope/issues/1033) - Add support for overflights
+- [#1003](https://github.com/openscope/openscope/issues/1003) - Add assigned runway to the aircrafts' strips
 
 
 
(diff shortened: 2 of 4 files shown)


src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -381,6 +420,8 @@ export default class StripViewModel extends BaseModel {
         this._flightPlan = flightPlan;
         this._categoryClassName = this._buildClassnameForFlightCategory(aircraftModel);
         this.isDeparture = aircraftModel.isDeparture();
+        // TODO: remarks
```
n8rzz: I'm a huge fan of `TODO` comments, as I'm sure you've seen already.  If you're planning to address this with the PR, please do so.  Otherwise, could you add a little more information here about what exactly this `TODO` is for?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -596,6 +645,27 @@ export default class StripViewModel extends BaseModel {
...
     }
 
+    /**
+     * Return the assigned runway based on whether an aircraft is a `departure` or an `arrival`
+     *
+     * @for AircraftStripView
+     * @method _buildRunway
+     * @return {string}
+     */
+    _buildRunway(aircraftModel) {
```
n8rzz: Why not something like the following:

```js
_buildRunway(aircraftModel) {
    let runwayName = '';
    
    if (aircraftModel.isDeparture()) {
        runwayName = aircraftModel.fms.departureRunwayModel.name;
    } else if (aircraftModel.isArrival()) {
        runwayName = aircraftModel.fms.arrivalRunwayModel.name;
    }
          
    return runwayName;
}
```

additionally, loose equal comparisons should throw a linting error: `!=` or `==`, we stick to strict always.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: I hate `TODO`. Always feels like an excuse to be lazy and not do what needs to be done.

Anyways, I have no idea what the remarks field is intended for. I found it in the template and added the missing code. 

@erikquinn  any thoughts?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: Loose equal comparison was my Java autopilot 🙈 but my eslint did not complain about it. But then again, I am not sure that my eslint is working properly.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: It's the remarks on the flight plan. Whenever somebody files a flight plan, they can add notes to it, and probably the reason it was built in is because that's where they would go.

felixscheffer:
The main purpose of this change is to give people a better understanding of what they have to do, i.e. to display the runway the aircraft wants to taxi to (see #1003). If we only show the runway AFTER taxi clearance was given, we are not solving the problem.

erikquinn:
Hm, okay. We just want it to be at least allegedly realistic. So we could achieve both like this:
- show the expected runway assignment at spawn, in red (color of pre-planning strip marking)
- when expected runway is changed, and aircraft still hasn't taxiied, update the strip
- when executing taxi instruction, update the strip to the runway he's taxiing to, in black
- when taxiied to a different runway, update the strip to the new runway, in black

That should be everything it would have to do to combine the two.

felixscheffer:
@erikquinn  What about arrivals?

erikquinn:
Same thing, basically:
- show the expected arrival runway assignment at spawn, in red (color of pre-planning strip marking)
- when expected arrival runway is changed, and aircraft still hasn't been cleared for an approach, update the strip (still in red, because it's not assigned yet)
- when receiving an approach clearance, update the strip to the runway he's cleared approach to, in black
- when given a new approach clearance to a different runway, update the strip to the new runway, in black

felixscheffer:
@erikquinn Can you confirm that it's now working as it should please?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: @erikquinn  Do we support remarks in flight plans already?

erikquinn: no, it would be an easy add, but I don't see the need. It was just easy enough to set up a bunch of fields when building the flight strips so they'd be plug-and-play with future things we could add to them.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -596,6 +654,36 @@ export default class StripViewModel extends BaseModel {
...
+     *
+     * @for AircraftStripView
+     * @method _buildRunwayInformation
+     * @param aircraftModel {AircraftModel}
+     * @return {object}
+     */
+    _buildRunwayInformation(aircraftModel) {
+        if (aircraftModel.isDeparture()) {
+            return {
+                assigned: aircraftModel.pilot.hasDepartureClearance,
```
erikquinn: `hasRunwayAssigned: aircraft.flightPhase === FLIGHT_PHASE.APRON`

or even better, add and use `aircraft.isApron()

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -596,6 +654,36 @@ export default class StripViewModel extends BaseModel {
...
+                name: aircraftModel.fms.departureRunwayModel.name
+            };
+        }
+
+        if (aircraftModel.isArrival() !== null) {
```
erikquinn: Just `if (aircraftModel.isArrival()) {`. This will always return a boolean.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: okay, so is there anything you want me to do here or do we keep it as it is right now?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: true...copy-n-paste-fail

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: no, you can ignore anything regarding remarks. I saw a `// TODO: remarks` somewhere you could also remove.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: done

felixscheffer:
@erikquinn Can you have another look, please? Probably not working as it should.

erikquinn:
The arrivals are good. The departures are triggering at the right time, but they are currently _starting black_ and changing to red when receiving the runway assignment. The colors should be reversed: red first, then black when assigned.

felixscheffer:
I fixed that a moment ago...

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -23,9 +23,17 @@ const STRIP_VIEW_SELECTORS = {
...
+    REMARKS: '.js-stripView-remarks',
+    RUNWAY: '.js-stripView-runway'
 };
 
+/**
+ * @property STRIP_VIEW_PREPLANNING_CLASS
+ * @type {string}
+ * @final
+ */
+const STRIP_VIEW_PREPLANNING_CLASS = "stripView-preplanning";
```
n8rzz: this should live in either `STRIP_VIEW_SELECTORS` above or in the main `SELECTORS` object.  Also, no double quotes, only single quote.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -320,12 +328,50 @@ export default class StripViewModel extends BaseModel {
         /**
          * HTML Element that holds the `#_flightPlan` value
          *
-         * @property $_flightPlanView
+         * @property $flightPlanView
```
n8rzz: was this change intentional?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -483,6 +532,9 @@ export default class StripViewModel extends BaseModel {
         this.$arrivalAirportView.text(this._arrivalAirport);
         this.$alternateAirportView.text(this._alternateAirport);
         this.$flightPlanView.text(this._flightPlan);
+        this.$remarks.text(this._remarks);
+        this.$runway.text(this._runwayInformation.name);
+        this.$runway.toggleClass(STRIP_VIEW_PREPLANNING_CLASS, !this._runwayInformation.assigned);
```
n8rzz: I'm a little wary about this one.  My goal for this method was to simple accept data and render it.  Modifying css and such feels like it should happen somewhere else within this class.

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: Don't we have a constants file for classes? That's where I would think it would belong.

felixscheffer: we have `CLASSNAMES` in `selectors.js`

erikquinn: Okay, exactly. So it would belong there then, no?

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: Seems like the only place that it could work. Maybe he means to put it in _layout()? That's another place where a style class is applied, and the docblock says it is run every redraw-- but it actually isn't. So without having redraw() call layout() each time, the only place it really could go is in redraw()...

n8rzz: sorry guys, phone PRs are not the best PRs ☹️ 

The `.toggleClass` mentioned here should live in `_.updateStripView()`

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: done

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
felixscheffer: done

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: moved there

PR #1129, opened 2018-10-20
feature/793 - Add `airac` command and property to airports

### LINKED ISSUE #793: Add AIRAC airport file key and system command
We need a way to share with the users the AIRAC version of the airport file they're using, or at least to record it for our own purposes of keeping them fairly up to date.

We can add a `airac` key to all airport files (and airport format documentation!) that we store in the `AirportModel`. A simple system command could then output the AIRAC number in use for the given airport. That way, if users are confused about why they can't use the procedures from current charts, we can all see the chart sets with which the airport is _currently_ equipped. :smiley:

More than anything, I think this will do a lot for our sanity in maintaining our library of airports. When we see an airport still rocking AIRAC 1704 and the current is 1711, we will :grimacing: and :wrench: it!

clscott42:
Resolves #793.

Add `airac` command and property to airports, such that users (or more likely, developers) can type `airac` and have the simulator respond with the airac number the airport is using.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -701,13 +701,14 @@ export default class InputController {
             }
             case PARSED_COMMAND_NAME.AIRAC: {
                 const airacCycle = AirportController.getAiracCycle();
+
                 if (airacCycle === null) {
                     UiController.ui_log('unknown airac cycle');
                 } else {
                     UiController.ui_log(`airac cycle ${airacCycle}`);
                 }
 
-                break;
+                return true;
             }
             // TODO: this will be removed entirely, eventually.
             case PARSED_COMMAND_NAME.RATE:
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -699,6 +699,16 @@ export default class InputController {
 
                 return true;
             }
+            case PARSED_COMMAND_NAME.AIRAC: {
+                const airacCycle = AirportController.getAiracCycle();
+                if (airacCycle === null) {
+                    UiController.ui_log('unknown airac cycle');
+                } else {
+                    UiController.ui_log(`airac cycle ${airacCycle}`);
+                }
+
+                break;
+            }
             // TODO: this will be removed entirely, eventually.
             case PARSED_COMMAND_NAME.RATE:
                 UiController.ui_log('this command has been deprecated', true);
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -700,13 +700,16 @@ export default class InputController {
                 return true;
             }
             case PARSED_COMMAND_NAME.AIRAC: {
+                const airportIcao = AirportController.current.icao.toUpperCase();
                 const airacCycle = AirportController.getAiracCycle();
 
-                if (airacCycle) {
-                    UiController.ui_log(`AIRAC cycle ${airacCycle}`);
-                } else {
-                    UiController.ui_log('unknown AIRAC cycle');
+                if (!airacCycle) {
+                    UiController.ui_log(`${airportIcao} AIRAC cycle: unknown`);
+
+                    return true;
                 }
+
+                UiController.ui_log(`${airportIcao} AIRAC cycle: ${airacCycle}`);
 
                 return true;
             }
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -702,10 +702,10 @@ export default class InputController {
             case PARSED_COMMAND_NAME.AIRAC: {
                 const airacCycle = AirportController.getAiracCycle();
 
-                if (airacCycle === null) {
-                    UiController.ui_log('unknown airac cycle');
+                if (airacCycle) {
+                    UiController.ui_log(`AIRAC cycle ${airacCycle}`);
                 } else {
-                    UiController.ui_log(`airac cycle ${airacCycle}`);
+                    UiController.ui_log('unknown AIRAC cycle');
                 }
 
                 return true;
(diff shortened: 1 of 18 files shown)


src/assets/scripts/client/InputController.js
```diff
@@ -715,6 +715,16 @@ export default class InputController {
...
             }
+            case PARSED_COMMAND_NAME.AIRAC: {
+                const airacCycle = AirportController.getAiracCycle();
+                if (airacCycle === null) {
+                    UiController.ui_log('unknown airac cycle');
+                } else {
+                    UiController.ui_log(`airac cycle ${airacCycle}`);
+                }
+
+                break;
```
felixscheffer:
```suggestion
                return true;
```

src/assets/scripts/client/InputController.js
```diff
@@ -715,6 +715,16 @@ export default class InputController {
 
                 return true;
             }
+            case PARSED_COMMAND_NAME.AIRAC: {
+                const airacCycle = AirportController.getAiracCycle();
+                if (airacCycle === null) {
```
felixscheffer: add empty line before if

src/assets/scripts/client/InputController.js
felixscheffer: done

src/assets/scripts/client/InputController.js
felixscheffer: done

felixscheffer:
I update the branch, addressed all my own review issues and did a quick test. As far as I can tell, everything works fine.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -114,6 +114,15 @@ export default class AirportModel {
          */
         this.wip = null;
 
+        /**
+         * This is what AIRAC cycle that the airport has been based off
```
n8rzz: This is a nitpick, but I think this could be worded better.  Just not sure how.

@erikquinn @Fechulo @indianbhaji thoughts?

Fechulo:
In relation to this PR, I think it would be good if we had a sim-wide 'airport review' every few months, where we make sure that all airports are up to date. Ideally, all the airports in the sim should be completely up to date. Procedures and such don't change very often so this 'review' would mostly consist of making sure nothing's changed and updating the AIRAC number in the airport file. In the case that a significant has been made, we should open a PR and deal with it as soon as possible. I don't know if we have enough man-power to do this every month, but doing it once every three months should suffice.

src/assets/scripts/client/airport/AirportModel.js
felixscheffer: I updated the description

erikquinn:
A very good idea! Maybe we can set that up with the calendar bot you have that posts to slack to remind us when that time comes? Maybe quartlerly at least?

src/assets/scripts/client/InputController.js
erikquinn: I've been told to use `if (!airacCycle)` in situations like this, in case we somehow end up with other falsey values like undefined.

src/assets/scripts/client/InputController.js
```diff
@@ -699,6 +699,17 @@ export default class InputController {
 
                 return true;
             }
+            case PARSED_COMMAND_NAME.AIRAC: {
+                const airacCycle = AirportController.getAiracCycle();
+
+                if (airacCycle === null) {
+                    UiController.ui_log('unknown airac cycle');
+                } else {
+                    UiController.ui_log(`airac cycle ${airacCycle}`);
```
erikquinn: AIRAC is an acronym that (except in variable names and such) should be capitalized. Stands for Aeronautical Information Regulation And Control.

src/assets/scripts/client/airport/AirportController.js
```diff
@@ -167,6 +167,15 @@ class AirportController {
         nextAirportModel.set(airportJson);
     }
 
+    /**
+     * @for InputController
+     * @property airac
```
erikquinn: `@method getAiracCycle`

src/assets/scripts/client/airport/AirportController.js
```diff
@@ -167,6 +167,15 @@ class AirportController {
         nextAirportModel.set(airportJson);
     }
 
+    /**
+     * @for InputController
```
erikquinn: `@for AirportController`

src/assets/scripts/client/constants/inputConstants.js
```diff
@@ -72,5 +72,6 @@ export const PARSED_COMMAND_NAME = {
     CLEAR: 'clear',
     AIRPORT: 'airport',
     RATE: 'rate',
-    TRANSMIT: 'transmit'
+    TRANSMIT: 'transmit',
+    AIRAC: 'airac'
```
erikquinn: Could you sort these alphabetically? I know it's not from this PR, but I'm not sure why they're in a big heap like this

src/assets/scripts/client/InputController.js
felixscheffer: done

src/assets/scripts/client/airport/AirportController.js
felixscheffer: done

src/assets/scripts/client/airport/AirportController.js
felixscheffer: done

src/assets/scripts/client/constants/inputConstants.js
felixscheffer: done

src/assets/scripts/client/InputController.js
felixscheffer: done

Fechulo:
@erikquinn Sounds good, I can add those in pretty easily. I'll try to make sure these reviews don't fall on end of school terms and such.

PR #1131, opened 2018-10-20
feature/1017 - Add more GA aircraft

### LINKED ISSUE #1017: Add more GA aircraft
I have noticed a few aircraft are missing when it comes to GA. I will be adding in a few more aircraft that will give more slow GA than the generic BE36 and cessnas that we currently have.

clscott42:
Resolves #1017.

Add more GA aircraft.


JPmAn24:
Got it. Taking over.

erikquinn:
@jpman24 do you have a specific goal in mind here? This seems like one that is very open ended and could go on forever since there's no shortage of GA aircraft that exist. If we can get a few done and approved in the next three days, we can get this into the next version, and then deal with any additional aircraft when we get around to it

JPmAn24:
@erikquinn I plan on also adding possibly a HondaJet, and also the Grumman Traveller, then we could merge this and move on to other things.

erikquinn:
Okay, try to get it done by Saturday and it can be merged by Sunday before the cutoff. :)

erikquinn:
And when it's ready, remember to request the review so I don't miss it.

PR #1132, opened 2018-10-20
refactor/981 - Add AircraftCollection class

### LINKED ISSUE #981: Create a proper AircraftCollection
`AircraftContorller#aircraft.list` is just not cutting it anymore...

clscott42:
Resolves #981.

Add `AircraftCollection` class to replace the basic array currently being used.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -7,6 +7,7 @@ import AirportController from '../airport/AirportController';
 import UiController from '../UiController';
 import EventBus from '../lib/EventBus';
 import AircraftTypeDefinitionCollection from './AircraftTypeDefinitionCollection';
+import AircraftCollection from './AircraftCollection';
 import AircraftModel from './AircraftModel';
 import AircraftConflict from './AircraftConflict';
 import StripViewController from './StripView/StripViewController';
@@ -141,6 +142,14 @@ export default class AircraftController {
 
         // TODO: this should its own collection class
         this.aircraft.list = [];
+
+        /**
+         * @property _collection
+         * @type {AircraftCollection}
+         * @private
+         */
+        this._collection = new AircraftCollection();
+
         this.aircraft.auto = { enabled: false };
         this.conflicts = [];
 
@@ -176,6 +185,7 @@ export default class AircraftController {
      * @chainable
      */
     _setupHandlers() {
+        this._onAddAircraftHandler = this._onAddAircraft.bind(this);
         this._onRemoveAircraftHandler = this.aircraft_remove.bind(this);
 
         return this;
@@ -187,7 +197,7 @@ export default class AircraftController {
      * @chainable
      */
     enable() {
-        this._eventBus.on(EVENT.ADD_AIRCRAFT, this.addItem);
+        this._eventBus.on(EVENT.ADD_AIRCRAFT, this._onAddAircraftHandler);
         this._eventBus.on(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickHandler);
         this._eventBus.on(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this.onSelectAircraftStrip);
         this._eventBus.on(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
@@ -203,7 +213,7 @@ export default class AircraftController {
      * @chainable
      */
     disable() {
-        this._eventBus.off(EVENT.ADD_AIRCRAFT, this.addItem);
+        this._eventBus.off(EVENT.ADD_AIRCRAFT, this._onAddAircraftHandler);
         this._eventBus.off(EVENT.STRIP_DOUBLE_CLICK, this._onStripDoubleClickHandler);
         this._eventBus.off(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, this._onSelectAircraftStrip);
         this._eventBus.off(EVENT.DESELECT_ACTIVE_STRIP_VIEW, this._onDeselectActiveStripView);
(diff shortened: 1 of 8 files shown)


n8rzz:
screen shot 2018-10-20 at 3 13 11 pm

Looks like the old api is in use in a few places, we should probably update those to use the collection.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -462,10 +461,12 @@ export default class AircraftController {
      *
      * @for AircraftController
      * @method removeAircraftModelFromList
-     * @param  {AircraftModel} aircraft the aircraft to remove
+     * @param  {AircraftModel} aircraftModel the aircraft to remove
      */
-    removeAircraftModelFromList(aircraft) {
-        this.aircraft.list = _without(this.aircraft.list, aircraft);
+    removeAircraftModelFromList(aircraftModel) {
+        console.warn('deprecated');
```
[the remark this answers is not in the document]
n8rzz: we should be a little more descriptive here, though.  I like the other comment further down, let's do something similar here too.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -462,10 +461,12 @@ export default class AircraftController {
...
+        this._collection.remove(aircraftModel);
+        this.aircraft.list = _without(this.aircraft.list, aircraftModel);
```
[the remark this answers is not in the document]

test/scope/_mocks/radarTargetMocks.js
```diff
@@ -1,16 +1,16 @@
...
-    ARRIVAL_AIRCRAFT_MODEL_MOCK,
-    DEPARTURE_AIRCRAFT_MODEL_MOCK
+    arrivalAircraftFixture,
+    departureAircraftFixture
 } from '../../aircraft/_mocks/aircraftMocks';
 import { THEME } from '../../../src/assets/scripts/client/constants/themes';
 
-export const RADAR_TARGET_ARRIVAL_MOCK = new RadarTargetModel(THEME.DEFAULT, ARRIVAL_AIRCRAFT_MODEL_MOCK);
-export const RADAR_TARGET_DEPARTURE_MOCK = new RadarTargetModel(THEME.DEFAULT, DEPARTURE_AIRCRAFT_MODEL_MOCK);
+export const RADAR_TARGET_ARRIVAL_MOCK = new RadarTargetModel(THEME.DEFAULT, arrivalAircraftFixture);
```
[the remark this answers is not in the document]
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftController.js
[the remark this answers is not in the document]

felixscheffer:
@n8rzz  can you provide some guidance on how to fix the api issues, please? Do we still want to keep AircraftContorller#aircraft.list?

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -466,10 +464,13 @@ export default class AircraftController {
      *
      * @for AircraftController
      * @method removeAircraftModelFromList
-     * @param  {AircraftModel} aircraft the aircraft to remove
+     * @param  {AircraftModel} aircraftModel the aircraft to remove
      */
-    removeAircraftModelFromList(aircraft) {
-        this.aircraft.list = _without(this.aircraft.list, aircraft);
+    removeAircraftModelFromList(aircraftModel) {
+        console.warn('AircraftController.removeAircraftModelFromList() has been deprecated');
```
n8rzz: so this is where my `new api` comment comes from.  The output says this has been deprecated, but I see we are calling `this._collection.remove(aircraftModel);` just below.  so is this method _really_ deprecated?  has it been replaced?  Will it be replaced?

If yes, then I would suggest that instead of doing the work with the collection here, this method would simply call the 'new' api method to do the work there.  Also, if this is going to be replace, I would do some finds to figure out where `.removeAircraftModelFromList()` is actually getting called and update those references as well.

test/aircraft/AircraftCollection.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+    collection.add(ARRIVAL_AIRCRAFT_MODEL_MOCK);
+
+    t.true(collection.length === 1);
+});
+
+ava('.remove() throws when the itemToRemove does not exist in `#_items`', (t) => {
+    const collection = new AircraftCollection();
+
+    collection.add(DEPARTURE_AIRCRAFT_MODEL_MOCK);
```
n8rzz: does the collection accept non-instantiated `AircraftModel` objects?  reading this test it seems like this is just data you're setting to the collection.

for a test like this, I'm cool with directly assigning to items and then running the assertion with `remove`

test/aircraft/AircraftCollection.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    collection.remove(DEPARTURE_AIRCRAFT_MODEL_MOCK);
+
+    t.true(collection.length === 0);
+});
+
+ava('.findAircraftByCallsign() returns null when no callsign is passed', (t) => {
+    const collection = new AircraftCollection();
+    const result = collection.findAircraftByCallsign();
+
+    t.falsy(result);
```
n8rzz: honestly not a fan of the `falsy` matcher.  in this case I'd prefer:
```js
t.true(typeof result === 'undefined');
```

remember tests are another form of documentation.  I'm already a fan of verbosity, be even more so with tests for that reason.

n8rzz:
as far as `AircraftController.aircraft.list` should totally go away in favor of the new Collection.

src/assets/scripts/client/aircraft/AircraftController.js
felixscheffer: Honestly, I don't know. According to the git history, you wrote the code for this PR 😉, so I hoped you know what to do here.

test/aircraft/AircraftCollection.spec.js
felixscheffer: done

felixscheffer:
Then there is plenty of work to do here.

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: ahh, so this is my breadcrumb then.

ok, with that in mind, I'd assume there is another method somewhere that is actually adding aircraft to the collection and should be used instead of this one.

PR #1136, opened 2018-10-20
refactor/696 - Remember preferred theme after refresh

### LINKED ISSUE #696: CanvasController loads DEFAULT theme on initial load
Discovered as part of #670, the `CanvasController` will _always_ instantiate with the `DEFAULT` theme.  There is an available `_setTheme()` method, but this is only called as a result of changing the selected option in the options menu.

clscott42:
Resolves #696.

### DIFF
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -21,6 +21,7 @@ import GameClockView from './game/GameClockView';
 import { speech_init } from './speech';
 import { EVENT } from './constants/eventNames';
 import { SELECTORS } from './constants/selectors';
+import { GAME_OPTION_NAMES } from './constants/gameOptionConstants';
 
 /**
  * Root controller class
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/AppController.js
```diff
@@ -147,7 +148,7 @@ export default class AppController {
 
         this.spawnPatternCollection = new SpawnPatternCollection(initialAirportData, this.navigationLibrary);
         this.spawnScheduler = new SpawnScheduler(this.spawnPatternCollection, this.aircraftController);
-        this.canvasController = new CanvasController(this.$element, this.navigationLibrary);
+        this.canvasController = new CanvasController(this.$element, this.navigationLibrary, GameController.getGameOption(GAME_OPTION_NAMES.THEME));
```
n8rzz: Since `GameController` is a singleton, there is no need to pass it here.  It can simply be referenced on assignment within the `CanvasController`

felixscheffer:
I merge develop but I haven't done any testing yet, so I have no idea if this is working as it should.

felixscheffer:
Should be working now...hopefully 🙈

src/assets/scripts/client/AppController.js
erikquinn: done

PR #1159, opened 2018-10-31
bugfix/1147 - Callsigns in tutorial should update when switching airports

### LINKED ISSUE #1147: Callsigns in tutorial should update when switching airports
#### Issue Description
The callsigns used in the tutorial dialogue are taken from the airport that's selected when the sim first loads. If the airport is switched, the tutorial keeps the old callsigns, making it harder to follow.

Reported by darren on Slack

#### Steps to reproduce
- Load sim and switch to another airport
- Open tutorial and navigate to third section (`Departing aircraft`)
- Example aircraft will not be present in the current airport

felixscheffer:
Resolves #1147

### DIFF
--- a/src/assets/scripts/client/tutorial/TutorialView.js
+++ b/src/assets/scripts/client/tutorial/TutorialView.js
@@ -207,12 +207,7 @@ export default class TutorialView {
      * @method onAirportChange
      */
     onAirportChange() {
-        const step = prop.tutorial.step;
-
         this.tutorial_init_pre();
-
-        prop.tutorial.step = step;
-
         this.tutorial_update_content();
     }
 
--- a/src/assets/scripts/client/tutorial/TutorialView.js
+++ b/src/assets/scripts/client/tutorial/TutorialView.js
@@ -87,6 +87,7 @@ export default class TutorialView {
         this.tutorial.open = false;
 
         this._init()
+            ._setupHandlers()
             .layout()
             .enable();
     }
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/tutorial/TutorialView.js
```diff
@@ -105,6 +105,7 @@ export default class TutorialView {
         this.$tutorialToggle = $(SELECTORS.DOM_SELECTORS.TOGGLE_TUTORIAL);
         this.$tutorialPrevious = this.$tutorialView.find(SELECTORS.DOM_SELECTORS.PREV);
         this.$tutorialNext = this.$tutorialView.find(SELECTORS.DOM_SELECTORS.NEXT);
+        this._onAirportChangeHandler = this.onAirportChange.bind(this);
```
n8rzz: you'll want to add a `._setupHandlers()` method that is called after `._init()` in the constructor.  Then move this assignment there.

src/assets/scripts/client/tutorial/TutorialView.js
```diff
@@ -182,6 +185,22 @@ export default class TutorialView {
...
     }
 
+    /**
+     * Reloads the tutorial when the airport is changed.
+     *
+     * @for tutorialView
+     * @method onAirportChange
+     */
+    onAirportChange() {
+        const step = prop.tutorial.step;
```
n8rzz: what is `step` here?  is this the `currentStep`?  meaning the current position in the tutorial?

src/assets/scripts/client/tutorial/TutorialView.js
felixscheffer: done

src/assets/scripts/client/tutorial/TutorialView.js
felixscheffer: I decided to remove that code as it causes new issues. If you are switching to another airport, the tutorial should now reset to the first page.

PR #1160, opened 2018-10-31
feature/898 - Add command to exit holding pattern

### LINKED ISSUE #898: Add command to exit holding pattern
We can exit holds by telling the aircraft to proceed direct to another fix further along its route, but what if we want them to go direct to the holding fix and resume via the last routing cleared? This is actually the _most common way_ to clear aircraft out of the hold, yet we aren't able to support it because we don't really have a command that allows the user to run `AircraftModel.pilot.exitHold()`.

Ideas for command: `nohold fixname`, or `xh fixname`? Those command names kinda stink, but nothing else comes to mind right now.

felixscheffer:
Resolves openscope/openscope#898

### DIFF
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -16,6 +16,7 @@
 [Routing Commands](#routing-commands)
 - [~~Fix~~](#fix)
 - [Hold](#hold)
+- [Exit Hold](#exit-hold)
 - [Proceed Direct](#proceed-direct)
 - [Route](#route)
 - [Reroute](#reroute)
@@ -159,6 +160,14 @@ with simply `[fixname]`. Any combination of these arguments provided in
 any order is acceptable, as long as the command `hold` comes first.
 
 _Syntax -_ `AAL123 hold [fixname] [left|right] [leg_time]min` or `AAL123 hold`
+
+### Exit Hold
+_Aliases -_ `exithold` `nohold` `xh`
+
+_Information -_ This command instructs the aircraft to leave a holding
+pattern and resume it's flight plan.
+
+_Syntax -_ `AAL123 exithold`
 
 ### Proceed Direct
 _Aliases -_ `direct` / `pd` / `dct`
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -165,7 +165,7 @@ _Syntax -_ `AAL123 hold [fixname] [left|right] [leg_time]min` or `AAL123 hold`
 _Aliases -_ `exithold` `nohold` `xh`
 
 _Information -_ This command instructs the aircraft to leave a holding
-pattern and resume it's flight plan.
+pattern and resume its flight plan.
 
 _Syntax -_ `AAL123 exithold`
 
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -162,12 +162,16 @@ any order is acceptable, as long as the command `hold` comes first.
 _Syntax -_ `AAL123 hold [fixname] [left|right] [leg_time]min` or `AAL123 hold`
 
 ### Exit Hold
-_Aliases -_ `exithold` `nohold` `xh`
+_Aliases -_ `cancelhold` `continue` `exithold` `nohold` `xh`
 
 _Information -_ This command instructs the aircraft to leave a holding
 pattern and resume its flight plan.
 
-_Syntax -_ `AAL123 exithold`
+_Parameters -_ Optionally, you may specify the fix at which to cancel the hold,
+by specifying the fix at the end of the command: `AAL123 cancelhold [fixname]`.
+This can be useful if multiple holds exist.
+
+_Syntax -_ `AAL123 continue` or `AAL123 xh BOTON`
 
 ### Proceed Direct
 _Aliases -_ `direct` / `pd` / `dct`
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -309,7 +309,6 @@ export default class AircraftCommander {
      * @for AircraftCommander
      * @method runExitHold
      * @param aircraft {AircraftModel}
-     * @param data {array}
      * @return {array} [success of operation, readback]
      */
     runExitHold(aircraft) {
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -307,15 +307,15 @@ export default class AircraftCommander {
      * Exit holding pattern and resume navigation
      *
      * @for AircraftCommander
-     * @method runExitHold
+     * @method runCancelHoldingPattern
      * @param aircraft {AircraftModel}
      * @param args {array}
      * @return {array} [success of operation, readback]
      */
-    runExitHold(aircraft, args) {
+    runCancelHoldingPattern(aircraft, args) {
         const fixName = args[0];
 
-        return aircraft.pilot.exitHold(fixName);
+        return aircraft.pilot.cancelHoldingPattern(fixName);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -300,6 +300,19 @@ export default class AircraftCommander {
         return aircraft.pilot.initiateHoldingPattern(fixName, holdParameters);
     }
 
+    /**
+     * Exit holding pattern and resume navigation
+     *
+     * @for AircraftCommander
+     * @method runExitHold
+     * @param aircraft {AircraftModel}
+     * @param data {array}
```
erikquinn: this param is not actually in the method definition, remove from docblock

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -551,6 +551,7 @@ export default class Pilot {
     *
     * @for Pilot
     * @method exitHold
+    * @return {array}                [success of operation, readback]
```
n8rzz: spacing looks a little strange here...

src/assets/scripts/client/info/AirportInfoController.js
```diff
@@ -0,0 +1,284 @@
+import $ from 'jquery';
+import { leftPad } from '../utilities/generalUtilities';
+import { radiansToDegrees } from '../utilities/unitConverters';
+import AirportController from '../airport/AirportController';
```
erikquinn: i'm confused? aren't these changes from a different PR?

src/assets/scripts/client/info/AirportInfoController.js
```diff
@@ -0,0 +1,284 @@
...
+    ELEVATION_VALUE: '.js-airportInfo-elevation-value'
+};
+
+/**
+ * Gets information about the current airport, specifically
+ * the airport's elevation, wind speed and direction, and altimeter.
+ *
+ * @class AirportInfoController
+ */
+export default class AirportInfoController {
```
n8rzz: Is this file supposed to be included in this PR?

src/assets/scripts/client/info/AirportInfoController.js
felixscheffer: whoops

src/assets/scripts/client/info/AirportInfoController.js
felixscheffer: no

src/assets/scripts/client/aircraft/Pilot/Pilot.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: done

test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -630,6 +628,25 @@ ava('.initiateHoldingPattern() returns correct readback when hold implemented su
...
+    const pilot = createPilotFixture();
+    const result = pilot.exitHold();
+
+    t.deepEqual(result, expectedResult);
+});
+
+ava('.exitHold() returns correct readback when the aircraft is holding', (t) => {
+    const expectedResult = [true, 'roger, cancelling hold over MISEN'];
+    const pilot = createPilotFixture();
+    pilot._fms.currentWaypoint.activateHold();
```
n8rzz: missing empty line after `const`

felixscheffer: done

felixscheffer:
@Marcel510 I added an entry to aircraft-commands.md. Feel free to extend or modify.

n8rzz:
@erikquinn with that in mind, would not having what you're asking for block this feature entirely?  Or is that work something that could happen after this is merged as part of a new issue?  perhaps we could call out the future work in the commands doc, so users are aware of the command's limitations?

felixscheffer:
While I think @erikquinn has a valid point, this feature isn't really high on my todo list. I accidentally stumbled across this ticket and thought it was easy and not a lot of work. If someone (@erikquinn?) wants to go ahead and extend this PR, feel free to. Otherwise, I would prefer to merge it as it is and create a follow-up ticket.

erikquinn:
Nah, it's just too much of an expectation if we introduce the feature that it would be able to do this. That use case is probably the majority of the time how this command would be used, so we are better off having nothing than half of it. So we can wait until somebody comes around to finishing it up.

felixscheffer:
@erikquinn I added an optional fix name parameter to the exithold command.

Ideally, we should add a test for canceling holds that the aircraft has not entered yet...

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -528,17 +528,27 @@ export default class Pilot {
...
-    exitHold() {
-        const currentWaypoint = this._fms.currentWaypoint;
+    exitHold(fixName) {
+        let waypoint = this._fms.currentWaypoint;
 
-        if (!currentWaypoint.isHoldWaypoint) {
+        if (fixName) {
+            waypoint = this._fms.findWaypoint(fixName);
+
+            if (!waypoint) {
```
n8rzz: let's move this `if` outside of the current `if`.  that'll give us a little extra defense

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -701,8 +711,8 @@ export default class Pilot {
             return problematicResponse;
         }
 
-        return [true, `hold ${cardinalDirectionFromFix} of ${fixName}, ` +
-            `${holdParameters.turnDirection} turns, ${holdParameters.legLength} legs`
+        return [true, `hold ${cardinalDirectionFromFix} of ${fixName}, `
```
n8rzz: we can just use multi-line strings here:
```js
 return [true, `
     hold ${cardinalDirectionFromFix} of ${fixName},
    ${holdParameters.turnDirection} turns, ${holdParameters.legLength} legs
`];
```

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandDefinitions.js
```diff
@@ -170,14 +164,20 @@ const SINGLE_ARG_AIRCRAFT_COMMANDS = {
         validate: singleArgumentValidator,
         parse: noop
     },
-    route: {
+    rate: {
         validate: singleArgumentValidator,
-        parse: noop
+        // calling method is expecting an array with values that will get spread later, thus we purposly
```
n8rzz: very nice 👍

src/assets/scripts/client/aircraft/Pilot/Pilot.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js
erikquinn: Multilining like that actually preserves the line breaks (how strange, right?), so that actually won't work. :disappointed:

erikquinn: @n8rzz if you'd like them some other way, just advise, i'm happy to change it :smiley:

documentation/aircraft-commands.md
```diff
@@ -162,12 +162,16 @@ any order is acceptable, as long as the command `hold` comes first.
 _Syntax -_ `AAL123 hold [fixname] [left|right] [leg_time]min` or `AAL123 hold`
 
 ### Exit Hold
```
felixscheffer: If we are going with  `cancel hold` we should change this to cancel hold as well

erikquinn: Sure, we can. cancel/exit/continue/etc are all still available aliases, so I thought it was a fine title for the user command documentation. Just found `.cancelHoldingPattern()` to be a clearer match to the opposite `.initiateHoldingPattern()`, compared to `.exitHold()`.

We can change the documentation though :+1:

PR #1180, opened 2018-11-08
enhancement/10 - Command for crossing restrictions (only altitude)

felixscheffer:
partially implements openscope/openscope#10 (only altitude)

Still todo:
- [x] Add tests

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -1948,7 +1948,7 @@ export default class AircraftModel {
 
         if (this.mcp.altitude < this.altitude) {
             // we want to descend...
-            if (!minimumAltitudeExists || altitudeMinimumWaypoint.altitudeMinimum <= this.mcp.altitude) {
+            if (!minimumAltitudeExists || altitudeMinimumWaypoint.altitudeMinimum < this.mcp.altitude) {
                 // ... and there is nothing that can stop us.
                 return this.mcp.altitude;
             }
@@ -1986,7 +1986,7 @@ export default class AircraftModel {
             }
         } else {
             // we want to climb...
-            if (!maximumAltitudeExists || this.mcp.altitude <= altitudeMaximumWaypoint.altitudeMaximum) {
+            if (!maximumAltitudeExists || this.mcp.altitude < altitudeMaximumWaypoint.altitudeMaximum) {
                 // ... and there is nothing that can stop us.
                 return this.mcp.altitude;
             }
@@ -2732,5 +2732,35 @@ export default class AircraftModel {
      */
     distanceToAircraft(anotherAircraftModel) {
         return this.positionModel.distanceToPosition(anotherAircraftModel.positionModel);
+    }
+
+    /**
+     * Ensure that the provided altitude is valid
+     *
+     * @for AircraftModel
+     * @method validateNextAltitude
+     * @param nextAltitude {number} altitude the aircraft should maintain
+     * @return {array|undefined}           [success of operation, readback]
+     */
+    validateNextAltitude(nextAltitude) {
+        if (nextAltitude === INVALID_NUMBER) {
+            const readback = {};
+            readback.log = 'unable, no altitude assigned';
+            readback.say = 'unable, no altitude assigned';
+
+            return [false, readback];
+        }
+
+        if (typeof nextAltitude !== 'number') {
+            return [false, `unable to maintain an altitude of ${nextAltitude}`];
+        }
+
+        if (!this.model.isAbleToMaintainAltitude(nextAltitude)) {
+            const readback = {};
+            readback.log = `unable to maintain ${nextAltitude} due to performance`;
+            readback.say = `unable to maintain ${radio_altitude(nextAltitude)} due to performance`;
+
+            return [false, readback];
+        }
     }
 }
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -2744,11 +2744,7 @@ export default class AircraftModel {
      */
     validateNextAltitude(nextAltitude) {
         if (nextAltitude === INVALID_NUMBER) {
-            const readback = {};
-            readback.log = 'unable, no altitude assigned';
-            readback.say = 'unable, no altitude assigned';
-
-            return [false, readback];
+            return [false, 'unable, no altitude assigned'];
         }
 
         if (typeof nextAltitude !== 'number') {
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -2754,7 +2754,7 @@ export default class AircraftModel {
      * @param nextAltitude {number} altitude the aircraft should maintain
      * @return {array}           [success of operation, readback]
      */
-    validateNextAltitude(nextAltitude) {
+    validateNextAltitude(nextAltitude, airportModel) {
         if (nextAltitude === INVALID_NUMBER) {
             return [false, 'unable, no altitude assigned'];
         }
@@ -2767,6 +2767,15 @@ export default class AircraftModel {
             const readback = {};
             readback.log = `unable to maintain ${nextAltitude} due to performance`;
             readback.say = `unable to maintain ${radio_altitude(nextAltitude)} due to performance`;
+
+            return [false, readback];
+        }
+
+        if (nextAltitude < airportModel.minAssignableAltitude) {
+            const minimumAltitude = airportModel.minAssignableAltitude;
+            const readback = {};
+            readback.log = `unable to maintain ${nextAltitude}, the MSA is ${minimumAltitude}`;
+            readback.say = `unable to maintain ${radio_altitude(nextAltitude)}, the MSA is ${radio_altitude(minimumAltitude)}`;
 
             return [false, readback];
         }
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -696,7 +696,7 @@ export default class AircraftModel {
      * @return cs {string}
      */
     getRadioCallsign() {
-        let weight = this.getRadioWeightClass();
+        let weight = this.model.getRadioWeightClass();
 
         if (!_isEmpty(weight)) {
             weight = ` ${weight}`;
@@ -709,26 +709,6 @@ export default class AircraftModel {
         return `${this.airlineCallsign} ${groupNumbers(this.flightNumber)}${weight}`;
     }
 
-    /**
-     * Get the weight classifier for an aircraft's callsign, as spoken over the radio
-     *
-     * @for AircraftModel
-     * @method getRadioWeightClass
-     * @return {string}
-     */
-    getRadioWeightClass() {
-        const { weightClass } = this.model;
-
-        if (weightClass === 'H') {
-            return 'heavy';
-        }
-
-        if (weightClass === 'J') {
-            return 'super';
-        }
-
-        return '';
-    }
 
     // TODO: this method should move to the `AircraftTypeDefinitionModel`
     /**
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -717,11 +717,13 @@ export default class AircraftModel {
      * @return {string}
      */
     getRadioWeightClass() {
-        const weightClass = this.model.weightClass;
+        const { weightClass } = this.model;
 
         if (weightClass === 'H') {
             return 'heavy';
-        } else if (weightClass === 'J') {
+        }
+
+        if (weightClass === 'J') {
             return 'super';
         }
 
@@ -738,9 +740,9 @@ export default class AircraftModel {
         let serviceCeilingClimbRate;
         let cr_uncorr;
         let cr_current;
-        const altitude = this.altitude;
+        const { altitude } = this;
         const rate = this.model.rate.climb;
-        const ceiling = this.model.ceiling;
+        const { ceiling } = this.model;
 
         if (this.model.engines.type === 'J') {
             serviceCeilingClimbRate = 500;
@@ -752,7 +754,8 @@ export default class AircraftModel {
         // in troposphere
         if (this.altitude < 36152) {
             // TODO: break this assignemnt up into smaller parts and holy magic numbers! enumerate the magic numbers
-            cr_uncorr = rate * 420.7 * ((1.232 * Math.pow((518.6 - 0.00356 * altitude) / 518.6, 5.256)) / (518.6 - 0.00356 * altitude));
+            cr_uncorr = rate * 420.7 * ((1.232 * (((518.6 - 0.00356 * altitude) / 518.6) ** 5.256))
+                / (518.6 - 0.00356 * altitude));
             cr_current = cr_uncorr - (altitude / ceiling * cr_uncorr) + (altitude / ceiling * serviceCeilingClimbRate);
         } else {
             // in lower stratosphere
@@ -831,7 +834,7 @@ export default class AircraftModel {
         const waypointSpeed = waypointModel.speedMaximum;
         const waypointDistance = this.positionModel.distanceToPosition(waypointModel.positionModel);
         const speedChange = waypointSpeed - this.speed;
-        const decelerationRate = -this.model.rate.decelerate / 2;   // units of rate.decel are 'knots per 2 seconds'
+        const decelerationRate = -this.model.rate.decelerate / 2; // units of rate.decel are 'knots per 2 seconds'
         const decelerationTime = speedChange / decelerationRate;
         const timeUntilWaypoint = waypointDistance / this.groundSpeed * TIME.ONE_HOUR_IN_SECONDS;
 
@@ -1055,9 +1058,9 @@ export default class AircraftModel {
      * @method isTaxiing
      */
     isTaxiing() {
-        return this.flightPhase === FLIGHT_PHASE.APRON ||
-            this.flightPhase === FLIGHT_PHASE.TAXI ||
-            this.flightPhase === FLIGHT_PHASE.WAITING;
+        return this.flightPhase === FLIGHT_PHASE.APRON
+            || this.flightPhase === FLIGHT_PHASE.TAXI
+            || this.flightPhase === FLIGHT_PHASE.WAITING;
     }
 
     // TODO: The function description and what it actually does do not match
@@ -1548,7 +1551,6 @@ export default class AircraftModel {
 
             default:
                 break;
-
         }
     }
 
@@ -1599,8 +1601,8 @@ export default class AircraftModel {
                 return this._calculateTargetedHeadingToInterceptCourse();
 
             default:
-                console.warn('Expected MCP heading mode of "OFF", "HOLD", "LNAV", or "VOR", ' +
-                    `but received "${this.mcp.headingMode}"`);
+                console.warn('Expected MCP heading mode of "OFF", "HOLD", "LNAV", or "VOR", '
+                    + `but received "${this.mcp.headingMode}"`);
                 return this.heading;
         }
     }
@@ -1642,8 +1644,8 @@ export default class AircraftModel {
             }
 
             default:
-                console.warn('Expected MCP speed mode of "OFF", "HOLD", "LEVEL_CHANGE", "N1", or "VNAV", but ' +
-                    `received "${this.mcp[MCP_MODE_NAME.SPEED]}"`);
+                console.warn('Expected MCP speed mode of "OFF", "HOLD", "LEVEL_CHANGE", "N1", or "VNAV", but '
+                    + `received "${this.mcp[MCP_MODE_NAME.SPEED]}"`);
                 return this._calculateLegalSpeed(this.speed);
         }
     }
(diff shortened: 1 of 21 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2753,4 +2753,30 @@ export default class AircraftModel {
...
+
+    /**
+     * Ensure that the provided altitude is valid
+     *
+     * @for AircraftModel
+     * @method validateNextAltitude
+     * @param nextAltitude {number} altitude the aircraft should maintain
+     * @return {array|undefined}           [success of operation, readback]
+     */
+    validateNextAltitude(nextAltitude) {
```
n8rzz: I like where your head is head here, however, this introduces some un-necessary complexity because of the variadic return.  at the very least, the return of this method should be consistent.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2753,4 +2753,30 @@ export default class AircraftModel {
...
+            return [false, `unable to maintain an altitude of ${nextAltitude}`];
+        }
+
+        if (!this.model.isAbleToMaintainAltitude(nextAltitude)) {
+            const readback = {};
+            readback.log = `unable to maintain ${nextAltitude} due to performance`;
+            readback.say = `unable to maintain ${radio_altitude(nextAltitude)} due to performance`;
+
+            return [false, readback];
+        }
```
n8rzz: continuing off the last comment, this method should always return the same value type.  please add:
```js
return [];
```

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -116,15 +117,10 @@ export default class Pilot {
...
-            const verbalRequestedAltitude = radio_altitude(altitude);
-            const readback = {};
-            readback.log = `unable to maintain ${altitude} due to performance`;
-            readback.say = `unable to maintain ${verbalRequestedAltitude} due to performance`;
-
-            this._mcp.setAltitudeFieldValue(aircraftModel.altitude);
+        const response = aircraftModel.validateNextAltitude(altitude);
 
-            return [false, readback];
+        if (response && !response[0]) {
```
n8rzz: if the above method is update to always return an array, then this if becomes much easier:
```
if (response.length > 0) {
    return response;
}
```

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -461,24 +457,10 @@ export default class Pilot {
...
-        }
+        const response = aircraftModel.validateNextAltitude(nextAltitude);
 
-        if (!aircraftModel.model.isAbleToMaintainAltitude(nextAltitude)) {
-            const readback = {};
-            readback.log = `unable to maintain ${nextAltitude} due to performance`;
-            readback.say = `unable to maintain ${radio_altitude(nextAltitude)} due to performance`;
-
-            return [false, readback];
+        if (response && !response[0]) {
```
n8rzz: same principle here, if `.validateNextAltitude()` always returns an array this if is much simpler.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -546,6 +528,47 @@ export default class Pilot {
...
+    cross(aircraft, fixName, altitude) {
+        if (!NavigationLibrary.hasFixName(fixName)) {
+            return [false, `unknown fix '${fixName}'`];
+        }
+
+        if (!this._fms.hasWaypointName(fixName)) {
+            return [false, `fix '${fixName}' is not on our route`];
+        }
+
+        const response = aircraft.validateNextAltitude(altitude);
```
n8rzz: same principle here, if `.validateNextAltitude()` always returns an array this if is much simpler.

src/assets/scripts/client/parsers/aircraftCommandParser/argumentParsers.js
```diff
@@ -209,3 +209,19 @@ export const timewarpParser = (args = []) => {
         convertStringToNumber(args[0])
     ];
 };
+
+/**
+ * the `timewarp` command needs to be able to provide a default value,
```
n8rzz: update docblock

src/assets/scripts/client/parsers/aircraftCommandParser/argumentParsers.js
```diff
@@ -209,3 +209,19 @@ export const timewarpParser = (args = []) => {
...
+ * this parser allows us to do that.
+ *
+ * @method crossingParser
+ * @param  {array} [args=[]]
+ * @return {array<string, number>}
+ */
+export const crossingParser = (args = []) => {
+    const fix = args[0];
+    // args[1] should be 'at' or '@'
```
n8rzz: this information should be added to the docblock.

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
```diff
@@ -279,3 +279,35 @@ export const squawkValidator = (args = []) => {
...
+
+    if (args[1] !== 'at' && args[1] !== '@') {
+        return ERROR_MESSAGE.AT;
+    }
+
+    const altitude = convertStringToNumber(args[2]);
+
+    if (_isNaN(altitude)) {
+        return ERROR_MESSAGE.ALTITUDE_MUST_BE_NUMBER;
+    }
```
n8rzz: I can't remember, do all our validators return `undefined` when and if they are successful?

test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -409,6 +401,29 @@ ava('.climbViaSID() correctly configures MCP and returns correct response when a
     t.true(pilot._mcp.altitude === 11000);
 });
 
+ava('.cross() correctly configures MCP and returns correct response ', (t) => {
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const pilot = createPilotFixture();
```
n8rzz: I'm going to nit-pick on naming a little bit here:

* things that are fixtures should be suffixed with the word `Fixture`.  so `pilot` -> `pilotFixture`
* things that are mocks, should similarly be suffixed with the word `Mock`.  so `fixName` -> `fixNameMock` and `altitude` -> `altitudeMock`

My guess is that this convention isn't followed everywhere, which is ok for now.  Going forward I want to try and be consistent.  As I'm sure you've discovered already, I tend to lean on the side of verbosity.

test/parsers/aircraftCommandParser/argumentParser.spec.js
```diff
@@ -157,3 +158,11 @@ ava('.timewarpParser() returns an array with 50 as a value when provided as an a
 
     t.true(result[0] === 50);
 });
+
+
+ava('.crossingParser() returns an array with the correct values', (t) => {
```
n8rzz: This is the happy path test, which is good, but we should also have un-happy and/or WTF path tests in here too. 
 could you add at least one negative test case for this?

test/parsers/aircraftCommandParser/argumentValidator.spec.js
```diff
@@ -136,7 +137,7 @@ ava('.altitudeValidator() returns a string when passed anything other than exped
 });
 
 ava('.optionalAltitudeValidator() returns undefined when no value is passed', t => {
-    let result = optionalAltitudeValidator([]);
+    const result = optionalAltitudeValidator([]);
     t.true(typeof result === 'undefined');
```
n8rzz: missing empty line after `const`

test/parsers/aircraftCommandParser/argumentValidator.spec.js
```diff
@@ -295,3 +296,55 @@ ava('.squawkValidator() returns string when passed invalid squawk', t => {
     result = squawkValidator(['1a11']);
     t.true(result === 'Invalid argument. Expected \'0000\'-\'7777\' for the transponder code.');
 });
+
+ava('.crossingValidator() returns a string when passed the wrong number of arguments', t => {
```
n8rzz: this whole set of tests is very nice.  well done! 👍

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
felixscheffer: yes

test/parsers/aircraftCommandParser/argumentValidator.spec.js
felixscheffer: :heart_eyes:

src/assets/scripts/client/aircraft/AircraftModel.js

src/assets/scripts/client/aircraft/AircraftModel.js

src/assets/scripts/client/aircraft/Pilot/Pilot.js
felixscheffer: done

src/assets/scripts/client/aircraft/Pilot/Pilot.js
felixscheffer: done

src/assets/scripts/client/aircraft/Pilot/Pilot.js
felixscheffer: done

test/parsers/aircraftCommandParser/argumentValidator.spec.js
felixscheffer: Technically I agree with you but the rest of the file follows the same pattern. I would like to keep the file consistent so we either stick with the current pattern or change the whole file. Let me know which way you prefer.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -484,6 +484,18 @@ export default class Fms {
...
 
+    /**
+     * Return the waypoint that matches the provided name
+     *
+     * @for Fms
+     * @method findWaypoiont
+     * @param waypointName {string} name of the waypoint
+     * @return {WaypointModel}
+     */
+    findWaypoint(waypointName) {
```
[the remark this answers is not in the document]
felixscheffer: There is no `findWaypointByName` in Fms or in RouteModel. If there is actually a method which returns a waypoint by its name, then I can't find it.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2753,4 +2753,30 @@ export default class AircraftModel {
...
         }
     }
+
+    /**
+     * Ensure that the provided altitude is valid
+     *
+     * @for AircraftModel
+     * @method validateNextAltitude
+     * @param nextAltitude {number} altitude the aircraft should maintain
+     * @return {array|undefined}           [success of operation, readback]
```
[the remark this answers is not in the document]
felixscheffer: `runAltitude` is not doing any validation at all. `validateNextAltitude` is basically what `maintainAltitude` used to do. Maybe some of it is redundant and is already done by the argument validator. The idea was to do the same validation as `maintainAltitude` (without copy-n-pasting code).

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -546,6 +528,47 @@ export default class Pilot {
...
+     * Cross a fix at a certain altitude
+     *
+     * @for Pilot
+     * @method cross
+     * @param aircraft {AircraftModel}
+     * @param fixName {string} name of the fix
+     * @param altitude {number} the altitude
+     * @return {array}                [success of operation, readback]
+     */
+    cross(aircraft, fixName, altitude) {
```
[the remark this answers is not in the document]
felixscheffer: renamed to `crossFix`. Hopefully, we'll add speed in the near future.

test/parsers/aircraftCommandParser/argumentValidator.spec.js
n8rzz: Works for me, let’s leave it!

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -546,6 +528,47 @@ export default class Pilot {
...
+        const response = aircraft.validateNextAltitude(altitude);
+
+        if (response && !response[0]) {
+            return response;
+        }
+
+        const waypoint = this._fms.findWaypoint(fixName);
+
+        waypoint.altitudeMaximum = altitude;
+        waypoint.altitudeMinimum = altitude;
```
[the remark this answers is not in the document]
felixscheffer: In Java I would agree but in Javascript? What's wrong with the set-syntax feature? I can barely see a test case for `setMaximumAltitude`. Maybe I am missing something here.

@n8rzz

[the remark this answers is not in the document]
n8rzz: There isn't actually anything wrong with doing it this way, however, we prefer a setter method just so the modification of properties is as consolidated as possible.  The way it's written now is totally valid, but, it makes it harder to track down if there are multiple places setting a property.  By consolidating that setter, we always know where the property is actually being changed.

test/aircraft/Pilot/Pilot.spec.js
felixscheffer: I absolutely dislike those verbose names. IMHO they don't provide any value and make code harder to read. It's like a modern version of the Hungarian notation. 

done...

test/parsers/aircraftCommandParser/argumentParser.spec.js
felixscheffer: No idea what to test for. Suggestions?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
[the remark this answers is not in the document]
erikquinn: I guess it doesn't exist, so this is fine

src/assets/scripts/client/aircraft/Pilot/Pilot.js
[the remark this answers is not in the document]
erikquinn: It's mostly because we want to protect the property from messed up values. If there's a setting method which checks the values for validity, then we can avoid some really whacky bugs down the line. I know we're validating the altitude a few lines up, but in general we avoid direct assignments like the plague-- especially when making those changes outside of the class being modified.

erikquinn:
Made some minor adjustments, but still needs tests.

test/parsers/aircraftCommandParser/argumentParser.spec.js
felixscheffer: We run a validator before the parser. Ideally, it catches all the invalid values. In case of #1249, it doesn't.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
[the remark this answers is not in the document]
felixscheffer: @erikquinn Have you fixed this one? i.e. can we mark it as resolved or is there something left to do here?

felixscheffer:
@erikquinn are you taking care of the tests?

felixscheffer:
@erikquinn No problem. I just wanted to make sure someone is taking care of this PR.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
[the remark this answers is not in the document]
erikquinn: The methods have been added. Resolving.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -717,11 +717,13 @@ export default class AircraftModel {
      * @return {string}
      */
     getRadioWeightClass() {
```
n8rzz: isn't this implemented already in the `AircraftTypeModel`?

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -116,20 +128,16 @@ export default class Pilot {
...
 
-            return [false, readback];
+        if (!response[0]) {
+            return response;
         }
 
         const currentAltitude = aircraftModel.altitude;
-        const { minAssignableAltitude, maxAssignableAltitude } = airportModel;
-        let clampedAltitude = clamp(minAssignableAltitude, altitude, maxAssignableAltitude);
+        let clampedAltitude = airportModel.clampWithinAssignableAltitudes(altitude);
```
n8rzz: 😌

test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -619,14 +638,97 @@ ava('.conductInstrumentApproach() returns a success message', (t) => {
...
+    t.deepEqual(result, expectedResult);
+});
+
+ava('.crossFix() returns early when the specified altitude fails .validateNextAltitude()', (t) => {
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const fixNameMock = 'kepec';
+    const altitudeMock = 13000;
+    const expectedResult = [false, 'some reply from validate-next-altitude'];
+
+    sinon.stub(aircraftModel, 'validateNextAltitude', () => expectedResult);
```
n8rzz: could we use the `returns` syntax here instead? 
```suggestion
    sinon.stub(aircraftModel, 'validateNextAltitude').returns(expectedResult);
```

I find, as usual, this is a little easier to read

test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -619,14 +638,97 @@ ava('.conductInstrumentApproach() returns a success message', (t) => {
...
+    const expectedResult = [true, {
+        log: 'cross COWBY at 13000',
+        say: 'cross cowby at one three thousand'
+    }];
+    const result = aircraftModel.pilot.crossFix(aircraftModel, fixNameMock, altitudeMock);
+
+    t.deepEqual(result, expectedResult);
+});
+
+// ava('.crossFix() correctly configures MCP and returns correct response ', (t) => {
```
n8rzz: do we plan to refactor this test in the future or is this no longer needed?  if the former, lets un-comment it but list it as skip: `ava.skip()`.  If the later, lets remove it.

src/assets/scripts/client/aircraft/AircraftModel.js
felixscheffer: I don't think so but I feel like this method should be part of `AircraftTypeDefinitionModel`

test/aircraft/Pilot/Pilot.spec.js
felixscheffer: If we need that test we should write it...instead of skipping it 😉

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: It was not. I agree I suppose, but the other `.getRadio()` stuff is in the AircraftModel, right next to this method, so it's okay to leave there for now I think.

test/aircraft/Pilot/Pilot.spec.js
erikquinn: works for me! done

test/aircraft/Pilot/Pilot.spec.js
erikquinn: nope, it's left over copy pasta 😃 all gone now.

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -475,6 +473,63 @@ export default class WaypointModel {
         this._holdParameters.timer = DEFAULT_HOLD_PARAMETERS.timer;
     }
 
+    // FIXME: test
```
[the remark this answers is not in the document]
erikquinn: Yes! Don't know how I missed that.... done!

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -475,6 +473,63 @@ export default class WaypointModel {
...
+
+        if (altitudeMaximum < 0 || altitudeMaximum > 60000) {
+            console.warn(`Expected requested waypoint '${this._name}' max altitude to be reasonable, `
+                + `but received altitude altitude of '${altitudeMaximum}'`);
+        }
+
+        this.altitudeMaximum = altitudeMaximum;
+    }
+
+    // FIXME: test
```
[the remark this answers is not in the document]
erikquinn: Yes! Don't know how I missed that.... done!

src/assets/scripts/client/aircraft/FlightManagementSystem/WaypointModel.js
```diff
@@ -475,6 +473,63 @@ export default class WaypointModel {
...
+     * @for WaypointModel
+     * @method setAltitude
+     * @param altitude {number} in feet
+     */
+    setAltitude(altitude) {
+        this.setAltitudeMaximum(altitude);
+        this.setAltitudeMinimum(altitude);
+    }
+
+    // FIXME: test
```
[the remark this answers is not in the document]
erikquinn: Yes! Don't know how I missed that.... done!

src/assets/scripts/client/aircraft/AircraftModel.js
felixscheffer: moved to AircraftTypeDefinitionModel and added tests

felixscheffer:
:shipit:

PR #1185, opened 2018-11-11
enhancement/604 - Add timeshare to data block

### LINKED ISSUE #604: Make data block STARs style with timeshare
Improve upon the current data blocks by adding what they call the "timeshare". The second line of the data block rotates between displaying two sets of information: the altitude and speed, and then the scratchpad and aircraft type.

The scratchpad only supports three alphanumeric characters, and should be defaulted to showing the aircraft destination. If departures could show their first fix, that would be handy, but more thought would be needed on how to achieve this given the various situations aircraft can be in, as we continue our quest to minimize how often we consider any given airplane an `arrival` or `departure`, since those terms are relative.

n8rzz:
Resolves #604 

Adds timeshare capability to data block


felixscheffer:
@n8rzz  is this still work in progress?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1399,12 +1412,16 @@ export default class CanvasController {
...
+        const gap = 3; // height of TOTAL vertical space between the rows (0 for touching)
         const lineheight = 4.5; // height of text row (used for spacing basis)
         const row1text = callsign;
         const aircraftAltitude = round(aircraftModel.altitude * 0.01);
         const aircraftSpeed = round(aircraftModel.groundSpeed * 0.1);
-        const row2text = `${leftPad(aircraftAltitude, 3)} ${leftPad(aircraftSpeed, 2)}`;
+        let row2text = `${leftPad(aircraftAltitude, 3)} ${leftPad(aircraftSpeed, 2)}`;
+
+        if (this.shouldShowSecondaryDataBlock) {
+            row2text = `${aircraftModel.model.icao.toUpperCase()} ${aircraftModel.fms.currentWaypoint.name}`;
```
felixscheffer: The ticket says "the scratchpad and aircraft type". The implementation is "aircraft type - scratchpad". @erikquinn

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +306,72 @@ export default class RadarTargetModel {
...
+
+    /**
+     * Generate a string to be used for the second row of a datablock
+     * when the timeshare section is active
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowTwoTimeshare
+     * @returns {string}
+     */
+    buildDataBlockRowTwoTimeshare() {
```
erikquinn: So there's `.buildDataBlockRowTwo()`, which calls `.buildDataBlockRowTwoTimeshare()`, making it seem like we're only "showing the timeshare" when the latter is called, which is an odd implication. For clarity, the second line _itself_ is the "timeshare", and it has a primary set of data and a secondary set, and the second row displays each of them in turn. You have it all working correctly, but I would suggest something more like `.buildDataBlockRowTwoWithPrimaryInfo()` and `.buildDataBlockRowTwoWithSecondaryInfo()` or similar. That way, it's more clear that the second row will be popluated by one or the other of these functions, without adding the inaccurate implication that the secondary info is the "timeshare".

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -520,6 +520,21 @@ export default class AircraftModel {
...
     }
 
+    /**
+     * Returns the appropriate string to use in
+     * row one of a data block.
+     *
+     * @property timeShareDestination
+     * @type {string}
+     */
+    get timeShareDestination() {
```
erikquinn: This element of the timeshare is actually called the `scratchpad`. It defaults to showing the values you've provided, but can also be used by the controller to input their own value, within the character and length limitations. The scratchpad is already plugged in and more or less set up to do exactly what this function does anyway, if you could just merge the two together.

Also, we are missing a `.toUpperCase()` when the destination airport is being carried into the data block.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +306,72 @@ export default class RadarTargetModel {
...
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowTwo
+     * @returns {string}
+     */
+    buildDataBlockRowTwo() {
+        const aircraftAltitude = round(this.aircraftModel.altitude * 0.01);
+        const aircraftSpeed = round(this.aircraftModel.groundSpeed * 0.1);
+
+        return `${leftPad(aircraftAltitude, 3)} ${leftPad(aircraftSpeed, 2)}`;
```
erikquinn: An extra space is needed between these elements since the altitude length is 3. The timeshare has the first set of 4 characters, a space, then the second set of four characters. In both the primary and secondary sets, the second elements should not move left to right during the transition.
```
|---|---|
110  29
KSEA B738
|---|---|
```

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +306,72 @@ export default class RadarTargetModel {
...
+    /**
+     * Generate a string to be used for the second row of a datablock
+     * when the timeshare section is active
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowTwoTimeshare
+     * @returns {string}
+     */
+    buildDataBlockRowTwoTimeshare() {
+        return `${this.aircraftModel.timeShareDestination} ${this.aircraftModel.model.icao.toUpperCase()}`;
```
erikquinn: The first element here (the "scratchpad") must be trimmed or padded to be 4 characters exactly to maintain the appropriate spacing.
```
|---|---|
R2   A320
KSEA B738
PAE  B712
|---|---|
```

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -213,6 +212,19 @@ export default class CanvasController {
...
+     *
+     * This provides a way to know when to show the primary
+     * dataBlock or the secondary dataBlock
+     *
+     * @property shouldShowSecondaryDataBlock
+     * @type {boolean}
+     */
+    get shouldShowSecondaryDataBlock() {
+        return _inRange(TimeKeeper.gameTimeMilliseconds % 3000, 2000, 3000);
+    }
```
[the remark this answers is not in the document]
n8rzz: whoops, was originally getter-able but not anymore.  done

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1291,135 +1271,188 @@ export default class CanvasController {
...
+
+    /**
+     * Draw the legacy dataBlock
+     *
+     * @for CanvasController
+     * @method
+     * @param {HTML CanvasContext} cc
+     * @param {AircraftModel} RadarTargetModel#aircraftModel
+     */
+    _drawLegacyDatablock(cc, aircraftModel) {
```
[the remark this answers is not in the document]
n8rzz: completely agree, except for the legacy data block.  A 'modern' data block is simply a collection of text displayed in a grid.  The `legacy` data block includes much more drawing logic to achieve the half circles and circle fills, etc.  That is the only data block that's going to look that way and it has a crapload more logic involved.  Thus, it gets it's own method.

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
```diff
@@ -101,6 +101,30 @@ export const DATA_BLOCK_THEME = {
...
+     */
+    HALF_WIDTH: 30,
+
+    /**
+     * Height of a data block
+     *
+     * @memberof DATA_BLOCK_THEME
+     * @property HEIGHT
+     */
+    HEIGHT: 32,
```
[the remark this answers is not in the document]
n8rzz: The goal here was to avoid calculations.  Remember, these methods live inside the loop which means they are going to fire up to 60 times *per second*.  If we are calculating that every frame that gets expensive. 

That said, this doesn't look good.  I'll look into it.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +306,72 @@ export default class RadarTargetModel {
...
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowOne
+     * @returns {string}
+     */
+    buildDataBlockRowOne() {
+        let dataBlockRowOne = `${this.aircraftModel.callsign}`;
+
+        if (this.aircraftModel.model.weightclass === 'H') {
+            dataBlockRowOne = `${this.aircraftModel.callsign}  ${this.aircraftModel.model.weightclass}`;
```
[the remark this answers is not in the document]
n8rzz: yes, it is needlessly specific.  I'm sure I had a good reason for it at the time though I can't figure what that reason would have been. ☹️

src/assets/scripts/client/scope/RadarTargetModel.js
n8rzz: I figured I jacked up the words here.  done

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +306,72 @@ export default class RadarTargetModel {
...
+    /**
+     * Generate a string to be used for the second row of a datablock
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowTwo
+     * @returns {string}
+     */
+    buildDataBlockRowTwo() {
+        const aircraftAltitude = round(this.aircraftModel.altitude * 0.01);
+        const aircraftSpeed = round(this.aircraftModel.groundSpeed * 0.1);
```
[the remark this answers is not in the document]
n8rzz: copy pasta from previous implementation

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: The only references I'm seeing to `scratchpad` as an existing concept are within the changelog and a breif item in the command reference doc about it not being implemented?

`.toUpperCase()` is done

src/assets/scripts/client/canvas/CanvasController.js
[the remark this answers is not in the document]
erikquinn: Okay, that sounds fine.

src/assets/scripts/client/constants/themes/default/modules/dataBlock.js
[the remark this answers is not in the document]
n8rzz: done.  

In my opinion, we should do away with this legacy data block entirely some day.  it's a lot of extra logic for, what I'm guessing is, not a lot of use.  We'll see what comes of #1197 and go from there.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: `RadarTargetModel._scratchPadText`. It's currently already being initialized to `aircraftModel.destination.substr(1)` by the RadarTargetModel. And it can be changed (including by the user) in `RadarTargetModel.setScratchpad()`.

We should use the scratchpad in lieu of this getter.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -521,6 +521,21 @@ export default class AircraftModel {
...
+        if (this.isArrival()) {
+            return this.destination.toUpperCase();
+        }
+
+        return this.fms.currentWaypoint.name.slice(0, 3);
```
erikquinn: This should be constantly showing the departure procedure or first fix in the route. Even after passing the end of the SID, or that first fix, or whatever-- it should continue to show that as its value. So set it and forget it.

A `this.fms.getFullRouteString()` method will need to be added, and then we pull the SID exit (or when absent, the first fix) from that route string.

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1223,53 +1238,18 @@ export default class CanvasController {
...
         if (prop.input.callsign.length > 0 && aircraftModel.matchCallsign(prop.input.callsign)) {
             match = true;
         }
 
-        // set color, intensity, and style elements
-        let red = this.theme.DATA_BLOCK.ARRIVAL_BAR_OUT_OF_RANGE;
-        let green = this.theme.DATA_BLOCK.BACKGROUND_OUT_OF_RANGE;
-        let blue = this.theme.DATA_BLOCK.DEPARTURE_BAR_OUT_OF_RANGE;
-        let white = this.theme.DATA_BLOCK.TEXT_OUT_OF_RANGE;
+        let white = aircraftModel.isControllable ? this.theme.DATA_BLOCK.TEXT_IN_RANGE : this.theme.DATA_BLOCK.TEXT_OUT_OF_RANGE;
```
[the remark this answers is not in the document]
n8rzz: Agreed, these color names should be updated and are from the original implementation, however, not in this PR.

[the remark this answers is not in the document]
n8rzz: ternaries...

They are still the devil.  In this case, though, it actually makes things easier to read and understand.  I still avoid them as much as possible, just not as much as I have in years past.  They sometimes have a place.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: Correct, it doesn't go here, but wherever the scratchpad is initialized for a departing aircraft is where this would be applied to get the desired behavior.

Currently, I think it initializes to the icao of destination airport, or `'XXX'` when the destination airport doesn't exist.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +307,77 @@ export default class RadarTargetModel {
...
 
+    /**
+     * Generate a string to be used for the first row of a datablock
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowOne
+     * @returns {string}
+     */
+    buildDataBlockRowOne() {
+        let dataBlockRowOne = `${this.aircraftModel.callsign}`;
```
felixscheffer: I don't understand this line? `callsign` should already be a string, right?

n8rzz: mmm, yeah.  that's a good catch.  I'll update that.

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -473,6 +484,29 @@ export default class RouteModel extends BaseModel {
...
+     * @for RouteModel
+     * @method getDepartureExitView
+     * @returns {string} First fix in flightPlan or exit fix of SID
+     */
+    getDepartureExitView() {
+        if (!this.hasSidLeg()) {
+            return this.getRouteString().split('..')[0];
+        }
+
+        return _last(this.getRouteString().split('.'));
```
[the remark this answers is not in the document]
erikquinn: Needs to use `.getFullRouteString()`. The value that goes into the scratchpad needs to be set only once, and should _not_ be changed as the aircraft passes fixes and legs. So examine the first leg (not the current one, the _first_ one), if it's a Sid leg, take the exit, and if it's anything else, take the first waypoint.

felixscheffer:
@n8rzz  Could you have a look at the `set scratchpad` command and ensure that it works properly. Right now, it response with `SET SCRATCHPAD`

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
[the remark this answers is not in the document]
n8rzz: @erikquinn almost made it!  That's why we do pull requests...

@felixscheffer thats a good callout, I can add a test for that

n8rzz:
@felixscheffer I'll have to verify the command tomorrow, I've run out of time for today.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +307,77 @@ export default class RadarTargetModel {
...
+     * Generate a string to be used for the second row of a datablock
+     * when the timeshare section is active
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowTwoSecondaryInfo
+     * @returns {string}
+     */
+    buildDataBlockRowTwoSecondaryInfo() {
+        const scratchPadText = _pad(this._scratchPadText, 4, ' ');
+        const aircraftModelIcao = _pad(this.aircraftModel.model.icao.toUpperCase(), 4, ' ');
```
erikquinn: Super sorry, I messed up-- the scratchpad should be _three_ characters max, not four. I saw 4 characters in the vSTARS documentation, and was like "that doesn't feel right, but it must be right"-- it must have been an old screenshot, because, in action, vSTARS itself also uses three characters, which is definitely correct.

These should both be changed to 3 characters. Perhaps this mistake of mine indicates that there's a point to be made to use an enum that's linked to theme changes instead of hardcoding this value. A todo to that effect would be great :smile: 

---

Also, the use of `_pad()` here is causing the value to be padded from the left, it should always be padded at the end only, in order for the first character to be aligned with the first character of the first line. `_padEnd()` should do what we need here.

image

src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -473,6 +484,32 @@ export default class RouteModel extends BaseModel {
...
+     */
+    getFlightPlanEntry() {
+        if (!this.hasSidLeg()) {
+            return this.getFullRouteString().split('..')[0];
+        }
+
+        const routeStringParts = this._divideRouteStringIntoSegments(this.getFullRouteString());
+        const temporaryLegModel = new LegModel(routeStringParts[0]);
+
+        return temporaryLegModel.getExitFixName();
```
erikquinn: As mentioned by @felixscheffer , suggest:
```suggestion
        const sidLegIndex = this._findSidLegIndex();
        const sidLegModel = this._legCollection[sidLegIndex];
        
        return sidLegModel.getExitFixName();
```

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -239,8 +243,8 @@ export default class RadarTargetModel {
      * @chainable
      */
     _initializeScratchPad() {
-        if (!this.aircraftModel.destination) {
-            this._scratchPadText = 'XXX';
+        if (this.aircraftModel.isDeparture()) {
+            this._scratchPadText = this.aircraftModel.fms.getFlightPlanEntry();
```
erikquinn: This will need to be limited to 3 characters and upper-cased, probably best to achieve by adding a setter which does a `text.substr(0, 3).toUpperCase()`.

Also, two lines down will need `.substr(1, 3).toUpperCase()` in order to show `SEA` in the data block instead of `sea`.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -303,6 +307,77 @@ export default class RadarTargetModel {
...
+     * Generate a string to be used for the first row of a datablock
+     *
+     * @for RadarTargetModel
+     * @method buildDataBlockRowOne
+     * @returns {string}
+     */
+    buildDataBlockRowOne() {
+        let dataBlockRowOne = this.aircraftModel.callsign;
+
+        if (this.aircraftModel.model.weightclass === 'H') {
```
erikquinn: This `H` should also be applied to super aircraft, which have a `weightclass` of `U`.

An `||` here is fine, or perhaps adding a method `AircraftTypeDefinitionModel.isHeavyOrSuper()`.

src/assets/scripts/client/scope/RadarTargetModel.js
n8rzz: screen shot 2018-11-25 at 1 41 26 pm
screen shot 2018-11-25 at 1 41 30 pm

n8rzz: @erikquinn just noticed these screenshots are still wrong.  `AAA-BBBB`, I'm missing a character in there.

PR #1188, opened 2018-11-12
refactor/727 - Update the tutorial

### LINKED ISSUE #727: Update the tutorial
Browser: All

Airport: All

#### Issue Description
The tutorial could use an update. It was written a while ago and while it is still correct, we could add some mechanics and remove some of the less important ones.

AbstractBeliefs:
Resolves openscope/openscope#727

The purpose of this pull request is to update the tutorial to better match the current simulator features, and to cover topics in more depth than is currently.

The PR comes in a couple of parts - it might be easier to look at the commits alone rather than the accumulated diff.

- [x] Formatting consistency, message construction
- [x] Fix message placeholders
  - [x] `taxi`/`wait` needs a runway
  - [x] `ils` needs a runway
  - [x] Remove surplus find/replaces
- [x] Fact checks
  - [x] Strips 2 - Strip doesn't show assigned speed
  - [x] Strips 2 - Strip doesn't show current activity
  - [x] Approach Clearances 1 - Do a/c actually have to be intercepting? They appear to try and do so themselves.
  - [x] Departure destination - what's the blue dashed line?
- [ ] Improvements
  - [x] Strips 1 - you may need to pull the strips out
  - [x] Windsock - you may need to collapse the strips
  - [ ] SID/STAR routing
  - [ ] General routing/handling demo
  - [x] Radar vectors - mention turn left/right by x degrees, rather than to a heading
  - [x] `<a>` tag colour for Good Job! linking to github

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Enhancements & Refactors
 - [#1264](https://github.com/openscope/openscope/issues/1264) &[#1265](https://github.com/openscope/openscope/issues/1265) - Change the "super" weight class identifier from `U` to `J`
 - [#1269](https://github.com/openscope/openscope/issues/1269) - Updates [Google Analytics](https://developers.google.com/analytics/devguides/collection/gtagjs/migration) tracking function from `ga` to `gtag`
+- [#727](https://github.com/openscope/openscope/issues/727) &[#1265](https://github.com/openscope/openscope/issues/1265) - Updates to the tutorial
 
 
 # 6.8.0 (December 1, 2018)
--- a/src/assets/scripts/client/ui/TutorialStep.js
+++ b/src/assets/scripts/client/ui/TutorialStep.js
@@ -25,6 +25,9 @@ export default class TutorialStep {
     }
 
     /**
+     * Replace tokens with values from current sim state
+     * or return simple text.
+     *
      * @for TutorialStep
      * @method getText
      * @return {string}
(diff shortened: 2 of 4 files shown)


felixscheffer:
What's the state of this PR? Can we merge it as it is?

AbstractBeliefs:
Can be merged, all the needed corrections are in, but I was hoping to get a
little bit on routing and add that in. Might be worth another pr.

src/assets/scripts/client/tutorial/TutorialView.js
```diff
@@ -283,9 +282,9 @@ export default class TutorialView {
         this.tutorial_step({
             title: 'Takeoff, part 1',
             text: ['When it appears at the start of runway ({RUNWAY}) (which may take a couple of seconds), click it (or press the up arrow once)',
-                   'and type in &lsquo;caf&rsquo; (for &lsquo;cleared as filed&rsquo;). This tells the aircraft it is cleared to follow its flightplan.',
+                   'and type in &lsquo;caf&rsquo; (for "cleared as filed"). This tells the aircraft it is cleared to follow its flightplan.',
```
n8rzz: if we're escaping quotations and not encoding them, we should be consistent in how we do that.  these should be escaped double quotes.

n8rzz:
```suggestion
                   'and type in "caf" (for "cleared as filed"). This tells the aircraft it is cleared to follow its flightplan.',
```

felixscheffer:
@AbstractBeliefs can you create a new issue for the routing part?

PR #1196, opened 2018-11-19
bugfix/1193

n8rzz:
adds SettingsController instantiation to UiController.init()

### DIFF
--- a/src/assets/scripts/client/ui/UiController.js
+++ b/src/assets/scripts/client/ui/UiController.js
@@ -146,10 +146,6 @@ class UiController {
         this._eventBus = null;
         this.tutorialView = null;
         this.settingsController = null;
-        // TODO: possibly deprecated
-        this.ui = {};
-        // TODO: possibly deprecated
-        this.ui.scale = INVALID_NUMBER;
 
         this.$element = null;
         this.$airportDialog = null;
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/ui/UiController.js
```diff
@@ -142,7 +145,10 @@ class UiController {
     destroy() {
         this._eventBus = null;
         this.tutorialView = null;
+        this.settingsController = null;
+        // TODO: possibly deprecated
         this.ui = {};
+        // TODO: possibly deprecated
         this.ui.scale = INVALID_NUMBER;
```
felixscheffer: I think these two lines can be removed. @erikquinn

erikquinn: I think so? Can we try removing it real quick? If it still zooms properly, then we should be good, and can answer the question rather than asking it

felixscheffer:
@erikquinn Can you merge this PR. The settings dialog in the develop branch isn't working and that's somewhat hindering.

src/assets/scripts/client/ui/UiController.js
erikquinn: Unable to find this being used anywhere. Removed and verified it doesn't break the zooming, so we should be good. Removed and pushed.

erikquinn:
@felixscheffer Yes, I was just waiting for word on whether or not removing the probably-dead code would break anything or not so we could do so. I verified that it could be removed, and it is now gone, so we're good to go here. :shipit:

Issue #1191, opened 2018-11-17
Consolidate command bar buttons


erikquinn:
We are accumulating a lot of buttons in the command bar, and they should be consolidated into groups so they take up less room, are more organized, and more intuitive to figure out what the icons are supposed to represent because you have additional context.

```
Scope --> [SIDs, STARs, terrain, video map, navaids, restricted, range rings]
Settings
Help --> [Tutorial, GitHub link, YouTube link, Slack link, Command Reference link, Airport Guide link]
```

felixscheffer:
Can someone create/find/suggest an icon for "scope"?

@Marcel510

erikquinn:
I can vectorize this one real quick...

http://chittagongit.com/icon/radar-icon-vector-22.html
![image](https://user-images.githubusercontent.com/5103735/48664589-0a4eb580-ea6f-11e8-9e96-82d18d987eaa.png)

felixscheffer:
Looks awesome! 👍

erikquinn:
All done!

![image](https://user-images.githubusercontent.com/5103735/48664748-5995e580-ea71-11e8-9efa-76c4d99547ef.png)


[scope.zip](https://github.com/openscope/openscope/files/2592150/scope.zip)

Fechulo:
It would be good to eventually have a link to the youtube channel and maybe even twitter or reddit

JPmAn24:
Possibly an icon for the help section?

PR #1198, opened 2018-11-20
feature/1105 - Show airport guide in game

### LINKED ISSUE #1105: Show airport guide in-game
An option to show the airport guide in game would be helpful. We would want it to be toggled on and off, and have the markdown file rendered into html. I know nothing about doing this, but came across "showdown" in a few minutes of half-baked research:

https://github.com/showdownjs/showdown

We have the guides, we just need to make them more accessible to the users!

Related: openscope/openscope#804 (same thing for the changelog)

JPmAn24:
Resolves openscope/openscope#1105 

The purpose of this pull request is to let the airport guide be shown in-game


sam-irl:
@JPmAn24 it's still yours, haha. Just thought I'd save you some of the work. If you don't want to continue it you can assign me

sam-irl:
todo:
- [ ] render whitespace so the ascii tables look less odd
- [ ] refactors: i'm pretty sure a good chunk of this is not compliant with style
- [ ] potentially format markdown headers to <h[x]> tags

src/assets/scripts/client/airportGuide/AirportGuideViewModel.js
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Toggles the view when the selector is clicked.
+     *
+     * @for AirportGuideViewModel
+     * @method toggleView
+     */
+    toggleView() {
+        this.$element.toggleClass(SELECTORS.CLASSNAMES.AIRPORT_GUIDE_IS_OPEN); /*AIRPORT_GUIDE_IS_*/
+    }
+}
```
felixscheffer: please add an empty line at the end of the file

src/assets/scripts/client/airportGuide/AirportGuideViewController.js
```diff
@@ -0,0 +1,171 @@
...
+     *
+     * @for AirportGuideViewController
+     * @method hasAirportGuide
+     * @param {String} icao
+     * @returns {Boolean} whether a guide was found
+     */
+    hasAirportGuide(icao) {
+        return _has(this._guideData, icao);
+    }
+}
```
felixscheffer: please add an empty line at the end of the file

src/assets/scripts/client/airportGuide/guides/cyhz.js
```diff
@@ -0,0 +1,138 @@
+export const CYHZ = `
```
felixscheffer: is this a copy of cyhz.md? can't we just use the original cyhz.md?

sam-irl:
Leaving this one for now:
- We need to develop a strategy for reading the markdown files, as `fs` won't work in the browser
- It might be an idea to wait for #1191 before continuing

erikquinn:
No need to wait for #1191 :smiley: 

Instead of manually copying the files like this, there is surely some way we can parse *.md files directly to strings. Perhaps the below link? If it doesn't work, then at the very worst, we could create all these files automatedly somehow-- we just want to avoid having two sets of documentation, where we're supposed to update both of them by hand.
https://github.com/parcel-bundler/parcel/issues/970#issuecomment-381403710

n8rzz:
I’m going to try and look into this stuff this week. 

It seems we have two options:
- complie md files into a readable format at build time
- deploy our express server and serve things from there. This requires a multi Buildpack, some configuration changes and actually deploying our node server.

erikquinn:
Stumbled across this in `tools/tasks/jsonAssembler.js`

image

n8rzz:
@erikquinn correct, we are using `fs` here.  I'm hoping to use a similar setup to convert the md to something usable and then either serve it or inject it for use in-browser.

n8rzz:
I don't know guys, looking into this a bit more, I think this is actually pretty close to what we will need.  I feel like this is maybe 75% of the way there already.

I'm going to take this over for a little bit to get more of the parts connected, but this is in better shape that I first thought.

nice work so far @JPmAn24 @sam-irl !

JPmAn24:
So far so good, y'all!

src/assets/scripts/client/App.js
```diff
@@ -127,18 +127,20 @@ export default class App {
     loadAirlinesAndAircraft(airportLoadList, initialAirportIcao, initialAirportResponse) {
         const airlineListPromise = $.getJSON('assets/airlines/airlines.json');
         const aircraftListPromise = $.getJSON('assets/aircraft/aircraft.json');
+        const airportGuideListPromise = $.getJSON('assets/guides/guides.json');
```
sam-irl: To avoid screwing around with `App` and the constructors, we could pass `this.contentQueue` in AppController to the Guide Controller's constructor, and then load it then, instead of at initial load.

What I'm thinking:

```javascript
// <AppController>
this.airportGuideInfoController = new AirportGuideInfoController(this.contentQueue);
// <AirportGuideInfoController>
loadGuideData(): void
const guidePromise = this.contentQueue.add({ url: `assets/guides/${nextIcao}`, immediate: true});
guidePromise.then((data, textStatus, jqXHR) => { // handle + fire a callback });
```

src/assets/scripts/client/AppController.js
```diff
@@ -139,8 +142,16 @@ export default class AppController {
...
+     * @param airportGuideData {object}
      */
-    setupChildren(airportLoadList, initialAirportIcao, initialAirportData, airlineList, aircraftTypeDefinitionList) {
+    setupChildren(
+        airportLoadList,
+        initialAirportIcao,
+        initialAirportData,
+        airlineList,
+        aircraftTypeDefinitionList,
+        airportGuideData
```
sam-irl: see above

src/assets/scripts/client/airportGuide/AirportGuideViewController.js
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Event handler for when a `airportGuideView` instance is clicked
+     *
+     * @for airportGuideViewController
+     * @method _onToggleAirportGuide
+     * @param event {JQueryEventObject}
+     * @private
+     */
+    _onToggleAirportGuide() {
+        this.airportGuideViewModel.toggleView();
```
sam-irl: We should toggle the `active` class on the button in the footer

n8rzz:
interestingly, the tutorial seems to be broken in this pr.  I just updated with develop, where things are good.  I'll see if I can find where and why things broke with that.

sam-irl:
Oddly enough, I get nicely formatted guides when I build locally, but not when I look at the deployment. 😕 

This isn't the first time that's happened - I think something's wrong with the Heroku stack

src/assets/scripts/client/App.js
n8rzz: something like that is a good idea here because the data we need to load is not needed before the app starts (like airports, airlines and aircraft).  Could you add an issue for that?  I'd like to be able to take our time on it and get this current thing completely working first.

src/assets/scripts/client/airportGuide/AirportGuideViewController.js
n8rzz: correct.  the opening and closing of the modal should be handled by the `UiController` (next thing I'm tackling), and within that, we toggle the icon's active class

src/index.hbs
```diff
@@ -13,6 +13,8 @@
         --}}
         {{> _airportSwitch}}
 
+        {{> _airportGuide}}
```
n8rzz: @sam-irl this is what I was talking about on your changelog PR

src/templates/_airportGuide.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="airportGuide-container nice-scrollbar">
```
n8rzz: @sam-irl (part 2).  If we can have the markup in the DOM, we should put it in the DOM.  In other words, if we don't need to generate markup via js, then we probably shouldn't.

n8rzz:
Holy crap guys, I think this ready to go!

src/assets/scripts/client/App.js
n8rzz: #1321

sam-irl:
Just needs to be added to `UiController.closeAllDialogs()`

n8rzz:
`esc` added

src/assets/scripts/client/airportGuide/AirportGuideView.js
```diff
@@ -0,0 +1,135 @@
...
+     * Lifecycle method
+     *
+     * Should be called once on instantiation
+     *
+     * @for AirportGuideView
+     * @method _init
+     * @private
+     * @chainable
+     */
+    _init() {
```
felixscheffer: @n8rzz  shoudn't this be public?

n8rzz: ahh, damn, I'm breaking one of my conventions.  fixed

sam-irl:
I hate to hold this up more because I'm excited to see this merged.

That said, would it be beneficial to have a search and replace for
`find: <a`
`replace: <a target="_blank"`

That way, the hyperlinks will open in a new tab

n8rzz:
@sam-irl that's a nice idea, however, we should treat the generated `html` as black-boxed.  we cannot modify it, nor should we modify it.  we load it and display it.  

we could create a new issue to find each `a` element and add a handler to open a new window on click?  then we can also track those outbound clicks.

tools/tasks/markdownAssembler.js
```diff
@@ -0,0 +1,124 @@
...
+ * is itself a string of HTML.
+ *
+ * @returns {Promise} promise
+ */
+function markdownAssembler() {
+    fancyLog(colors.cyan('--- preparing guides.json'));
+
+    const airportGuideDict = _generateAirportGuideDict();
+    const markdown = parseFiles(airportGuideDict);
+    _writeFileOutput(markdown);
```
felixscheffer: add a blank line before this line

tools/tasks/markdownAssembler.js
```diff
@@ -0,0 +1,124 @@
+const fs = require('fs');
+const fancyLog = require('fancy-log');
+const colors = require('ansi-colors');
+const path = require('path');
+const mkdirp = require('mkdirp');
+const showdown = require('showdown');
+
```
felixscheffer: remove empty line

src/assets/scripts/client/ui/UiController.js
```diff
@@ -417,6 +629,18 @@ class UiController {
         this._onClickOpenAirportDialog();
     }
 
+    /**
+     * @for UiController
+     * @method onToggleAirportGuide
+     */
+    onToggleAirportGuide() {
```
felixscheffer: IMHO all the show/hide dialog code should live in the respective controller. We can refactor that later though. Any thoughts?

@n8rzz @erikquinn

src/assets/scripts/client/airportGuide/AirportGuideViewController.js
```diff
@@ -0,0 +1,234 @@
...
+     * @for airportGuideViewController
+     * @method destroy
+     * @chainable
+     */
+    destroy() {
+        this._eventBus = null;
+        this._$element = null;
+        this._initialIcao = null;
+        this._airportGuideView = null;
+        this.$airportGuideTrigger = null;
```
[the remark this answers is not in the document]
n8rzz: good catch!

src/assets/scripts/client/airportGuide/AirportGuideView.js
```diff
@@ -0,0 +1,135 @@
...
+     * Should be run only once on instantiation
+     *
+     * @for AirportGuideView
+     * @method _createChildren
+     * @param {JQuery|HTMLElement} $element
+     * @private
+     * @chainable
+     */
+    _createChildren($element) {
+        this.$element = $element.find(SELECTORS.DOM_SELECTORS.AIRPORT_GUIDE_CONTAINER);
```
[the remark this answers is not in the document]
n8rzz: with these view classes, my _intent_ is that the `$element` in question is the root element the view is concerned with, and has a similar name.

But, you know me, I'm all for verbose naming.  Doesn't really matter to me

src/assets/scripts/client/ui/UiController.js
```diff
@@ -214,6 +420,11 @@ class UiController {
         return this.$airportDialog.hasClass(SELECTORS.CLASSNAMES.OPEN);
     }
 
+
+    isAirportGuideDialogOpen() {
+        return this.$airportGuideDialog.hasClass(SELECTORS.CLASSNAMES.AIRPORT_GUIDE_IS_OPEN);
```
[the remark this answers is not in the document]
n8rzz: only problem with that is that the UiController would then need to have a reference to an instance of the `AirportGuideController`.  which it currently does not.  I see where you're going with this and I tend to agree, however, we're just not set up to do it this way right now.

and maybe that's a different problem, the AirportGuide *should* live under the `UiController`?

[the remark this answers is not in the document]
felixscheffer: My main issue is that we keep adding more and more dialogs and the UiController is getting more and more messy. So my idea was that everything that is related to a dialog should live in the respective controller but if someone has a better idea, let me know.

Yes IMHO the AirportGuideController should live under the UiController as airport guides are "ui feature" much like the tutorial.

src/assets/scripts/client/ui/UiController.js
n8rzz: I agree with you there.  The `UiController` should listen and delegate events that occur in the UI.

Looking at this a little more, that's more or less what it's doing.  with most of these options, the thing we are toggling is the active state of the footer button.  after that, we trigger a `TrackableEvent` and then trigger an `EventBus` event which then causes whatever controller is responsible for the view to respond and do it's thing from there.

This could be made more clear, for sure.  This implementation is inline with the others, for now anyway.

guyndumais:
Sorry pressed "close PR" instead of "goto PR"...

erikquinn:
I think this PR got lost somehow. @n8rzz this is done right?

n8rzz:
review app looks good, I'd say yes

PR #1200, opened 2018-11-20
Add 'hold short' portion to taxi command

### LINKED ISSUE #1199: Taxi command should say "Taxi to and hold short runway..."
#### Issue Description

When controller types in command "taxi 22", currently the log says "Taxi to runway 22". Since it is really holding short, the taxi command should say "Taxi to and hold short runway 22".

#### Steps to reproduce
- type in taxi command

artur-gajewski:
Resolves #1199 

The purpose of this pull request is to correctly state to hod short of specified runway when entering `taxi` command.


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1494,6 +1494,12 @@ export default class AircraftModel {
                 if (elapsed > this.taxi_time) {
                     this.setFlightPhase(FLIGHT_PHASE.WAITING);
                     this.moveToRunway(runwayModel);
+                    speech_say([
+                        { type: 'text', content: AirportController.airport_get().radio.twr },
+                        { type: 'callsign', content: this },
+                        { type: 'text', content: `, holding short runway ${runwayModel.getRadioName()}` }
+                    ]);
+                    UiController.ui_log(`${AirportController.airport_get().radio.twr}, ${this.callsign}, holding short runway ${runwayModel.name}`);
```
felixscheffer: Not really pretty.

@n8rzz Any thoughts?

n8rzz: agreed, not pretty, but not bad either.  two things here:
- I just had to look up `.speech_say()` and, holy crap, we need to re-work that one.  yikes!!
- I'd set the string actually being logged (LN#1502) to a constant first and then simply pass it on to `.ui_log()`

`.speech_say()` is super gross, but that's not this method's fault.  added #1204 to account for that future work

indianbhaji:
I guess there’s another problem in that phraseology is different across the world? In Germany it would be ‘taxi to holding point runway 26R via S and A13’, in the UK ‘taxi to holding point R1 (runway 22) via A, H and R’

PR #1201, opened 2018-11-20
devops/1197 - Expand analytics and event tracking

### LINKED ISSUE #1197: improve analytics event tracking
- [ ] clicks on command bar items
- [ ] clicks (and values) of game options
- [ ] invalid commands
- [ ] tutorial progress

n8rzz:
Resolves #1197 

adds EventTracker class and begins adding trackable events

### DIFF
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -17,9 +17,10 @@ import ScopeModel from './scope/ScopeModel';
 import SpawnPatternCollection from './trafficGenerator/SpawnPatternCollection';
 import SpawnScheduler from './trafficGenerator/SpawnScheduler';
 import UiController from './ui/UiController';
+import { speech_init } from './speech';
 import { EVENT } from './constants/eventNames';
 import { SELECTORS } from './constants/selectors';
-import { speech_init } from './speech';
+import { TRACKABLE_EVENT } from './constants/trackableEvents';
 
 /**
  * Root controller class
@@ -139,7 +140,7 @@ export default class AppController {
      * @param aircraftTypeDefinitionList {array<object>}
      */
     setupChildren(airportLoadList, initialAirportIcao, initialAirportData, airlineList, aircraftTypeDefinitionList) {
-        EventTracker.sendEvent('airports', 'initial-load', initialAirportIcao);
+        EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'initial-load', initialAirportIcao);
 
         this.$canvasesElement = this.$element.find(SELECTORS.DOM_SELECTORS.CANVASES);
 
@@ -268,7 +269,7 @@ export default class AppController {
             return;
         }
 
-        EventTracker.sendEvent('airports', 'airport-switcher', nextAirportJson.icao);
+        EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-switcher', nextAirportJson.icao);
 
         NavigationLibrary.reset();
         this.airlineController.reset();
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -270,7 +270,6 @@ export default class AppController {
         }
 
         EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-switcher', nextAirportJson.icao);
-
         NavigationLibrary.reset();
         this.airlineController.reset();
         this.aircraftController.aircraft_remove_all();
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -140,7 +140,7 @@ export default class AppController {
      * @param aircraftTypeDefinitionList {array<object>}
      */
     setupChildren(airportLoadList, initialAirportIcao, initialAirportData, airlineList, aircraftTypeDefinitionList) {
-        EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'initial-load', initialAirportIcao);
+        EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'initial-load', initialAirportIcao);
 
         this.$canvasesElement = this.$element.find(SELECTORS.DOM_SELECTORS.CANVASES);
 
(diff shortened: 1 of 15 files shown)


felixscheffer:
Needs changelog entry

n8rzz:
@felixscheffer yep, that's coming in the next commit along with the missing documentation parts

src/assets/scripts/client/ui/TutorialView.js
```diff
@@ -702,10 +705,11 @@ export default class TutorialView {
...
         prop.tutorial.step = clamp(0, prop.tutorial.step - 1, prop.tutorial.steps.length - 1);
 
+        EventTracker.trackEvent(TRACKABLE_EVENT.TUTORIAL, 'prev', `${prop.tutorial.step}`);
         this.tutorial_update_content();
     }
 
-    // TODO: this function never gets called in this file
+    // TODO: this method never gets called ever, remove
     /**
      * @method tutorial_resize
```
erikquinn: I concur-- can you just remove it then?

src/assets/scripts/client/ui/UiController.js
```diff
@@ -67,6 +69,7 @@ class UiController {
         this.$toggleTutorial = this.$element.find(SELECTORS.DOM_SELECTORS.TOGGLE_TUTORIAL);
         this.$toggleOptions = this.$element.find(SELECTORS.DOM_SELECTORS.TOGGLE_OPTIONS);
         this.$toggleVideoMap = this.$element.find(SELECTORS.DOM_SELECTORS.TOGGLE_VIDEO_MAP);
+        this.$githubLinkElement = this.$element.find(SELECTORS.DOM_SELECTORS.GITHUB_EXTERNAL_LINK)
```
erikquinn: Also add to `.destroy()`, and can you place it alphabetically in both?

src/assets/scripts/client/ui/UiController.js
```diff
@@ -104,6 +107,7 @@ class UiController {
         this.$toggleTutorial.on('click', (event) => this.onToggleTutorial(event));
         this.$toggleOptions.on('click', (event) => this.onToggleOptions(event));
         this.$toggleVideoMap.on('click', (event) => this.onToggleVideoMap(event));
+        this.$githubLinkElement.on('click', (event) => this.onClickGithubLink(event));
```
erikquinn: Can you move this up to be in alphabetical order as the others are?

src/assets/scripts/client/ui/UiController.js
```diff
@@ -129,6 +132,7 @@ class UiController {
         this.$toggleTutorial.off('click', (event) => this.onToggleTutorial(event));
         this.$toggleOptions.off('click', (event) => this.onToggleOptions(event));
         this.$toggleVideoMap.off('click', (event) => this.onToggleVideoMap(event));
+        this.$githubLinkElement.off('click', (event) => this.onClickGithubLink(event));
```
erikquinn: Can you move this up to be in alphabetical order as the others are?

src/assets/scripts/client/ui/UiController.js
```diff
@@ -387,6 +393,8 @@ class UiController {
      * @private
      */
     _onClickCloseAirportDialog() {
+        EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-switcher', 'close');
+
         this.$airportDialog.removeClass(SELECTORS.CLASSNAMES.OPEN);
```
erikquinn: The empty line is probably not needed, since both of these are method calls. If you think it should be this way, then there are others to change to be like this.

n8rzz:
Our current settings in GA only listen for requests on specific urls.  Remember we already have GA setup to listen for page views, sessions, etc.  We are ignoring our dev envs and local envs so those settings shouldn't need to change.  If at any point in the future we decide that we should ignore specific urls, we cant do that within the settings controls of GA.

src/assets/scripts/client/AppController.js
```diff
@@ -265,6 +269,8 @@ export default class AppController {
             return;
         }
 
+        EventTracker.trackEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-switcher', nextAirportJson.icao);
+
```
felixscheffer: Correct me if I am wrong but I think there should be no blank line here.

src/assets/scripts/client/game/GameController.js
```diff
@@ -421,6 +428,7 @@ class GameController {
         if (this.game.score === this.game.last_score) {
             return;
         }
+
         const $scoreElement = $(SELECTORS.DOM_SELECTORS.SCORE);
         $scoreElement.text(round(this.game.score));
```
felixscheffer: There should be a blank line...I think 😉

src/assets/scripts/client/AppController.js
n8rzz: you are correct, good catch!

src/assets/scripts/client/game/GameController.js
n8rzz: correct again!

src/templates/_footer.hbs
```diff
@@ -45,7 +45,7 @@
             <img src="assets/images/icons/cog.png" />
         </li>
         <li class="control" title="View on GitHub">
-            <a href="https://github.com/openscope/openscope" target="_blank">
+            <a href="https://github.com/openscope/openscope" id="github-external-link" target="_blank">
```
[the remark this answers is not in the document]
n8rzz: Good question!  Correct, I do advocate against them in just about every case.  This case is the exception for one reason:
1) there should only ever be one instance of this in the app, using the id here helps reinforce that assumption

One oversight on my side was the missing `js-` in the id name.  so I've updated that.

src/assets/scripts/client/EventTracker.js
```diff
@@ -0,0 +1,88 @@
...
+     * Send a custom event to google analytics
+     *
+     * @for EventTracker
+     * @method trackEvent
+     * @param category {TRACKABLE_EVENT}
+     * @param action {string}
+     * @param label {string}
+     * @param value {string|null} [optional]
+     */
+    trackEvent(category, action, label, value = null) {
```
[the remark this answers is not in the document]
erikquinn: I agree that "track" has more meaning, and that's my concern: as I interpret the word, its meaning is contradictory to what it's actually doing. When I see `trackEvent`, it sounds like `begin tracking this event`, when in reality, it actually means `advise GA that the following tracked event has occurred`.

Do any of these work for you? `recordEvent()` / `reportEvent()` / `catalogEvent()` / maybe even `documentEvent()`?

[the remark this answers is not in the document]
n8rzz: I see what you mean, lets go with `recordEvent`

PR #1203, opened 2018-11-20
bugfix/1202

### LINKED ISSUE #1202: Wrong command makes pilot respond "say again" twice
Browser: Chrome Version 70.0.3538.77 (Official Build) (64-bit) 

Airport: Any

Approximate play time before issue: couple hours

#### Issue Description

When you send a command to an airplane that does not exist, the pilot responds "say again , say again".

#### Steps to reproduce
- Select airplane
- Type in "where ya at, fool?"
- Pilot responds: "say again , say again"

screen shot 2018-11-20 at 20 04 09

artur-gajewski:
Resolves #1202 

The purpose of this pull request is to fix little error of pilot requesting to say again twice in a row when invalid command is given.

### DIFF
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -648,7 +648,7 @@ export default class Pilot {
 
         const readback = {};
         readback.log = 'intercept glidepath';
-        readback.log = 'intercept glidepath';
+        readback.say = 'intercept glidepath';
 
         return [true, readback];
     }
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -648,7 +648,7 @@ export default class Pilot {
 
         const readback = {};
         readback.log = 'intercept glidepath';
-        readback.log = 'intercept glidepath';
+        readback.say = 'intercept glidepath';
 
         return [true, readback];
```
felixscheffer:
```suggestion
        return [true, 'intercept glidepath'];
```

artur-gajewski: Also removed the portion of code that creates an object.

artur-gajewski:
@n8rzz as far as I know and have checked the code, the tests do test for read back but based on the controller functions called and not free text (argument parser).

n8rzz:
That sounds familiar, I think it's just going to return `undefined` or something and then the `Pilot` handles it from there.

Either way, nice work! 👍

PR #1206, opened 2018-11-21
Readback in correct order

### LINKED ISSUE #1205: Pilot's readback is in incorrect order
Browser: All

Airport: All

Approximate play time before issue: Hours

#### Issue Description

When pilot is saying the readback, the order is <callsign> <readback> when it should be the other way.

#### Steps to reproduce
- Just give any instructions to the pilot

artur-gajewski:
Resolves #1205

The purpose of this pull request is to educate pilots in which order to read the instructions. First the instruction then the callsign.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -127,10 +127,10 @@ export default class AircraftCommander {
             const r_log = _map(response, (r) => r.log).join(', ');
             const r_say = _map(response, (r) => r.say).join(', ');
 
-            UiController.ui_log(`${aircraft.callsign}, ${r_log} ${response_end}`, redResponse);
+            UiController.ui_log(`${r_log} ${response_end}, ${aircraft.callsign}`, redResponse);
             speech_say([
-                { type: 'callsign', content: aircraft },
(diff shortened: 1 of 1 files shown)


artur-gajewski:
How about if text would be as is but the voice would have the swapped version?

In what instances the text/voice is the ATC instruction instead of pilot readback?

erikquinn:
That could be kind of tough to have them different. For example, it's easy to reverse this:
```
AAL123, descend and maintain 3000
    v    v    v    v    v
Descend and maintain 3000, AAL123
```
but less easy when you consider that a variety of things are being said, for example this:
```
Miami Approach, AAL123 with you at 16000
    v    v    v    v    v
?????????????????????
```

felixscheffer:
@erikquinn `with you` has its own speech_say instruction and does not use the normal readback code. I am not sure if there are other cases that may cause problems but this one should be easy to deal with.

felixscheffer:
I like the idea to keep the text readback as it is and reverse the speech readback.

CHANGELOG.md
```diff
@@ -7,12 +7,12 @@
 
 
 
-
```
felixscheffer: ?

artur-gajewski: What was this line with no content other than - ?

Issue #1137, opened 2018-10-20
Update project dependencies


n8rzz:
Its been some time since we've updated our libraries, which is like a lifetime in JS.

Things we should update/add:
- [x] node reference in `package.json`
- [x] npm reference in `package.json`
- [x] babel and supporting babel plugins
- [x] eslint and supporting packages
- [x] create `.nvmrc` file for current node version
- [x] update to `node@11.3.0` and current npm
- [x] upgrade to babel 7
- [x] update babelrc
- [x] upgrade to `gulp@4.0.0`
- [x] refactor gulp tasks to use v4 api
- [x] remove/replace `gulp.util` things

Work here will very likely expose new issues with upgraded packages.  Those should all be addressed as part of this issue.  it could be a big chunk of work.

felixscheffer:
gulp-util is deprecated and needs to be replaced.

https://www.npmjs.com/package/gulp-util

felixscheffer:
gulp-minify-css is deprecated. Please use gulp-clean-css.

https://www.npmjs.com/package/gulp-minify-css

n8rzz:
@erikquinn you want me to merge this one or wait till next month?

n8rzz:
hmm, never mind, I answered my own question.  going to swing for the fences on this one and upgrade `gulp` and move our `node` dependency all the way to modern with `11.3.0`.  

Moving back to WIP

PR #1207, opened 2018-11-21
devops/526 - Add typescript support

### LINKED ISSUE #526: Add Typescript support
Use typescript compiler and `tsify` to produce similar `bundle.js` and `bundle.js.map` files during build process

This issue does not cover a complete conversion.  This is just the first step to an eventual complete conversion to typescript

n8rzz:
Resolves openscope/openscope#526 
depends on openscope/openscope#1137 

Switches over to Typescript and updates build pipeline to use `tsify`

- [ ] Enable typescript for tests
- [ ] Testing


erikquinn:
@n8rzz @felixscheffer What's the status of this? Is it planned to continue, or is it a no-go?

felixscheffer:
I dont know what's the status of this. It's @n8rzz's baby but it would be awesome if we could switch to typescript.

n8rzz:
Very stalled.

This is the closest I've been able to get with it, but it's still miles away from actually working correctly.  I would really love to make this a thing someday, but I fear it will take a very large amount of work to accomplish (a la 2016 refactor).

n8rzz:
My plan for this was to get just enough TS working so that we could use the compiler to do our transpiling while still allowing us to use existing es6 while we slowly convert everything to TS.  I'm not sure that is possible in our case.  Or, if it is, I haven't had any luck making that happen.  

I think the path forward here may be to convert everything all at once with loose compiler settings.  Problem with doing it all at once is where do you start?  If you start at `index.js` and work your way down, you very quickly end up making a ton of changes just to get the compiler building things.

felixscheffer:
I fixed the imports and the sim is loading again but now the tests are broken.

n8rzz:
Review app is building now

felixscheffer:
@n8rzz Looks like we have to fix the tests first...could you have a look?

n8rzz:
@felixscheffer Looks like the tests need the same path updates for lodash imports.  I've been swamped at work so I barely had a chance to look at this, but that is my first impression.  Ava should be able to work with TS without too much extra configuration, be we'll see what happens once those imports are updated.

erikquinn:
Closing I guess. Can always reopen!

Issue #186, opened 2017-01-28
Speech Readback


indianbhaji:
Just a few issues that may need addressing:

1. Wrong wind! The guy says 'wind' as in to wind someone up. Not the wind that blows things away.
2. It would be nice to say '100 degrees at 0 knots' instead of '100 at 0'
3. The guy speaks quite slowly?

erikquinn:
1. Haha! Maybe we can tell her to say "wynd" instead and hope that gives the correct pronunciation.
 2. `wind one zero zero at five` is correct in FAA phraseology, which (when applicable) is the phraseology rule set we will be sticking to until we can have multiple sets of phraseology (ICAO phraseology mode vs FAA, etc). However, you definitely bring up a good point that if it's _literally_ `100 @ 0`, we should instead be saying `wind calm`.
 3. We can speed it up a bit! :+1:

felixscheffer:
@indianbhaji @erikquinn  can you confirm that these issues still exist (the ticket is more than a year old)?

1. I am not a native speaker but "wind" sounds good me.
2.  ...
3. IMHO the speed is fine. Maybe someone increased it (currently 1.125).

indianbhaji:
@felixscheffer I’ll check later but the lady was saying wind as in to twist something like a rope around an object

indianbhaji:
@felixscheffer My apologies for not getting back to you on this. I was on my phone and completely forgot... anyways, the issue still remains. The lady is saying 'wind' as in the childhood riddle 'wind the bobbin up, wind the bobbin up, pull, pull, clap clap clap' instead of the movement of air

Fechulo:
I wonder if simply adding the word "surface" before it could fix it, just giving it some context.

erikquinn:
Resolved by #1213. Closing.

PR #1213, opened 2018-11-22
feature/1212 - Create aircraft specific pilot voice for read backs

### LINKED ISSUE #1212: Randomize female/male speech synth for each aircraft
Browser: n/a

Airport: n/a

#### Issue Description

Each aircraft should have their own voice, either male or female. Randomize which one when aircraft initialized.

artur-gajewski:
Resolves #1212 and #186

The purpose of this pull request is to assign aircraft specific pilot voice so that we as controllers get different voices over the radio instead of just one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 
 
+
 ### Bugfixes
 - [#1147](https://github.com/openscope/openscope/issues/1147) - Fix Callsigns in tutorial should update when switching airports
 - [#1045](https://github.com/openscope/openscope/issues/1045) - Fix descent planning logic so arrivals can meet their altitude restrictions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 - [#1003](https://github.com/openscope/openscope/issues/1003) - Add assigned runway to the aircrafts' strips
 - [#945](https://github.com/openscope/openscope/issues/945) - Add airport information panel to scope
 - [#159](https://github.com/openscope/openscope/issues/159) - Add in-game link to GitHub repo
-
+- [#1212](https://github.com/openscope/openscope/issues/1212) - Randomize pilot voice for each aircraft
 
 
 
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -652,7 +652,7 @@ export default class AircraftCommander {
         readback.log = `wind ${roundedWindAngleInDegrees} at ${roundedWindSpeed}, Runway ${runway.name}, ` +
             'cleared for takeoff';
 
-        // We have to make it say winned to make it sound like "Wind" and not "Wined"
+        // We have to make it say winned to make it sound like "Wind" and not "Whined"
         readback.say = `winned ${radio_spellOut(roundedWindAngleInDegrees)} at ` +
             `${radio_spellOut(roundedWindSpeed)}, Runway ${radio_runway(runway.name)}, cleared for takeoff`;
 
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -132,7 +132,7 @@ export default class AircraftCommander {
                     { type: 'callsign', content: aircraft },
                     { type: 'text', content: `${r_say} ${response_end}` }
                 ],
-                aircraft.voiceName
+                aircraft.pilotVoice
             );
         }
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 - [#1003](https://github.com/openscope/openscope/issues/1003) - Add assigned runway to the aircrafts' strips
 - [#945](https://github.com/openscope/openscope/issues/945) - Add airport information panel to scope
 - [#159](https://github.com/openscope/openscope/issues/159) - Add in-game link to GitHub repo
-
```
felixscheffer: Should be a new line.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -647,9 +650,11 @@ export default class AircraftCommander {
         aircraft.setFlightPhase(FLIGHT_PHASE.TAKEOFF);
         aircraft.scoreWind('taking off');
 
-        readback.log = `wind ${roundedWindAngleInDegrees} at ${roundedWindSpeed}, runway ${runway.name}, ` +
+        readback.log = `wind ${roundedWindAngleInDegrees} degrees at ${roundedWindSpeed}, runway ${runway.name}, ` +
```
erikquinn: no "degrees". In the absence of a feature to use multiple sets of phraseology, we use FAA phraseology as the standard, in which it is stated as `wind two one zero at one five gust two two`, so this should be reverted.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -647,9 +650,11 @@ export default class AircraftCommander {
...
             'cleared for takeoff';
-        readback.say = `wind ${radio_spellOut(roundedWindAngleInDegrees)} at ` +
+
+        // We have to make it say winned to make it sound like "Wind" and not "Wined"
+        readback.say = `winned ${radio_spellOut(roundedWindAngleInDegrees)} dgrees at ` +
```
erikquinn: no degrees, and probably want to change `wined` to `whined`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -168,6 +168,17 @@ export default class AircraftModel {
...
+        /**
+         * Pilot's voice
+         *
+         * Initially generated and assined on instantiation by the `AircraftController`
+         *
+         * @for AircraftModel
+         * @property pilotVoice
+         * @type {string}
+         */
+        this.pilotVoice = randomize_pilot_voice();
```
erikquinn: Just as a note, usually we would instantiate the property empty, then populate it in an `.init()` method, and reset it in a `.reset()` method. That way, it would be easily recycled and reused by the `ModelSourceFactory`. Sadly, we do not do this with the `AircraftModel` yet, which is a big missed opportunity that we can address sometime 😃 

Since we don't currently do so, this is perfectly good! 👍

src/assets/scripts/client/speech.js
```diff
@@ -21,12 +21,33 @@ export const speech_init = () => {
...
+        'Daniel',
+        'Fiona',
+        'Google US English',
+        'Google UK English Female',
+        'Google UK English Male'
+    ];
+
+    const voice = voices[Math.floor(Math.random() * voices.length)];
+    const pitch = (Math.random() * (0.9 - 1.125) + 1.125).toFixed(1);
+    const rate = (Math.random() * (1 - 1.125) + 1.125).toFixed(3);
```
erikquinn: Please break out these magic numbers into well-labeled variables so we can understand the math by reading the variable names and how they're used.

src/assets/scripts/client/aircraft/AircraftModel.js
artur-gajewski: We can refactor when we address this issue later.

src/assets/scripts/client/speech.js
```diff
@@ -21,12 +21,39 @@ export const speech_init = () => {
     }
 };
 
+export const randomize_pilot_voice = () => {
```
felixscheffer: should be camel-case without underscores. names with underscores are legacy and should be replaced in the future.

felixscheffer: Also, the function needs a docblock

src/assets/scripts/client/speech.js
```diff
@@ -21,12 +21,39 @@ export const speech_init = () => {
...
+    const voices = [
```
felixscheffer: should be a constant at file level like `SPEECH_VOICES`. same of `SPEECH_HIGHEST_PITCH` and `SPEED_LOWEST_PITCH` (see below).

src/assets/scripts/client/speech.js
```diff
@@ -21,12 +21,39 @@ export const speech_init = () => {
...
+        'Alex',
+        'Daniel',
+        'Fiona',
+        'Google US English',
+        'Google UK English Female',
+        'Google UK English Male'
+    ];
+
+    const voice = voices[Math.floor(Math.random() * voices.length)];
+
```
n8rzz: remove empty line

src/assets/scripts/client/speech.js
```diff
@@ -21,12 +21,39 @@ export const speech_init = () => {
...
+    const lowerPitch = 0.9;
+    const higherPitch = 1.125;
+    const pitch = (Math.random() * (lowerPitch - higherPitch) + higherPitch).toFixed(1);
+
```
n8rzz: remove empty line

src/assets/scripts/client/speech.js
artur-gajewski: I was thinking about keeping the whole file with the same format and refactor whole underscore case when the time comes. Should both formats be mixed until we change all of them to camelcase?

n8rzz: Ordinarily yes, however, my vote here is to leave it for now simply because I have an issue out there to refactor this entire module.

Issue #1230, opened 2018-11-25
upgrade herkou stack (docker/docker-compose?)


n8rzz:
I'm starting to see the following in build logs of review apps:

```
 !   Warning: You are running on a deprecated stack.
 !   Please upgrade to the latest stack by following the instructions on:
 !   https://devcenter.heroku.com/articles/upgrading-to-the-latest-stack
```
This suggests we need to upgrade some things.  And if we're going to upgrade some things we might as well modernize a few things too with docker and/or docker compose.

see:
- https://devcenter.heroku.com/articles/container-registry-and-runtime
- https://devcenter.heroku.com/articles/local-development-with-docker-compose
- https://stackoverflow.com/questions/46904060/how-to-push-docker-containers-managed-by-docker-compose-to-heroku
- https://docs.travis-ci.com/user/docker/#Pushing-a-Docker-Image-to-a-Registry

felixscheffer:
So we are basically talking about running openScope on Heroku in a docker container? If that's the case, that should be super easy to do.

n8rzz:
Yep. Made even easier because in production we use nginx to serve static files.

This task will likely require some updates to our ci build process, but those look pretty straightforward too.

n8rzz:
Current nginx config is used via heroku-static-buildpack, which is a wrapper for nginx 😷.  Our current attempts at configuration can be found in [`static.json`](https://github.com/openscope/openscope/blob/develop/static.json).

The error above calls out a deprecated stack.  according to `heroku apps:info`, we are currently running on `cedar-14`.  yikes, way old.

n8rzz:
this might also be an option (https://github.com/beanieboi/nginx-buildpack).  It at least gets us closer to a real nginx configuration

PR #1235, opened 2018-11-27
devops/1230 - creates Dockerfile for nginx image

n8rzz:
resolves openscope/openscope#1230 

- [ ] create `Dockerfile`
- [ ] push container to heroku registry
- [ ] release container to heroku registry
- [ ] add container + image build to travis-ci
- [ ] publish dockerfile to heroku registry via travis-ci mapped to appropriate environment

The steps above may all be moot if we decide to go with the [nginx-buildpack](https://github.com/beanieboi/nginx-buildpack), but really what's the fun in that?

This gets a `Dockerfile` and basic `nginx.conf` out there.  Still some work to do here yet.

@felixscheffer do you have any experience with deploying a standalone `nginx` container on heroku?  The biggest thing I'm getting tripped up on is the variable port number.  Not really sure how I can both pass that correctly to configuration *_and_* test it locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM nginx:stable
+
+COPY public /usr/share/nginx/html
+COPY nginx.conf /etc/nginx/conf.d/default.conf
--- a/documentation/heroku-docker.md
+++ b/documentation/heroku-docker.md
@@ -1,0 +1,13 @@
+# Using Docker with Heroku
+
+You will need the following in order to complete the steps below:
+1) Docker
+1) Heroku cli
+1) Access to Openscope Heroku
+1) Travis-cli gem
+1) Access to Travis-cli
+
+- `heroku container:login`
+- `heroku container:push openscope:nginx --app openscope-dev`
+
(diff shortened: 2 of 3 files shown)


felixscheffer:
@n8rzz The idea is to pass the port to docker as an env var when you are testing locally (i.e -e port=8080) and then map the port to the host machine (-p 8080:8080).

For nginx part, as far as I know, nginx does not support env vars in its config files, so one has to write the value of the env var (=the port) into nginx's config file with something like sed.

https://ntotten.com/2018/07/22/nginx-on-heroku/

If you want me to, I create a Dockerfile that does that.

n8rzz:
@felixscheffer yeah, I've seen a few examples of both `sed` and port variables.  The more I dug into this, though, I'm not convinced creating our own container is the best solve here.  It may just be as simple as using a different buildpack: https://github.com/beanieboi/nginx-buildpack

I don't want to add to the build just for the sake of adding things.  updating the buildpack may be both easiest and more efficient.  Though I do love the Dockers, don't think thats best here.

n8rzz:
docker doesn't really make sense for us at the moment, closing

PR #1237, opened 2018-11-27
bugfix/1154 - Ensure that the runway is valid for SID before takeoff

### LINKED ISSUE #1154: Ensure that the runway is valid for SID before takeoff
original route: `EDDF25C.OBOKA1W.OBOKA`
- `caf`
- `taxi 25c`
- `route EDDF18.OBOKA1L.OBOKA`
- `to`
-  Aircraft is taking off and is flying the OBOKA1L departure 

Aircraft should complain that runway 25C is not valid for OBOKA1L before takeoff.

felixscheffer:
Resolves #1154

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -229,7 +229,7 @@ export default class AircraftCommander {
      * @return {array} [success of operation, readback]
      */
     runClearedAsFiled(aircraft) {
-        return aircraft.pilot.clearedAsFiled(aircraft);
+        return aircraft.pilot.clearedAsFiled();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -643,6 +643,16 @@ export default class AircraftCommander {
             return [false, 'unable to take off, we never received an IFR clearance'];
         }
 
+        // see #1154, we may have been rerouted since we started taxiing.
```
n8rzz: I like the reference here! 👍

erikquinn:
Minor adjustments made to remove protective logic which is unecessary after adopting the new (much simpler and better) approach.

felixscheffer:
@erikquinn  👍

PR #1248, opened 2018-11-30
enhancement/1232 - Distance based sep for same-rwy departures

### LINKED ISSUE #1232: Subtask: Distance based sep for same-rwy departures
epic openscope/openscope#12

felixscheffer:
Resolves #1232

- [x] Check separation during takeoff
- [x] Check separation before landing
- [x] Add tests for `calculateRunwaySeparationDistanceInFeet`


src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -653,15 +653,25 @@ export default class AircraftCommander {
             return [false, readback];
         }
 
+        // TODO: move somewhere else
```
felixscheffer: I am looking for a better place for this code... but couldn't find one. Open for suggestions.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1335,14 +1355,55 @@ export default class AircraftModel {
...
 
+    /**
+     * Initialize all autopilot systems after being given an IFR clearance to destination and execute takeoff.
+     *
+     * @for Pilot
+     * @method takeoff
+     * @param runway {RunwayModel} the runway taking off on
+     * @param initialAltitude {number} the altitude aircraft can automatically climb to at this airport
+     */
+    takeoff(runway, initialAltitude) {
```
felixscheffer: moved here from Pilot, so that we can remove the Pilot in a distanced future. But I think we have to split AircraftModel sooner or later or we end up with an unmaintainable file.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -653,15 +653,25 @@ export default class AircraftCommander {
...
+        const previousAircraft = this._aircraftController.findAircraftByCallsign(runway.lastAircraftTakenoff);
+        if (previousAircraft) {
+            const actualDistance = aircraft.positionModel.distanceToPosition(previousAircraft.positionModel) * 1000;
+            const requiredDistance = aircraft.model.calculateRunwaySeparationDistanceInFeet(previousAircraft.model);
+
+            if (actualDistance < requiredDistance) {
```
erikquinn: `|| previousAircraft.isOnGround()`

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -653,15 +653,25 @@ export default class AircraftCommander {
             return [false, readback];
         }
 
+        // TODO: move somewhere else
+        const previousAircraft = this._aircraftController.findAircraftByCallsign(runway.lastAircraftTakenoff);
+        if (previousAircraft) {
```
erikquinn: Add new line above

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -473,6 +472,30 @@ export default class AircraftModel {
...
+         */
+        this.mcp = new ModeController();
+
+        /**
+         * @for AircraftModel
+         * @property fms
+         * @type {Fms}
+         * @private
+         */
+        this.fms = new Fms(options);
```
erikquinn: These properties are not private

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -653,15 +653,25 @@ export default class AircraftCommander {
             return [false, readback];
         }
 
+        // TODO: move somewhere else
+        const previousAircraft = this._aircraftController.findAircraftByCallsign(runway.lastAircraftTakenoff);
```
erikquinn: Adding the AircraftController into the AircraftCommander feels wrong for some reason... Perhaps it's better to give the runway access to the AircraftController so it can gain methods to return the aircraft in the runway queue, not just their IDs, and then simply use those methods here? I'm not sure, tbh... @n8rzz?

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -653,15 +653,25 @@ export default class AircraftCommander {
...
+        if (previousAircraft) {
+            const actualDistance = aircraft.positionModel.distanceToPosition(previousAircraft.positionModel) * 1000;
```
erikquinn: This returns the distance in nautical miles, which are about 6000ft. You can use `nm_ft()` to convert and drop the `* 1000`.

src/assets/scripts/client/airport/runway/RunwayModel.js
```diff
@@ -119,6 +119,15 @@ export default class RunwayModel extends BaseModel {
...
 
+        /**
+         * The flight number of the last aircraft that used the runway for takeoff.
+         *
+         * @property lastAircraftTakenoff
+         * @type {string}
+         * @default null
+         */
+        this.lastAircraftTakenoff = null;
+
```
erikquinn: Since this is essentially the same as the queue, but in the past, it should follow the same convention. So either we could have the queue store the aircraft themselves (makes the most sense to me but I think @n8rzz had reasons why not to), or have this property also store the aircraft id only. Methods can be added to allow us to retrieve the aircraft from the controller. Resulting names could be `this.lastDepartedAircraftModel` or `this.lastDepartedAircraftId`, based on the path taken.

src/assets/scripts/client/game/GameController.js
```diff
@@ -31,7 +31,8 @@ const GAME_EVENTS_POINT_VALUES = {
     ILLEGAL_APPROACH_CLEARANCE: -10,
     LOCALIZER_INTERCEPT_ABOVE_GLIDESLOPE: -10,
     NOT_CLEARED_ON_ROUTE: -25,
-    SEPARATION_LOSS: -200
+    SEPARATION_LOSS: -200,
+    NO_TAKEOFF_SEPARATION: -50
```
erikquinn: I like tracking these separately like this, but a deal is a deal-- they should probably all have the same penalty value.

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: `nm_ft()` does not exist.

src/assets/scripts/client/airport/runway/RunwayModel.js
felixscheffer: renamed to `lastDepartedAircraftId`. Adding AircraftModel to RunwayModel creates a circular dependency.

src/assets/scripts/client/game/GameController.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftModel.js
felixscheffer: They weren't before.

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: There is more separation code to come (for example wake turbulence separation), so I definitely won't keep it here. Maybe we should create something like a `SeparationController`?

I was playing with the idea to turn `_aircraftController.findAircraftByCallsign()` into `AircraftController.findAircraftByCallsign()`.

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: moved to ScoreLogic

src/assets/scripts/client/aircraft/AircraftCommander.js
felixscheffer: removed AircraftController

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,139 @@
+import EventBus from '../lib/EventBus';
+import GameController, { GAME_EVENTS } from './GameController';
+import UiController from '../ui/UiController';
+import { AIRCRAFT_EVENT } from '../constants/eventNames';
+import { nm_ft } from '../utilities/unitConverters';
+
+/**
+ * @class ScoreLogic
+ */
+export default class ScoreLogic {
```
felixscheffer: Not the best name. If someone has a better idea, let me know.

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /**
+     * @for ScoreLogic
+     * @method enable
+     * @chainable
+     */
+    enable() {
+        EventBus.on(AIRCRAFT_EVENT.TAKEOFF, this._onTakeoffHandler);
+        EventBus.on(AIRCRAFT_EVENT.LANDING, this._onLandingHandler);
```
felixscheffer: Still looking for a good spot to trigger AIRCRAFT_EVENT.LANDING.

src/assets/scripts/client/aircraft/AircraftModel.js
erikquinn: Sorry, I mean to say that the docblocks show these as `@private`, and they shouldn't because they are public.

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -473,20 +474,54 @@ export default class AircraftModel {
...
 
         this.buildCurrentTerrainRanges();
         this.buildRestrictedAreaLinks();
         this.parse(options);
-        this.initFms(options);
 
-        this.mcp = new ModeController();
-        this.model = new AircraftTypeDefinitionModel(options.model);
-        this.pilot = new Pilot(this.fms, this.mcp);
+        const airport = AirportController.airport_get();
```
felixscheffer: this code was originally in `initFms` which I removed because it wasn't doing anything related to the FMS.

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -235,4 +235,28 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+    /**
+     * Returns the minimal distance that is required to a previous aircraft before another aircraft is allowed to use the runway.
+     *
+     * @for AircraftTypeDefinitionModel
+     * @method calculateRunwaySeparationDistanceInFeet
+     * @param previousTypeModel {AircraftTypeDefinitionModel} the aircraft type that used the runway before us.
+     * @returns {number} distance in feet
+     */
+    calculateRunwaySeparationDistanceInFeet(previousTypeModel) {
+        if (typeof previousTypeModel.category.srs === 'undefined'
```
n8rzz: Let's move this logic to a method

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -235,4 +235,28 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+            || previousTypeModel.category.srs === 3) {
+            return 6000;
```
n8rzz: lets set this to a constant to give it a little more context.  **_what_** does the value `6000` mean?

src/assets/scripts/client/game/ScoreLogic.js
n8rzz: `PlayerScoreController` would be closer to the conventions we're currently using

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,274 @@
...
+            .enable();
+    }
+
+    init() {
+        return this;
+    }
+
+    /**
+     * @for ScoreLogic
+     * @method setupHandler
```
n8rzz: `setupHandler` -> `setupHandlers` (should be plural)

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,274 @@
...
+        if (aircraftModel.mcp.headingMode !== MCP_MODE.HEADING.LNAV) {
+            this._onAirspaceExitWithoutClearance(aircraftModel);
+
+            return;
+        }
+
+        this._onAirspaceExitWithClearance(aircraftModel);
+    }
+
+
```
n8rzz: remove extra empty line

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,274 @@
...
+     * @for ScoreLogic
+     * @method _scoreRunwaySeparation
+     * @param aircraftModel {AircraftModel}
+     * @param runwayModel {RunwayModel}
+     * @param action {string}
+     */
+    _scoreRunwaySeparation(aircraftModel, runwayModel, action) {
+        const previousAircraft = this._aircraftController.findAircraftByCallsign(runwayModel.lastDepartedAircraftId);
+
+        if (previousAircraft) {
```
n8rzz: lets early return here when `!previousAircraft`

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * Display a waring and record an illegal glideslope intercept event
+     *
+     * @for ScoreLogic
+     * @method _penalizeLocalizerInterceptAltitude
+     * @param aircraftModel {AircraftModel}
+     */
+    _penalizeLocalizerInterceptAltitude(aircraftModel) {
+        if (aircraftModel.isAboveGlidepath()) {
```
n8rzz: same here, reverse logic and return early

src/assets/scripts/client/game/ScoreLogic.js
```diff
@@ -0,0 +1,274 @@
...
+     * @method _radioCall
+     * @param msg {string}
+     * @param sectorType {string}
+     * @param isWarning {string}
+     */
+    _radioCall(aircraftModel, msg, sectorType, isWarning = false) {
+        const writtenCallsign = aircraftModel.callsign;
+        const spokenCallsign = aircraftModel.getRadioCallsign();
+
+        const logMessage = (callsign) => `${AirportController.airport_get().radio[sectorType]}, ${callsign} ${msg}`;
```
n8rzz: 🤔 

can we just make this a private method in this class or on the `AirportModel`?

test/game/ScoreLogic.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    const scoreLogic = new ScoreLogic();
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const uiControllerUiLogStub = sandbox.stub(UiController, 'ui_log');
+    const gameControllerRecordEventStub = sandbox.stub(GameController, 'events_recordNew');
+    const expectedLogMessage = `${aircraftModel.getCallsign()} intercepted localizer above glideslope`;
+
+    sandbox.stub(aircraftModel, 'isAboveGlidepath', () => true);
+    scoreLogic._penalizeLocalizerInterceptAltitude(aircraftModel);
+
+    t.true(uiControllerUiLogStub.calledWithExactly(expectedLogMessage, true));
```
n8rzz: if we aren't controlling the returns here, and are simply verifying call params, we should use `spies` instead of `stubs`.

test/game/ScoreLogic.spec.js
```diff
@@ -0,0 +1,59 @@
...
+ava('._penalizeLocalizerInterceptAltitude() does not record an event when at or below glideslope', (t) => {
+    const scoreLogic = new ScoreLogic();
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const uiControllerUiLogStub = sandbox.stub(UiController, 'ui_log');
+    const gameControllerRecordEventStub = sandbox.stub(GameController, 'events_recordNew');
+
+    sandbox.stub(aircraftModel, 'isAboveGlidepath', () => false);
+    scoreLogic._penalizeLocalizerInterceptAltitude(aircraftModel);
+
+    t.true(uiControllerUiLogStub.notCalled);
```
n8rzz: same here, we can use spies instead of stubs for verifying call args

test/game/ScoreLogic.spec.js
```diff
@@ -0,0 +1,59 @@
...
+ava('._penalizeLocalizerInterceptAngle() records an event and notifies the user of their error', (t) => {
+    const scoreLogic = new ScoreLogic();
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const uiControllerUiLogStub = sandbox.stub(UiController, 'ui_log');
+    const gameControllerRecordEventStub = sandbox.stub(GameController, 'events_recordNew');
+    const expectedLogMessage = `${aircraftModel.getCallsign()} approach course intercept angle was greater than 30 degrees`;
+    const result = scoreLogic._penalizeLocalizerInterceptAngle(aircraftModel);
+
+    t.true(typeof result === 'undefined');
+    t.true(uiControllerUiLogStub.calledWithExactly(expectedLogMessage, true));
```
n8rzz: same, spy not stub

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: done

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: done

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: done

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: removed. the code is now using `AircraftModel.radioCall()`

src/assets/scripts/client/game/ScoreLogic.js
felixscheffer: renamed to ScoreController. Not a fan of the term `player` as it suggests openScope is a game

src/assets/scripts/client/aircraft/AircraftModel.js
felixscheffer: Good question. No idea. One thing that I have already done in this PR is to move scoring/procedure related code out of AircraftModel. I think there is probably more code that could be extracted (collision, airspace bust). 

Maybe we could have a pilot class that controls the aircraft (the machine) and an aircraft class that represents the aircraft as a machine. But I haven't really thought about that. Maybe we should also have a flight/flight plan class for things like flight number, origin, and destination.

src/assets/scripts/client/game/ScoreController.js
```diff
@@ -0,0 +1,251 @@
...
+     * @for ScoreController
+     * @method _scoreRunwaySeparation
+     * @param aircraftModel {AircraftModel}
+     * @param runwayModel {RunwayModel}
+     * @param action {string}
+     */
+    _scoreRunwaySeparation(aircraftModel, runwayModel, action) {
+        const previousAircraft = this._aircraftController.findAircraftByCallsign(runwayModel.lastDepartedAircraftId);
+
+        if (previousAircraft) {
```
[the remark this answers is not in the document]
felixscheffer: done

src/assets/scripts/client/game/ScoreLogic.js
erikquinn: done

src/assets/scripts/client/airport/runway/RunwayModel.js
```diff
@@ -119,6 +119,15 @@ export default class RunwayModel extends BaseModel {
...
         this.queue = [];
 
+        /**
+         * The flight number of the last aircraft that used the runway for takeoff.
+         *
+         * @property lastDepartedAircraftId
+         * @type {string}
+         * @default null
+         */
+        this.lastDepartedAircraftId = null;
```
erikquinn: Looks like this has since been changed to use the aircraft's callsign instead of its id number. We should either:
a. Change this property name to `.lastDepartedAircraftCallsign`, or
b. Change all uses of this property to store and expect to receive an aircraft id number instead of a callsign

In both cases, the docblock here should be updated for specificity, as `the flight number` implies a portion of the callsign (`1134 of AAL1134`), when it is actually something else.

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
```diff
@@ -235,4 +235,28 @@ export default class AircraftTypeDefinitionModel extends BaseModel {
...
+        }
+
+        switch (this.category.srs) {
+            case 1:
+                return 3000;
+            case 2:
+                return 4500;
+            default:
+                return 6000;
```
erikquinn: I agree with Nate, the 3000, 4500, and 6000 could be moved to `AIRPORT_CONSTANTS`, where the other similar ones are (or else to a new file for `SEPARATION_CONSTANTS` or something).
```
const AIRPORT_CONSTANTS = {
    ...,
    SRS_REDUCED_MINIMA: {
        CAT_1_IN_BACK: 3000,
        CAT_2_IN_BACK: 4500,
        CAT_3_INVOLVED: 6000
    },
    ...
}
```

src/assets/scripts/client/game/ScoreController.js
```diff
@@ -0,0 +1,253 @@
...
+    /**
+     * @for ScoreController
+     * @method _onAirspaceExitWithClearance
+     * @param aircraftModel {AircraftModel}
+     * @private
+     */
+    _onAirspaceExitWithClearance(aircraftModel) {
+        aircraftModel.radioCall('switching to center, good day', AIRPORT_CONTROL_POSITION_NAME.DEPARTURE);
+        GameController.events_recordNew(GAME_EVENTS.DEPARTURE);
+    }
```
erikquinn: I love all these methods doing this! Sooo much cleaner now, and we actually know where to go to find where this happens :relieved:

src/assets/scripts/client/game/ScoreController.js
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        const actualDistance = nm_ft(aircraftModel.distanceToAircraft(previousAircraft));
+        const requiredDistance = aircraftModel.model.calculateSameRunwaySeparationDistanceInFeet(previousAircraft.model);
+
+        if (actualDistance < requiredDistance || previousAircraft.isOnGround()) {
+            const isWarning = true;
+
+            GameController.events_recordNew(GAME_EVENTS.NO_TAKEOFF_SEPARATION);
+            UiController.ui_log(`${aircraftModel.callsign} ${action} while another aircraft was using the same runway`, isWarning);
```
erikquinn: Slight tweak, if you could: `without adequate separation from another aircraft using the same runway`

src/assets/scripts/client/game/ScoreController.js
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /**
+     * @for ScoreController
+     * @method _onAirspaceExitWithoutClearance
+     * @param aircraftModel {AircraftModel}
+     * @private
+     */
+    _onAirspaceExitWithoutClearance(aircraftModel) {
+        aircraftModel.radioCall('leaving radar coverage without proper clearance', AIRPORT_CONTROL_POSITION_NAME.DEPARTURE, true);
```
erikquinn: Let's take this opportunity to clarify this response (since people always ask about what it means), and change it to `leaving airspace without being on our route`

src/assets/scripts/client/utilities/unitConverters.js
```diff
@@ -159,6 +159,17 @@ export const kn_ms = (kn = 0) => {
...
 };
 
+/**
+ * nautical miles to feet
+ *
+ * @function nm_ft
+ * @param nm {number}
+ * @return {number}
+ */
+export const nm_ft = (nm) => {
```
erikquinn: this should just be a normal function. Making it an arrow function causes some unnecessary overhead which is very insignificant here, but has been very significant elsewhere, so we've stuck to the pattern of using them only when specifically necessary, and instead using `function` and `.bind()` as appropriate for the added efficiency.
```
function nm_ft(nm) {
```

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
felixscheffer: done

src/assets/scripts/client/airport/runway/RunwayModel.js
felixscheffer: done

src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js
felixscheffer: done

src/assets/scripts/client/game/ScoreController.js
felixscheffer: :relaxed:

src/assets/scripts/client/game/ScoreController.js
felixscheffer: done

src/assets/scripts/client/game/ScoreController.js
felixscheffer: done

src/assets/scripts/client/utilities/unitConverters.js
felixscheffer: Basically, I agree with you but the rest of the file uses the "export const"-syntax, so I would like to stick with that syntax or change the whole file to the function-syntax. Up to you.

n8rzz: I don't have a problem with that.  I agree, it probably makes more sense to change all these at once.

src/assets/scripts/client/game/ScoreController.js
```diff
@@ -0,0 +1,253 @@
+import EventBus from '../lib/EventBus';
+import GameController, { GAME_EVENTS } from './GameController';
+import UiController from '../ui/UiController';
+import { AIRCRAFT_EVENT } from '../constants/eventNames';
+import { nm_ft } from '../utilities/unitConverters';
+import AirportController from '../airport/AirportController';
+import { speech_say } from '../speech';
+import { AIRPORT_CONTROL_POSITION_NAME } from '../constants/airportConstants';
+import { MCP_MODE } from '../aircraft/ModeControl/modeControlConstants';
```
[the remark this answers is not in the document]
felixscheffer: is this documented somewhere?

src/assets/scripts/client/AppController.js
```diff
@@ -21,6 +21,7 @@ import { speech_init } from './speech';
 import { EVENT } from './constants/eventNames';
 import { SELECTORS } from './constants/selectors';
 import { TRACKABLE_EVENT } from './constants/trackableEvents';
+import ScoreController from './game/ScoreController';
```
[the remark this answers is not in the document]
felixscheffer: done

test/game/ScoreLogic.spec.js
felixscheffer: done

test/game/ScoreLogic.spec.js
felixscheffer: @n8rzz I tried to replace the stubs with spies but I get `Attempted to wrap ui_log which is already wrapped`. Maybe you can get it to work.

n8rzz: you'll usually see a warning like that when some other test has stubbed a method.  the way to resolve that is to make sure that in any test where you _are_ using stubs, you also call `someMethodStub.restore()` after your assertions have been called.

src/assets/scripts/client/game/ScoreController.js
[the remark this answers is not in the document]
n8rzz: sadly, no.  Though I wish `eslint` could handle something like this or, if it can, that we were set up to enforce this order with `eslint`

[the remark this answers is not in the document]
felixscheffer: That's unfortunate. Can we add this to our own documentation? Maybe to the tools/README.md.

PR #1257, opened 2018-12-08
bugfix/1240 - Fix readback from departures in runway queue is not correct

### LINKED ISSUE #1240: Readback from departures in runway queue is not correct
The messages about not being first in line for the runway (when trying to clear someone for takeoff) are not correctly getting the aircraft callsign, because they expect the runway queue to contain instances of `AircraftModel`, when in fact, they contain the aircraft IDs. The aircraft can be looked up by these IDs instead to achieve the desired result.

#### Steps to reproduce
1. Taxi a departure to a runway.
1. Taxi another departure to the same runway.
1. Wait at least three seconds to allow them to both reach the runway).
1. Clear the _second_ aircraft for takeoff (this will be the one whose data block is _not_ shown).
1. Observe the response using `undefined`, etc, instead of correct callsigns.

felixscheffer:
Resolves #1240

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -383,16 +383,33 @@ export default class AircraftController {
     }
 
     /**
+     * Finds an aircraft by its callsign
+     *
      * @method findAircraftByCallsign
      * @param  {string} [callsign='']
      * @return {AircraftModel|null}
      */
     findAircraftByCallsign(callsign = '') {
-        if (callsign === '') {
+        if (!callsign) {
             return null;
         }
 
         return _find(this.aircraft.list, (aircraft) => aircraft.callsign.toLowerCase() === callsign.toLowerCase());
+    }
+
+    /**
+     * Finds an aircraft by its internal id
+     *
+     * @method findAircraftById
+     * @param  {string} [id]
+     * @return {AircraftModel|null}
+     */
+    findAircraftById(id) {
+        if (!id) {
+            return null;
+        }
+
+        return _find(this.aircraft.list, (aircraft) => aircraft.id === id);
     }
 
     /**
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -75,8 +75,8 @@ export default class AircraftController {
     constructor(aircraftTypeDefinitionList, airlineController, scopeModel) {
         if (isEmptyOrNotArray(aircraftTypeDefinitionList)) {
             // eslint-disable-next-line max-len
-            throw new TypeError('Invalid aircraftTypeDefinitionList passed to AircraftTypeDefinitionCollection. ' +
-                `Expected and array but received ${typeof aircraftTypeDefinitionList}`);
+            throw new TypeError('Invalid aircraftTypeDefinitionList passed to AircraftTypeDefinitionCollection. '
+                + `Expected and array but received ${typeof aircraftTypeDefinitionList}`);
         }
 
         // TODO: this may need to use instanceof instead, but that may be overly defensive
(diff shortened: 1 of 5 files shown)


src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,18 +383,35 @@ export default class AircraftController {
...
     /**
+     * Finds an aircraft by its callsign
+     *
      * @method findAircraftByCallsign
      * @param  {string} [callsign='']
      * @return {AircraftModel|null}
      */
     findAircraftByCallsign(callsign = '') {
-        if (callsign === '') {
+        if (!callsign) {
```
n8rzz: I like this better, but perhaps we should change the default assignment to `null` then instead of an empty string?

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,18 +383,35 @@ export default class AircraftController {
...
             return null;
         }
 
         return _find(this.aircraft.list, (aircraft) => aircraft.callsign.toLowerCase() === callsign.toLowerCase());
```
n8rzz: could you add a `TODO` here to normalize `callsign` casing?  in my opinion, the only thing we should need to lowerCase here would be the passed value

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,18 +383,35 @@ export default class AircraftController {
...
     }
 
+    /**
+     * Finds an aircraft by its internal id
+     *
```
n8rzz: _why_ would be need to find and aircraft by an internal ID?  don't we reference aircraft by their callsigns? 

I know the answer here, but we should document this reason for our future selves.

felixscheffer: That's good because I have no idea why we use the internal ID in the runway queue. 🙈

src/assets/scripts/client/aircraft/AircraftController.js
felixscheffer: Good point

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,16 +383,35 @@ export default class AircraftController {
...
-     * @param  {string} [callsign='']
+     * @param  {string} [callsign=null]
+     * @return {AircraftModel|null}
+     */
+    findAircraftByCallsign(callsign = null) {
+        if (!callsign) {
+            return null;
+        }
+
+        const normalizedCallSign = callsign.toUpperCase();
```
felixscheffer: @n8rzz I tried to evolve your previous comment here (see `AircraftModel get callsign()`)

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: I don't remember either.. I think it was some circular dependency issue, where we added the id because we couldn't store the AircraftModel? I dunno.

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: addressed below

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,16 +383,35 @@ export default class AircraftController {
...
+     *
+     * @method findAircraftById
+     * @param  {string} [id]
      * @return {AircraftModel|null}
      */
-    findAircraftByCallsign(callsign = '') {
-        if (callsign === '') {
+    findAircraftById(id) {
+        if (!id) {
             return null;
```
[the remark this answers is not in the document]
erikquinn: done

PR #1262, opened 2018-12-26
bugfix/1259 - Fixes lack of heavy/super designation on vox

### LINKED ISSUE #1259: Voice does not call heavy/super aircraft "heavy" or "super"
The speech synthesis voice should be calling heavy aircraft `American one twenty three heavy`, but just calls it `American one twenty three`. When the aircraft is of weightclass "super" (`U`), it would be `super` instead of `heavy`.

kmartino2168:
The issue was caused by capitalization inconsistencies throughout the codebase (`weightclass` vs `weightClass`).

Resolves #1259


felixscheffer:
@n8rzz offtopic question: why do we track `assets/aircraft/aircraft.json` in git? It looks like a file that is created during the build process.

PR #1274, opened 2018-12-30
bugfix/1158 - fixes Firefox strip bay scrollbars not working

### LINKED ISSUE #1158: Firefox strip bay not hiding the scrollbars
The scrollbars in the strip bay are supposed to be styled nicely, and hidden when not needed. In Firefox (61.0.1), they appear at all times anyway, and are icky looking.

image

felixscheffer:
Resolves #1158 

Also contains the fix for `this._aircraftController is undefined`

Ideally, we can get this into 6.9.0


erikquinn:
👍 We can always merge bugfixes during the testing period, just no new features 😃

PR #1277, opened 2018-12-31
enhancement/1275

### LINKED ISSUE #1275: Add time to build information in html
Related: #1272.

The build information shown in the HTML has only the date of the build, but the time would be handy (say we just pushed a change, and somebody refreshed the app before it finished rebuilding-- when they say "it's still broken", it'd be nice to be able to verify that both the user and developer are running the same build).

image

n8rzz:
Resolves #1275 

Updates `buildMarkup` task to output date and time when generating the `index.html` file


n8rzz:
currently using `UTC` time, to avoid all the timezone shenanigans.  Result looks like:
screen shot 2018-12-30 at 10 51 07 pm

PR #1281, opened 2018-12-31
bugfix/1279 - Fix erroneous altitudes appearing in flight strip

### LINKED ISSUE #1279: Altitude commands not processed correctly
Clear somebody to 15000 and you'll see some odd responses. It appears to be that arguments are being improperly passed through the InputController to the AircraftCommander due to some recent change.

image

erikquinn:
Resolves #1279.


erikquinn:
More of those heap out of memory errors... This fixes the issue though

PR #1283, opened 2019-01-02
feature/1234 - adds "toggle halo" command

### LINKED ISSUE #1234: Add "toggle halo" command
originally by @erikquinn:

With a halo "on", it is showed regardless. When it's off, irl, it is never shown.

The issue is that we show a dim halo when two aircraft are "in conflict", when that is actually shown by flashing irl. The change to a flash during conflict doesn't need to happen now, but one important distinction is that the halo turned on by the user should be fairly bright (much brighter than the current "conflict halo" currently appears). So the two are unrelated.

The radius should be definable in a scope command:
HALO draws a 3nm ring (or removes it if one is present)
HALO 5 draws a 5nm ring
...etc. If an aircraft has a ring of one dimension and receives a command specifying a new dimension, the old halo is replaced with the new one.

See openscope/openscope#14 for the planned key mappings for these commands as well.

inspired by openscope/openscope#1217

felixscheffer:
Resolves #1234

- [ ] Halo radius should be part of the command, with a default value set in the settings dialog
- [x] ~~Needs a better readback (or none at all) than "TOGGLE HALO"~~
- [ ] When a halo exists, and we enter a command to change its size, respond `RESIZE HALO`
- [x] Added support for F7 key
- [x] Changed from jQuery.which to KeyboardEvent.code. Hopefully, this will be platform-independent.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code


src/assets/scripts/client/InputController.js
```diff
@@ -414,17 +415,26 @@ export default class InputController {
                 this.onCommandInputChangeHandler();
 
                 break;
+*/
+/*
             case KEY_CODES.DASH: // mac -
                 this.$commandInput.val(`${currentCommandInputValue} - `);
```
felixscheffer: why are we doing this?

src/assets/scripts/client/constants/inputConstants.js
```diff
@@ -12,34 +12,45 @@ export const COMMAND_CONTEXT = {
...
+    F6: 'F6',
+    F7: 'F7',
+    F8: 'F8',
+    F9: 'F9',
+    F10: 'F10',
+    F11: 'F11',
+    F12: 'F12',
     // `
-    BAT_TICK: 192
+    BAT_TICK: 'Backquote'
```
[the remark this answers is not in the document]
n8rzz: grave accent according to: https://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes

n8rzz:
For reference, our browser usage for the month of December 2018:

screen shot 2019-01-04 at 12 13 30 pm

src/assets/scripts/client/InputController.js
```diff
@@ -331,7 +331,7 @@ export default class InputController {
         const currentCommandInputValue = this.$commandInput.val();
 
         // TODO: this swtich can be simplified, there is a lot of repetition here
-        switch (event.which) {
+        switch (event.originalEvent.code) {
```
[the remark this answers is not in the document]
felixscheffer: Can we use `key` instead?
https://caniuse.com/#feat=keyboardevent-key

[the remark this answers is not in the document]
felixscheffer: 1. You are correct. It breaks IE support. I didn't notice that :sob:
2. I dropped the idea to use the `key` attribute

src/assets/scripts/client/constants/inputConstants.js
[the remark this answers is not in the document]
felixscheffer: Are we talking about this key?

src/assets/scripts/client/InputController.js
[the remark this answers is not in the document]
felixscheffer: IE/Edge should be working again.

src/assets/scripts/client/constants/inputConstants.js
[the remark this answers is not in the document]
felixscheffer: Great, then 'Backquote' should be the right code. Fun fact: in IE '192' is the semicolon key.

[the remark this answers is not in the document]
felixscheffer: fixed bat tick key code in Edge.

I also tried to test in IE11 but the sim isn't even rendering in IE11 😭

erikquinn:
Yep, that's the idea. They're mimicking system responses you would get from entering real scope commands.

erikquinn:
@felixscheffer Do you want to add the radius parameter to the halo command now, or in a later PR?

erikquinn:
@felixscheffer Do you want to add the radius parameter to the halo command now, or in a later PR?

felixscheffer:
Let's do it in this PR

erikquinn:
@felixscheffer Can you have a look through the changes I added and advise if you're happy with them? Some minor adjustments, and adds the ability to specify the size of the halo. By Monday night if you can, as that's the last day for merging new features. Thanks 😃

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1009,16 +1029,15 @@ export default class CanvasController {
...
 
         cc.translate(aircraftCanvasPosition.x, aircraftCanvasPosition.y);
 
         this._drawAircraftVectorLines(cc, aircraftModel);
-
-        if (aircraftModel.notice || alerts[0]) {
-            this._drawAircraftRings(cc, aircraftModel);
-        }
+        this._drawAircraftHalo(cc, radarTargetModel);
+        this._drawAircraftConflictRings(cc, radarTargetModel);
```
felixscheffer: @erikquinn I changed the order to ensure that violations are visible. i.e. violations are rendered on top of the halo (not the other way round).

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -480,30 +504,58 @@ export default class RadarTargetModel {
...
      *
      * @for RadarTargetModel
-     * @method toggleHalo
+     * @method setHalo
+     * @param radius {number}
      * @return {array} [success of operation, system's response]
      */
-    toggleHalo() {
-        this._hasHalo = !this._hasHalo;
+    setHalo(radius) {
```
felixscheffer: I changed the name to `setHalo` to better reflect what the method is doing and to be in line with `removeHalo`. Maybe `toggleHalo` was already the best name.

documentation/scope-commands.md
```diff
@@ -49,17 +49,17 @@ _More Info -_ On real ATC systems, moving the data block is sometimes used by co
...
-~~_Shortcut -_ `[F7] [CID]`~~
+### Toggle Halo
+_Syntax -_ `QP_J [CID]` or `QP_J [radius] [CID]`
+_Shortcut -_ `[F7] [CID]` or `[F7] [radius] [CID]`
 
-~~_Description -_ This toggles a circle around the center of the radar target, with a radius definable in the scope's theme (generally 3nm for terminal, or 5nm for en route).~~
+_Description -_ This toggles a circle around the center of the radar target, with a radius you can specify in the command. If the radius is not specified, the default halo size will be used (generally 3nm for terminal, or 5nm for en route).
 
-~~_More Info -_ **THIS COMMAND IS NOT YET AVAILABLE**~~
+_More Info -_ Note that if a resized halo exists, entering `[F7] [CID]` will actually be interpreted by the system as `[F7] 3 [CID]`-- meaning the halo will be resized to 3nm, not removed. With the halo back to 3nm, entering `[F7] [CID]` again will remove the halo. If you add a 10nm halo with `[F7] 10 [CID]`, it can be removed instead of resized by specifying `[F7] 10 [CID]`, or use the above method of entering `[F7] [CID]` twice.
```
felixscheffer: why 10nm?

src/assets/scripts/client/canvas/CanvasController.js
erikquinn: It was intentional at the time, as thats often why I would have a halo up, but after thinking about it, I agree that the violation should be drawn on top instead of hidden by the halo.

src/assets/scripts/client/scope/RadarTargetModel.js
erikquinn: Eh, that'll work. Better than toggle either way, because it now will adjust the halo size if you do `.setHalo(4); .setHalo(7);`, so it doesn't always "toggle" it on and off each call anymore.

That should be fine :+1:

documentation/scope-commands.md
erikquinn: It's the standard behavior, per the vERAM user guide. It's not inherently 10, but rather that an attempt to draw a halo when one of that size currently exists will result in it being removed.
```
QP_J THY7 (draws 3nm ring)
QP_J 6 THY7 (adjusts ring to 6nm)
QP_J THY7 (adjusts to 3nm ring)
QP_J THY7 (removes ring)

QP_J 10 THY8 (draws 10 mile ring)
QP_J 10 THY8 (removes ring)
```

erikquinn:
@felixscheffer Great, thanks for the feedback. All of the changes are good beside the one listed above. Would you say this is good to go, or is there anything further?

documentation/scope-commands.md
felixscheffer: My bad. I think I misread the text. Works for me

felixscheffer:
:shipit:

PR #1297, opened 2019-01-26
feature/1296 - allow defining circular restricted areas by center + r…

### LINKED ISSUE #1296: Allow restricted airspace definition as a circle
There is a lot of restricted areas in Russian airspace defined like 'A circle radius of 3 KM centred at 481815N 0401222E'.

grandsbor:
Resolves #1296

### DIFF
--- a/src/assets/scripts/client/airport/AirportModel.js
+++ b/src/assets/scripts/client/airport/AirportModel.js
@@ -20,6 +20,7 @@ import {
     cos,
     round
 } from '../math/core';
+import { tau } from '../math/circle';
 import {
     vlen,
     vsub,
@@ -465,6 +466,27 @@ export default class AirportModel {
 
     /**
      * @for AirportModel
+     * @method prepareCircularRestrictedArea
+     * @param center {array<number>}
+     * @param radius {number}   circle radius, in nm
+     */
+    getCircularCoordinates(center, radius) {
+        const NUM_POINTS = 32;
+        const delta = tau() / NUM_POINTS;
+        let coords = [];
+        let centerPos = new StaticPositionModel(center, this._positionModel, this._magneticNorth);
+
+        for (let i = 0; i < NUM_POINTS; i++) {
+            let bearing = delta * i;
+            const rawPoint = centerPos.generateCoordinatesFromBearingAndDistance(bearing, radius);
+            coords.push(DynamicPositionModel.calculateRelativePosition(rawPoint, this._positionModel, this.magneticNorth));
+        }
+
+        return coords;
+    }
+
+    /**
+     * @for AirportModel
      * @method buildRestrictedAreas
      * @param restrictedAreas
      */
@@ -482,26 +504,32 @@ export default class AirportModel {
             }
 
             obj.height = parseElevation(area.height);
-            obj.coordinates = _map(
-                area.coordinates,
-                (v) => DynamicPositionModel.calculateRelativePosition(v, this._positionModel, this.magneticNorth)
-            );
 
-            let coords_max = obj.coordinates[0];
-            let coords_min = obj.coordinates[0];
+            if (area.radius && area.center) {
+                obj.center = DynamicPositionModel.calculateRelativePosition(area.center, this._positionModel, this.magneticNorth);
+                obj.coordinates = this.getCircularCoordinates(area.center, area.radius);
+            } else {
+                obj.coordinates = _map(
+                    area.coordinates,
+                    (v) => DynamicPositionModel.calculateRelativePosition(v, this._positionModel, this.magneticNorth)
+                );
 
-            _forEach(obj.coordinates, (v) => {
-                coords_max = [
-                    Math.max(v[0], coords_max[0]),
-                    Math.max(v[1], coords_max[1])
-                ];
-                coords_min = [
-                    Math.min(v[0], coords_min[0]),
-                    Math.min(v[1], coords_min[1])
-                ];
-            });
+                let coords_max = obj.coordinates[0];
+                let coords_min = obj.coordinates[0];
 
-            obj.center = vscale(vadd(coords_max, coords_min), 0.5);
+                _forEach(obj.coordinates, (v) => {
+                    coords_max = [
+                        Math.max(v[0], coords_max[0]),
+                        Math.max(v[1], coords_max[1])
+                    ];
+                    coords_min = [
+                        Math.min(v[0], coords_min[0]),
+                        Math.min(v[1], coords_min[1])
+                    ];
+                });
+
+                obj.center = vscale(vadd(coords_max, coords_min), 0.5);
+            }
 
             this.restricted_areas.push(obj);
         });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Enhancements & Refactors
 - [#1247](https://github.com/openscope/openscope/issues/1247) - Determine initial climb altitude from the SID
 - [#1290](https://github.com/openscope/openscope/issues/1290) - Add Pull Reminders badge to repository readme
+- [#1296](https://github.com/openscope/openscope/issues/1296) - Allow defining restricted areas as circles
 
 
 # 6.9.1 (January 4, 2019)
--- a/documentation/airport-format.md
+++ b/documentation/airport-format.md
@@ -395,6 +395,18 @@ Areas of restricted airspace may be added to the `restricted` property of the ai
 ],
 ```
 
+An area may be defined as an array of points as in example above, or as a circle with known center and radius (in nm):
+```javascript
+"restricted": [
+    {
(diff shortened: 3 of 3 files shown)


src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,27 @@ export default class AirportModel {
...
+    /**
+     * @for AirportModel
+     * @method prepareCircularRestrictedArea
+     * @param center {array<number>}
+     * @param radius {number}   circle radius, in nm
+     */
+    getCircularCoordinates(center, radius) {
+        const NUM_POINTS = 32;
+        const delta = tau() / NUM_POINTS;
+        let coords = [];
```
felixscheffer: const

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,27 @@ export default class AirportModel {
...
+        let centerPos = new StaticPositionModel(center, this._positionModel, this._magneticNorth);
```
felixscheffer: const

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,27 @@ export default class AirportModel {
...
+
+        for (let i = 0; i < NUM_POINTS; i++) {
+            let bearing = delta * i;
```
felixscheffer: const

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,27 @@ export default class AirportModel {
...
+            const rawPoint = centerPos.generateCoordinatesFromBearingAndDistance(bearing, radius);
+            coords.push(DynamicPositionModel.calculateRelativePosition(rawPoint, this._positionModel, this.magneticNorth));
```
felixscheffer: please add an empty line before this line

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,27 @@ export default class AirportModel {
         });
     }
 
+    /**
+     * @for AirportModel
+     * @method prepareCircularRestrictedArea
+     * @param center {array<number>}
+     * @param radius {number}   circle radius, in nm
+     */
+    getCircularCoordinates(center, radius) {
```
felixscheffer: should be private IMHO 

@n8rzz

src/assets/scripts/client/airport/AirportModel.js
erikquinn: @grandsbor Arrays can have their _contents_ modified through `.push()`, `.pop()`, `.unshift()`, etc, even if they are a const. The const will only ensure that there are no _reassignments_ of the value, such as later doing `coords = 'words';` or `coords = [3.14159].

You should actually be seeing a lint error that points out the preference to use const here. If you install eslint addons for your IDE, it will tell you any issues like these:

image

grandsbor: Thanks!
(Never written in Node before)

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -463,6 +464,30 @@ export default class AirportModel {
...
 
+    /**
+     * @for AirportModel
+     * @method prepareCircularRestrictedArea
+     * @param center {array<number>}
+     * @param radius {number}   circle radius, in nm
+     * @return {array}
+     * @private
+     */
+    _getCircularCoordinates(center, radius) {
```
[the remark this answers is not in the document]
grandsbor: Would it be ok if I put it into `navigationUtilities.js`? I fail to see any other appropriate files in `utilities/`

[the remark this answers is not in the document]
erikquinn: Yeah, you're right, none of those are great. It's not navigation related really, but I agree it's the best place for now. We can reorganize them at a later time.

grandsbor:
Tests added, code moved to utilities.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -473,6 +474,8 @@ export default class AirportModel {
             return;
         }
 
+        const airportPositionAndDeclination = [this.positionModel, this.magneticNorth];
```
felixscheffer: @n8rzz what do you think?

n8rzz: not good, as I'm sure you can imagine.  Why not just create a getter for this?

```js
get airportPositionAndDeclination() {
    return [this.positionModel, this.magneticNorth];
}
```

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -482,26 +485,32 @@ export default class AirportModel {
...
-                    Math.max(v[1], coords_max[1])
-                ];
-                coords_min = [
-                    Math.min(v[0], coords_min[0]),
-                    Math.min(v[1], coords_min[1])
-                ];
-            });
+            if (area.radius && area.center) {
+                obj.center = DynamicPositionModel.calculateRelativePosition(area.center, ...airportPositionAndDeclination);
+                obj.coordinates = getCircularCoordinates(area.center, area.radius, airportPositionAndDeclination);
```
felixscheffer: @n8rzz follow-up (including the line above)

test/utilities/navigationUtilities.spec.js
```diff
@@ -10,3 +17,26 @@ ava('assembleProceduralRouteString() concatenates provided strings with appropri
 
     t.true(result === expectedResult);
 });
+
+ava('getCircularCoordinates() provides a set of points equidistant from center', (t) => {
+    const TOLERABLE_ERROR = 0.025  // 25m
+
+    const posAndDecl = [airportModelFixture.positionModel, MAGNETIC_NORTH_MOCK];
```
n8rzz: use full words for variables

test/utilities/navigationUtilities.spec.js
```diff
@@ -10,3 +17,26 @@ ava('assembleProceduralRouteString() concatenates provided strings with appropri
...
+    const center = GPS_COORDINATES_MOCK;
```
n8rzz: You shouldn't reassign and rename constant like this.  you're loosing context and opening the door to modification.  just use `GPS_COORDINATES_MOCK`

src/assets/scripts/client/airport/AirportModel.js
erikquinn: Doesn't the real issue lie with the way the parameters are being passed to getCircularCoordinates()? I don't see any reason to create and pass an array like this instead of passing the two parameters directly, eg

`getCircularCoordinates(area.center, area.radius, airportPosition, magneticDeclination);`

erikquinn: Also, since this is not returning coordinates, but rather instances of `StaticPositionModel`, it ought to be renamed to reflect that. Perhaps `generatePositionsFromCircle()` or something.

test/utilities/navigationUtilities.spec.js
```diff
@@ -10,3 +17,26 @@ ava('assembleProceduralRouteString() concatenates provided strings with appropri
...
+    const radius = 1.0;
+
+    const circle = getCircularCoordinates(center, radius, posAndDecl);
+    t.true(circle.length === 32);
+
+    const trueCenter = DynamicPositionModel.calculateRelativePosition(center, ...posAndDecl)
+
+    let ok = true;
```
n8rzz: you can simplify this entire block using `t.plan()` (https://github.com/avajs/ava/blob/master/docs/03-assertions.md#assertion-planning):

```js
t.plan(circle.length);

for (let i = 0; i < circle.length; i++) {
    const distance = distance2d(trueCenter, circle[i]);
    const isWithinTolerance = abs(distance - km(1)) <= TOLERABLE_ERROR;

    t.true(isWithinTolerance);
}
```

src/assets/scripts/client/airport/AirportModel.js
n8rzz: @erikquinn you know where my head is at there 😉 My fault for not looking very carefully.

I'd much prefer only passing what we need, that is certainly the better way to go.  And I agree on your second point as well.  Descriptive names are better names.

test/utilities/navigationUtilities.spec.js
```diff
@@ -10,3 +17,26 @@ ava('assembleProceduralRouteString() concatenates provided strings with appropri
 
     t.true(result === expectedResult);
 });
+
+ava('getCircularCoordinates() provides a set of points equidistant from center', (t) => {
+    const TOLERABLE_ERROR = 0.025  // 25m
```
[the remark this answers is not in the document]
n8rzz: @erikquinn @felixscheffer ^^

PR #1302, opened 2019-01-28
feature/1272 - add version to settings menu

### LINKED ISSUE #1272: Add simulator version number to settings menu
Related: #1275.

Like we have a title on the airport selection dialog, the settings should have a title. And more importantly, we should somewhere in the settings window display the version of the sim so people have more than just the splashscreen and F12 to find the version number. Should also help with bug reports.
image
image

sam-irl:
Resolves #1272 

The purpose of this pull request is to add the sim's version number to the settings menu.

~~Also adds a `getVersion` method to the `GameController`~~

### DIFF
--- a/src/assets/scripts/client/game/GameController.js
+++ b/src/assets/scripts/client/game/GameController.js
@@ -8,7 +8,7 @@ import TimeKeeper from '../engine/TimeKeeper';
 import { round } from '../math/core';
 import { EVENT } from '../constants/eventNames';
 import { GAME_OPTION_NAMES } from '../constants/gameOptionConstants';
-import { TIME, INVALID_VERSION } from '../constants/globalConstants';
+import { TIME } from '../constants/globalConstants';
 import { TRACKABLE_EVENT } from '../constants/trackableEvents';
 import { SELECTORS } from '../constants/selectors';
 import { THEME } from '../constants/themes';
--- a/src/assets/scripts/client/game/GameController.js
+++ b/src/assets/scripts/client/game/GameController.js
@@ -8,7 +8,7 @@ import TimeKeeper from '../engine/TimeKeeper';
 import { round } from '../math/core';
 import { EVENT } from '../constants/eventNames';
 import { GAME_OPTION_NAMES } from '../constants/gameOptionConstants';
-import { TIME } from '../constants/globalConstants';
+import { TIME, INVALID_VERSION } from '../constants/globalConstants';
 import { TRACKABLE_EVENT } from '../constants/trackableEvents';
 import { SELECTORS } from '../constants/selectors';
 import { THEME } from '../constants/themes';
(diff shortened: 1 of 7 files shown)


src/assets/scripts/client/game/GameController.js
```diff
@@ -608,6 +610,23 @@ class GameController {
...
+
+    /**
+     * Retrieves the current version of the simulator from the `window` global object.
+     *
+     * @for GameController
+     * @method _getVersion
+     * @return {string}
+     */
+    _getVersion() {
+        const version = window.GLOBAL.VERSION;
```
n8rzz: is there ever a case where this wouldn't actually be set?  This is set via build tool, so I don't think it would *never* be defined

sam-irl: Probably not, I was just being needlessly overcautious

Also sorry for requesting review, just force of habit (it's pretty much done anyway)

n8rzz: reviews are required, so you did good there 😉

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -127,4 +130,24 @@ export default class SettingsController {
...
+     * Build the html for the simulator version psuedo-option.
+     *
+     * @for SettingsController
+     * @method _buildVersionTemplate
+     * @return {JQuery|HTML element}
+     */
+    _buildVersionTemplate() {
+        const simulatorVersion = window.GLOBAL.VERSION;
+        const $container = $(UI_OPTION_CONTAINER_TEMPLATE);
+        const $version = $(UI_OPTION_SELECTOR_TEMPLATE);
```
felixscheffer: IMHO version should have its own template

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -127,4 +130,24 @@ export default class SettingsController {
...
+        
+        $version.text(simulatorVersion);
+        $version.addClass('simulator-version');
```
felixscheffer: the css class should be part of the template. Also, I would prefer a class name like `option-static-text` or something. Maybe we are going to add more text someday.

n8rzz:
I'm happy with how this is written and how it works.  I think, though, that we can do some small tweaks to the styles:

Currently this looks like:
screen shot 2019-01-28 at 8 07 36 pm

Since this is an effective subtitle, or aside of some sort, I feel we should de-emphasize the text a little bit so it is differentiated from the other options.  Maybe something like:
screen shot 2019-01-28 at 8 04 51 pm

erikquinn:
Closed #1303 so we can continue here.

felixscheffer:
I added the styling @n8rzz suggested. Is there anything else left to do in this pr?

n8rzz:
No concerns from me. :shipit:

PR #1306, opened 2019-01-29
bugfix/1305 - fixes range rings center is undefined

### LINKED ISSUE #1305: Range rings center is undefined

felixscheffer:
Resolves #1305

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 
 ### Bugfixes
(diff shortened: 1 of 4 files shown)


erikquinn:
What are some steps to reproduce the issue this fixes?

erikquinn:
bump ^

erikquinn:
Yeah, I just never know if we've fixed it all if I can't black and white see the before/after. Nonetheless, I agree, so here we go

PR #1307, opened 2019-01-30
enhancement/391 - adds support for multiple airspace sections

### LINKED ISSUE #391: Add support for multiple airspace sections
Currently, only the first entry within `airspace` is used for anything. Support the ability of an airport contributor to add airspace in multiple sections to achieve the appropriate stratification of the real-world airspace.

felixscheffer:
Resolves #391

- works for EDDF and KSEA.

Todo:
- [ ] Update / check airspaces for all airports
- [x] Remove / replace the `perimeter` property in `AirportModel` (with airspace)
- [x] Fix EDDF airspace
- [x] Add altitude to `isPointWithinAirspace`


felixscheffer:
@erikquinn can you deploy this branch please?

erikquinn:
@felixscheffer Building now!

felixscheffer:
@Marcel510 could you have a look at  EDDF's airspace? The airspace (`_airspace` property) doesn't seem to have the same shape as the video map.

src/assets/scripts/client/math/flightMath.js
```diff
@@ -144,17 +126,19 @@ export function isWithinAirspace(airport, pos) {
  * @return {boolean}
  */
 export function calculateDistanceToBoundary(airport, pos) {
+    // TODO: remove perimeter
     const perim = airport.perimeter;
```
felixscheffer: @erikquinn Can we come up with a solution that does not require a `perimeter` property? maybe something based on the actual airspace.

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -101,7 +100,7 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
             continue;
         }
 
-        if (isWithinAirspace(airport, waypointModel.relativePosition)) {
+        if (airport.isPointWithinAirspace(waypointModel.relativePosition)) {
```
felixscheffer: Not ideal...should be 3D, not 2D. :see_no_evil:

src/assets/scripts/client/math/flightMath.js
erikquinn: Probably? It was only intended to reduce the number of "is this aircraft within this polygon?" calculations. If we had a whole bunch of polygons that defined many shelves of airspace, we might have to do an enormous number of these calculations, each of which are probably extremely expensive. My original thought was having a perimeter would mean that anybody outside of the outermost extents of the airspace can at least only be compared against a single polygon instead of many. Perhaps this could be generated from the others mathematically at instantiation to have the same effect? Could be tough though.

We just want to make sure we're doing everything we can to keep the sim running smoothly, and determining whose airspace an aircraft is in without it taking too long and causing the whole thing to slow down.

src/assets/scripts/client/airport/AirspaceModel.js
```diff
@@ -102,34 +113,46 @@ export default class AirspaceModel extends BaseModel {
         this.airspace_class = '';
     }
 
+    transformPoly() {
```
n8rzz: missing docblock.  I'd really love a description of this one, what is this and why is it important?

src/assets/scripts/client/airport/AirspaceModel.js
```diff
@@ -102,34 +113,46 @@ export default class AirspaceModel extends BaseModel {
...
+        if (!this.isPointInside2D(point)) {
+            return false;
         }
 
-        return polyPositionModels;
+        return this.floor <= altitude && altitude <= this.ceiling;
+    }
+
+    /**
+     *
```
n8rzz: please adda description here

src/assets/scripts/client/airport/AirspaceModel.js
felixscheffer: added docblock. The description could be better though

src/assets/scripts/client/airport/AirspaceModel.js
felixscheffer: done

src/assets/scripts/client/trafficGenerator/buildPreSpawnAircraft.js
```diff
@@ -101,21 +97,45 @@ const _calculateDistancesAlongRoute = (waypointModelList, airport) => {
             continue;
         }
 
-        if (isWithinAirspace(airport, waypointModel.relativePosition)) {
-            distanceFromClosestFixToAirspaceBoundary = nm(calculateDistanceToBoundary(airport, waypointModel.relativePosition));
-            totalDistance += distanceFromClosestFixToAirspaceBoundary;
+        const distanceBetweenWaypoints = distance2d(previousWaypoint.relativePosition, waypointModel.relativePosition);
```
erikquinn: I knew I should have finished that damn spawn method fixing PR.... these will be not so fun to resolve conflicts on... but hey, that's what I get! 😢

erikquinn (dismissed):
1. This is f***ing awesome beyond words. A _huge_ step forward in what we'll be able to simulate. Awesome!

2. We need a way to display the _altitudes_ in each shelf, so people can easily view them when they're working. Formatted like `000-120`, `030-070`, etc.

3. We need a way to display and hide the overlay which shows the boundaries between the shelves. For now, if this being toggleable means we see either everything or _no_ highlighting at all, that's totally fine. Just for people who know the airspace from the video map, removing the overlay would be a fairly critical thing since the map will show the boundaries and the overlay isn't needed. But if we want to see the altitudes as a reminder, they can be turned on for reference.

This may take a bit longer, but with these two additions, we are looking at probably the most significant recent change to the sim since the recent reworkings of the FMS. So thank you very much for your work on this one!

felixscheffer:
@Marcel510 Could you have a look at the airspaces at EDDF? They seem to overlap each other in a strange way.

Fechulo:
@felixscheffer @erikquinn What's the status on this? It would be incredible if we could get this out soon

erikquinn:
@Fechulo I guess it just got lost in the GH issue land... adding to top priority so i'll at least see it. Will be near the front of the list for sure!

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1515,129 +1517,76 @@ export default class CanvasController {
...
         }
 
-        cc.linewidth = 1;
-        cc.strokeStyle = this.theme.SCOPE.RANGE_RING_COLOR;
-
-        // Fill up airportModel's ctr_radius with rings of the specified radius
-        for (let i = 1; i * rangeRingRadius < airportModel.ctr_radius; i++) {
-            cc.beginPath();
-            cc.arc(0, 0, rangeRingRadius * CanvasStageModel.scale * i, 0, tau());
-            cc.stroke();
```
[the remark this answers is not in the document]
erikquinn: no longer observed

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1647,7 +1596,7 @@ export default class CanvasController {
      * @param poly {array<array<number, number>>}
      * @private
      */
-    _drawPoly(cc, poly) {
+    _drawPoly(cc, poly, fill = true) {
```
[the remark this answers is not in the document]
erikquinn: added docblock info

erikquinn:
Almost there... still need to allow the user to toggle the labels on/off.

j-651:
It looks like there's a floor of 3,500 set for EICK Cork so as soon as the airport loads, there's an error about leaving coverage without being on route. There's also no departure strips for I assume the same reason. Not sure if that would be tweaked on this branch or changed afterwards.

<details>
<summary>Screenshot:</summary>
<br>

</details>

Fechulo:
That bit of airspace is 'temporary'. Once this is merged, the actual airspace strata can be activated and this temporary one should be removed. I'd say there will be similar issues with many airports so it might be worth fixing those up in this branch, rather than breaking them and fixing them after.

erikquinn:
@Fechulo could you compile a list of those airports? Or possibly even work out some interim solutions so they don't break upon this being merged?

I think we will push this back to next month anyway since it would probably be unwise to force it out now and potentially break multiple airports.

Fechulo:
@erikquinn Yeah I'll do that and I'd be happy to activate the airspace shelves for all those myself. Would you prefer that to be its own PR or part of this one?

erikquinn:
yeah part of this is good. you can push directly here :)

Fechulo:
Ok so the issue described by @j-651 above might actually need to be dealt with differently. The problem is that there are multiple airports where the airspace doesn't start at 0ft (or whatever the airport elevation is), so aircraft on the ground at the airport will technically be outside the airspace, so they won't show up on the strips list and points are deducted to to departures leaving the airspace without clearance. The two solutions I see are:
1) Force the airspace floor to be 0 around the airport
2) Add an exception for departures, where they will still show up on the flight strips list despite not being in our airspace. Possibly have them grayed out to show we still can't 'control' them but they can be cleared as filed and taxied?

The third solution would be to finally remove all tower functionality from the sim but that seems a little more complicated. @erikquinn thoughts?

Fechulo:
Here's a spreadsheet with all the current active airports and their situation regarding this PR. Eventually all airports should be updated to have multiple airspace sections.

PR #1314, opened 2019-02-03
feature/1313 - Add scope command for changing PTL length

### LINKED ISSUE #1313: Add scope command for changing vector length
Currently, the vector lines / PTLs can only be adjusted through the settings menu. Add the ability to increase and decrease them with the F1/F2 keys.

erikquinn:
Resolves #1313.


src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -25,7 +26,17 @@ export default class ScopeModel {
...
+
+        /**
+         * Length of PTL lines (aka "vector lines") for all aircraft
+         *
+         * @for ScopeModel
+         * @property _ptlLength
+         * @type {number} length in minutes
+         * @private
+         */
+        this._ptlLength = null;
```
felixscheffer: should be 0, not null

erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -25,7 +26,18 @@ export default class ScopeModel {
          * @type {EventBus}
          * @private
          */
-        this._eventBus = EventBus;
+        this._eventBus = null;
```
n8rzz: pretty sure our existing pattern is to assign this in the `constructor`.  Some day I'd love to remove these local re-assignments of the `EventBus`.  I think I had a good reason for this once, but I've seen grown to really not like it.

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -25,7 +26,18 @@ export default class ScopeModel {
...
+        /**
+         * Length of PTL lines (aka "vector lines") for all aircraft
+         *
+         * @for ScopeModel
+         * @property _ptlLength
+         * @type {number} length in minutes
+         * @default 0
+         * @private
+         */
+        this._ptlLength = 0;
```
n8rzz: if this is a number and is expected to be updated before use, this should actually be `-1`.  assuming only positive numbers are valid for this property, assigning a negative number makes it obvious that this property is invalid

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -49,36 +61,55 @@ export default class ScopeModel {
...
      * @for ScopeModel
-     * @method _init
-     * @private
+     * @method init
      * @chainable
      */
-    _init() {
+    init() {
+        this._eventBus = EventBus;
+        this._theme = THEME.DEFAULT;
```
n8rzz: why move these two here?

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -205,6 +300,21 @@ export default class ScopeModel {
...
     }
 
+    /**
+     * Set the length of the PTL lines for all aircraft
+     *
+     * @for ScopeModel
+     * @method setPtlLength
+     * @param {number} length - length of PTL line, in minutes
+     */
+    setPtlLength(length) {
```
n8rzz: I'd almost prefer this param to be called `value` or `nextValue`.

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -126,6 +163,64 @@ export default class ScopeModel {
         return radarTargetModel.amendAltitude(altitude);
     }
 
+    /**
+     * Decrease the length of the PTL lines for all aircraft
+     *
+     * @for ScopeModel
+     * @method decreasePtlLength
+     */
+    decreasePtlLength() {
```
n8rzz: you could reduce the code here to something like:
```js
const validValues = Object.keys(PTL_LENGTHS);
const currentIndex = validValues.indexOf(this._ptlLength);
const nextIndexValue = validValues[currentIndex + 1];

if (nextIndexValue < 0 || nextIndexValue > validValues.length) {
    nextIndexValue = 0;
}

return PTL_LENGTHS[nextIndexValue]
```
then you only really need a single method, call it `changePtlLength()`.  The caller(s) would need to be updated to pass a direction (increasing/decreasing), but that could be as simple as a (+1/-1).  You'd only have to change one line in my example:
```js
changePtlLength(direction) {
    // ...
    const nextIndexValue = validValues[currentIndex + direction];
    // ...
}
```

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: Will change to `minutes`

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: Hm. Pretty clever, and more flexible. :+1:

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: @n8rzz So this is something we need to establish because we sometimes do it differently in different classes.

I'd like to greatly increase our utilization of the (awesome) ModelSourceFactory, which requires a full and proper reset and re-initialization of objects. The way I see it, we have two cases:

1. The class instance in question will be reused: We should set null values in the constructor and the `.reset()` methods. We should set default values in `.init()`. This way we can be sure a newly constructed object is the same as one to which we've done `.reset().init(data);`.

2. The class instance in question will not be reused: We should set initial values in the constructor, and any inferred values via `.init(data)` and it's submethods (such as `.initFms()` or whatever else. A `.reset()` method will not be present.

Is this a pattern you think makes sense?

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: See above

src/assets/scripts/client/scope/ScopeModel.js
n8rzz: Absolutely, and (I think) that's where my head was at when we first rolled that out.  I too would love to get more object pooling happening!

erikquinn: 👍 Added #1386 for that. Will make the changes here.

erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: It should always default to zero, and thus does not need to be set during instantiation.

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

erikquinn:
Work on this is complete, just need to add tests.

src/assets/scripts/client/AppController.js
```diff
@@ -305,7 +305,7 @@ export default class AppController {
     updateViewControls() {
         const { current: airport } = AirportController;
 
-        this._eventBus.trigger(EVENT.MARK_DIRTY_CANVAS);
+        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
```
n8rzz: this should probably be `MARK_DEEP_RENDER` since we are potentially updating items that require a deep render

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -108,7 +125,7 @@ export default class ScopeModel {
      * @param radarTargetModel {RadarTargetModel}
      * @return result {array} [success of operation, system's response]
      */
-    acceptHandoff(radarTargetModel) {
+    acceptHandoff(/* radarTargetModel */) {
```
n8rzz: could you add a comment (maybe even in the `@class` docblock) about why so many params are commented out?

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -127,6 +144,59 @@ export default class ScopeModel {
...
+    changePtlLength(direction) {
+        const validValues = GameController.getGameOption(GAME_OPTION_NAMES.PROJECTED_TRACK_LINE_LENGTHS)
+            .split('-')
+            .map((val) => parseFloat(val));
+        const currentIndex = validValues.indexOf(this._ptlLength);
+        const nextIndex = currentIndex + Math.sign(direction);
+
+        if (nextIndex < 0) {
+            this._ptlLength = 0;
+            this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
```
n8rzz: missing empty line above method call

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -127,6 +144,59 @@ export default class ScopeModel {
...
+
+            return;
+        }
+
+        if (nextIndex >= validValues.length) {
+            return;
+        }
+
+        this._ptlLength = validValues[nextIndex];
+        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
```
n8rzz: missing empty line above method call

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -127,6 +144,59 @@ export default class ScopeModel {
...
+        const nextIndex = currentIndex + Math.sign(direction);
+
+        if (nextIndex < 0) {
+            this._ptlLength = 0;
+            this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+
+            return;
+        }
+
+        if (nextIndex >= validValues.length) {
```
n8rzz: this should be the first check we do.

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -127,6 +144,59 @@ export default class ScopeModel {
...
+     * @for ScopeModel
+     * @method changePtlLength
+     * @param {number} direction - either -1 or 1 to indicate increment direction
+     */
+    changePtlLength(direction) {
+        const validValues = GameController.getGameOption(GAME_OPTION_NAMES.PROJECTED_TRACK_LINE_LENGTHS)
+            .split('-')
+            .map((val) => parseFloat(val));
+        const currentIndex = validValues.indexOf(this._ptlLength);
+        const nextIndex = currentIndex + Math.sign(direction);
```
n8rzz: this is clever, yet elegant.  nice 👍

test/scope/ScopeModel.spec.js
```diff
@@ -29,6 +31,35 @@ ava('creates RadarTargetCollection on instantiation', (t) => {
     t.true(model.radarTargetCollection instanceof RadarTargetCollection);
 });
 
+ava('#ptlLength returns #_ptlLength', (t) => {
+    const model = new ScopeModel();
+    const expectedResult = 17;
+    model._ptlLength = expectedResult;
+    const result = model.ptlLength;
+
+    t.true(result === expectedResult);
```
n8rzz: if you want to use `t.is()` for these instead, be my guest.  `.is()` will provide better error logging and also will check identity.  Not required, just calling it out 😉

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: Generally they're for lint since they are dummies that aren't actually used yet. I'll add a note for it!

src/assets/scripts/client/AppController.js
erikquinn: As I went to do this, I found there actually _is no_ MARK_DEEP_RENDER event. Each thing that necessitates a deep render is handled through its own event, which is registered in the CanvasController to call its handler, each of which calls the appropriate re-render method...

erikquinn: For example,
image

erikquinn: Not sure how to proceed; we can open up an issue for it to get it figured out separately perhaps.

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: done

test/scope/ScopeModel.spec.js
erikquinn: I like consistency, so I'll add an issue to find-and-replace the whole repo, and see if anything bad happens rather than just doing a few here.

erikquinn: Created #1393.

src/assets/scripts/client/AppController.js
erikquinn: Also worth noting (though you probably picked up on it), that all that happened was a rename of the existing event, so we can at least count on the behavior being the same as it was before the change, so there should be no new bugs from this. Doesn't mean it isn't wrong, but it does mean we won't break something new. :relieved:

erikquinn:
:man_facepalming: 

I forgot to push my changes before pressing the button, these changes are now being pushed onto the PR #1374 which is going in now.

src/assets/scripts/client/AppController.js
n8rzz: Biggest issue I had here was the name, it through me off a little. As long as it works I’m ok with it, we can circle back later.

PR #1315, opened 2019-02-04
bugfix/1312 - Fix KRDU arrivals not descending in time

### LINKED ISSUE #1312: RDU arrivals
Browser: Chrome

Airport: RDU

Approximate play time before issue: 0

#### Issue Description

Arrivals into RDU are often descending and calling up far too late, sometimes over or beyond the airport.

#### Steps to reproduce
-

Marcel510:
Resolves #1312 
- seperate spawnpatterns for GA aircraft with lower descent rates
- lower spawn altitudes
- improve a couple of GA values


Marcel510:
This doesnt solve the problem 100% as the vnav logic only looks at the next waypoint (I think?):
screenshot_210
The inbounds descend to ~15000 by `ARGAL ` and cant reach 6000 by `PAINS` eventhough they could descend down to 10000 by `ARGAL` already.

Only solution for this problem from my side would be to lower the spawning altitude to the airspace ceiling. 
What do you think? @Fechulo @indianbhaji

erikquinn:
Maybe. I do know that the current altitude is better (real world, they get 10 N HPW @ FL220). I may even just go with a straight FL200 for the spawn altitude. I know the FMS's VNAV is not very smart, it's just a lot harder than it seems to really get it right. But I know we need to... 😞

Marcel510:
FL200 wont change anything as the aircraft already maintain their spawning altitude for a couple of nm before commencing descent

erikquinn:
Yeah, I know, I only meant to confirm that I by chance happen to actually know the restriction that's given irl, since it's right next to my airspace. It's definitely an issue with the VNAV descent logic, and it's one we've known about for a very long time. It's just gruesome to try to get it to behave correctly in all situations.

assets/aircraft/hdjt.json
```diff
@@ -5,16 +5,16 @@
         "number": 2,
         "type": "J"
     },
-    "weightClass": "S",
+    "weightClass": "L",
```
erikquinn: Honda Jet is a small. Reverting this.

erikquinn: done

Marcel510: screenshot_211
never heard of the weightclass small so I looked it up and it said light that why I changed it 🤔

erikquinn: Bizarre, I would have thought ICAO would at least go with M... the FAA has it as a small:

image

It weighs 7,200 pounds, and small is <12,500. So the small is correct I guess, though it's real strange that your screenshot shows it as a large. 🤷‍♂️

PR #1317, opened 2019-02-05
enhancement/804 - add a changelog in-game

### LINKED ISSUE #804: In-game changelog
Related: #1105.

After each update, it might be helpful to have a changelog in-game so end-users can see new changes to the simulator.

How this might be implemented:
- A "changelog" box (perhaps similar in size to the current settings menu)
- A localStorage key (ATC_LAST_VERSION?) to check if the simulator has been played on the current version
- If ATC_LAST_VERSION is not the current version or does not exist, show the Changelog

sam-irl:
Resolves #804 

The purpose of this pull request is to add a changelog that can be viewed in-game.

- Shows only the latest version's changes
- Displays on startup if the user has not played this version before

Todo:

- [x] get different icon
- [ ] redo css, similar to how it is styled in the airport guide

screen shot 2019-02-07 at 2 24 42 pm


src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,223 @@
...
+import { STORAGE_KEY } from '../constants/storageKeys';
+
+/**
+ * A div template for the changelog for the container.
+ * Compare to the container elements for option selection and airport switching.
+ *
+ * @property CHANGELOG_CONTAINER
+ * @final
+ */
+const CHANGELOG_CONTAINER = '<div class="changelog-container nice-scrollbar"></div>';
```
n8rzz: Any reason these templates can’t just live in a handlebars template (.hbs)?

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Determines whether the user has played this version before,
+     * and if the changelog should display on load.
+     *
+     * @for ChangelogController
+     * @method _shouldToggleOnStart
+     * @returns {Boolean} if the user has not played this version
+     */
+    _shouldToggleOnStart() {
```
n8rzz: Let’s change this to ‘_shouldShowOnLoad()`

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,223 @@
...
+    loadChangelogContent() {
+        const options = {
+            url: 'assets/changelog.json',
+            immediate: true
+        };
+        const changelogPromise = this.contentQueue.add(options);
+
+        changelogPromise.done((data, textStatus, jqXHR) => {
+            this.content = data.changelog;
+            this.load_complete();
```
n8rzz: No underscores in method names.  Also, does this need to be a blind handler?

sam-irl: I believe it does need to be bound

sam-irl:
@erikquinn I used the contentqueue here because passing everything through the `AppController` constructor seemed excessive. I guess it doesn't really matter though, if there's any concerns with the contentqueue we can just put them in an object and pass *that* to the appcontroller, rather than having ten billion parameters

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+         */
+        this.contentQueue = contentQueue;
+
+        /**
+         * The DOM of the changelog container, containing version, data, and dismiss.
+         *
+         * @property $changelogContainer
+         * @type {JQuery|HTMLElement}
+         */
+        this.$changelogContainer = $(SELECTORS.DOM_SELECTORS.CHANGELOG_CONTAINER);
```
n8rzz: these should all be defaulted to `null` and then set within a `._createChildren()` method that's called before `._setupHandlers()`

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * @for ChangelogController
+     * @method _setupHandlers
+     * @private
+     * @chainable
+     */
+    _setupHandlers() {
+        this._onChangelogToggleHandler = this._onChangelogToggle.bind(this);
+
+        this.$changelogToggle.on('click', this._onChangelogToggleHandler);
```
n8rzz: these should be moved to the `enable` method

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+     * Sets the changelog data
+     *
+     * @for ChangelogController
+     * @method enable
+     */
+    enable() {
+        this.content = '<p>Loading...</p>';
+        this.version = window.GLOBAL.VERSION;
+
+        this.$changelogVersion.text(`openScope v${this.version}`);
```
n8rzz: you're doing too much work here.  I think if the basic changelog template is in an `.hbs` file already, we can leverage the build to do this work for us (see: https://github.com/openscope/openscope/blob/develop/src/templates/layout.hbs#L4)

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+            url: 'assets/changelog.json',
+            immediate: true
+        };
+        const changelogPromise = this.contentQueue.add(options);
+
+        changelogPromise.done((data, textStatus, jqXHR) => {
+            let loadedContent = data.changelog;
+
+            // Strip out the version, as it's already in the title
+            loadedContent = loadedContent.split('</h1>')[1];
```
n8rzz: why?  why not just use what we get?  or, at worst, hide this via css?

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+
+            // We're going to manually set the `target` attribute to
+            // `_blank`, so that the links open in a new tab.
+            const searchKey = 'a href=';
+            const replaceWith = 'a target="_blank" href=';
```
n8rzz: this is dangerous.  I'd strongly caution you against doing it this way.

src/assets/style/module/changelog/_changelog.less
```diff
@@ -0,0 +1,97 @@
+.changelog-container {
+    width: 480px;
```
n8rzz: please alphabetize these rules

src/assets/style/module/changelog/_changelog.less
```diff
@@ -0,0 +1,97 @@
...
+    opacity: 0;
+    border-radius: 1px;
+    box-sizing: border-box;
+    box-shadow: 0 0 0 6px @blue-medium;
+    pointer-events: none;
+    transition: opacity 0.5s ease;
+    z-index: 200;
+}
+
+.changelog-container.changelog-container-is-visible {
```
n8rzz: just stick with the single class declarations here and below.  `.changelog-container` can be removed from all of the following items

n8rzz: `changelog-container-is-visible` -> `changelog-container_isOpen`

src/assets/style/module/changelog/_changelog.less
```diff
@@ -0,0 +1,97 @@
...
+    pointer-events: auto;
+    opacity: 1;
+}
+
+.changelog-container .changelog-title {
```
n8rzz: `changelog-title` -> `changelog-hd`

src/assets/style/module/changelog/_changelog.less
```diff
@@ -0,0 +1,97 @@
...
+
+.changelog li + li {
+    padding: 5px 0;
+}
+
+.changelog a {
+    color: @light-gray;
+}
+
+.changelog table,
```
n8rzz: do we have tables in the changelog?  or is this just copied from the airportGuide stuff?

tools/tasks/mediaTasks.js
```diff
@@ -5,6 +5,7 @@ module.exports = function(gulp, config) {
     const path = require('path');
     const jsonmin = require('gulp-jsonmin');
     const merge = require('merge-stream');
+    const changelog = require('./copyChangelog');
```
n8rzz: keep the naming consistent here.  if the file is called `copyChangelog` then the function should also be called `copyChangelog`.  Especially since that is the name of the exported function from that file

tools/tasks/mediaTasks.js
```diff
@@ -24,12 +25,18 @@ module.exports = function(gulp, config) {
         return merge(fonts, images);
     };
 
+    const copyChangelog = () => {
```
n8rzz: this can be removed, it's not needed.  check out how I did this on #1198

tools/tasks/copyChangelog.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/**
+ * Gets the first `h1` HTML element in the stringified HTML.
+ *
+ * @method getHtmlElement
+ * @param {String} totalHtml
+ * @returns {String} the HTML section
+ */
+function getHtmlElement(totalHtml) {
```
n8rzz: why is this function important?  what is this actually doing and why do we need it?

src/templates/_changelog.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="changelog-container nice-scrollbar">
+    <h1 class="changelog-title">Changelog</h1>
+    <h2 class="changelog-version"></h2>
```
n8rzz:
```suggestion
    <h2 class="changelog-version">{{version}}</h2>
```

src/templates/_changelog.hbs
```diff
@@ -0,0 +1,6 @@
+<div class="changelog-container nice-scrollbar">
```
n8rzz: I have concerns about this markup, I think we could structure this a little better.  I'd like to see what this actually looks like first, though, before I suggest ripping this all apart.

src/assets/scripts/client/changelog/ChangelogController.js
```diff
@@ -0,0 +1,200 @@
...
+        this.$changelogData = $(SELECTORS.DOM_SELECTORS.CHANGELOG_CONTENT);
+
+        /**
+         * The link at the bottom of the container to
+         * dismiss the changelog.
+         *
+         * @property $changelogDismiss
+         * @type {JQuery|HTMLElement}
+         */
+        this.$changelogDismiss = $(SELECTORS.DOM_SELECTORS.CHANGELOG_DISMISS);
```
felixscheffer: I am not a big fan of the dismiss button. None of our dialogs has a close / dismiss button and I would prefer to keep it consistent. Maybe we should add a "x" button to the top right corner of the dialogs (in another pull request).

src/assets/style/module/changelog/_changelog.less
felixscheffer: IMHO we should use the same styling for the airport guides and the changelog.

n8rzz: agreed.  maybe move that to a different css module and call it `blackBox` or something?

src/assets/style/module/changelog/_changelog.less
felixscheffer: I started to standardize the dialog styling in https://github.com/openscope/openscope/pull/1192/files#diff-8b0e11a201d389745a7546e04a8dfedc

It would be awesome if you could add it to this PR so we don't have to refactor it later.

felixscheffer: For the record, @n8rzz  created a ticket for modal styling #1323

src/assets/scripts/client/changelog/ChangelogController.js
erikquinn: I agree it doesn't match the convention and would make sense to remove. I also agree it makes sense to at some point show this at initial load (put last loaded version in localstorage, and open changelog view when loaded version is newer than the one in localstorage).

So let's do this:
- create an issue for showing the changelog view when loading a version newer than the version noted in localstorage. Here we can determine how best to dismiss the window.
- Change the "dismiss" in this PR to "View Full Changelog", and link it to the changelog document, since people may miss more than 1 version, and wonder what changed since they last played

sam-irl: @erikquinn The logic's already there for displaying the changelog if the version is new, should this be removed?

sam-irl: Follow-up, should the link to the full changelog be pointed at the `master` or `develop` branch?

n8rzz: neither.  I'd point it to the releases page: (https://github.com/openscope/openscope/releases)

sam-irl: sounds good

erikquinn: If it shows itself when a new version is loaded, then great!

sam-irl:
@n8rzz As the built HTML is meant to be blackboxed, I think it might be beneficial if we only write the most recent changelog entry to the JSON file.

I came up with the following regexp for matching a changelog entry header, please tell me if this is an ok implementation: `/# [0-9]\.[0-9]+\.[0-9] \(.*\)/g`

https://regexr.com/4815b

src/templates/_changelog.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="changelog-container js-changelogContainer nice-scrollbar">
+    <h1 class="changelog-hd">Changelog</h1>
+    <h2 class="changelog-version">v{{version}}</h2>
+    <p class="js-changelog blackBox"></p>
+    <div class="changelog-view-full">
+        <a href="https://github.com/openscope/openscope/releases">View full changelog</a>
```
felixscheffer: IMHO the link should open in a new window/tab.

tools/tasks/copyChangelog.js
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Strips out everything but the most recent changelog entry.
+ *
+ * @method getLatestEntry
+ * @param {String} markdown
+ * @returns {String} the data from the latest entry
+ */
+function getLatestEntry(markdown) {
+    const headerRegexp = /# [0-9]\.[0-9]+\.[0-9] \(.*\)/g;
```
felixscheffer: we should consider moving this to `globalconstants.js` `REGEX`. maybe...I am not sure.

src/assets/style/module/blackBox/_blackBox.less
```diff
@@ -0,0 +1,40 @@
+/*
+    These are designed for use with "black-boxed" content - that is,
+    we generate the HTML from something else (i.e. markdown), so we
+    shouldn't modify it.
+
+    Because the HTML is generated from markdown, we can't assign classes
+    to it directly, so we have to use element selectors.
+*/
+.blackBox h1,
```
felixscheffer: IMHO the term `.blackBox` is too ambiguous. I would prefer something like `.showdown` to make it more obvious where the markup is coming from and to distinguish it from other generated content.

erikquinn: I agree. Already pressed the button :grimacing: we can get it some other time I guess.

tools/tasks/copyChangelog.js
erikquinn: definitely

src/templates/_changelog.hbs
erikquinn: We are pushing that back to a new issue for some reason: #1322.

PR #1325, opened 2019-02-07
refactor/1323 - create modal css module

### LINKED ISSUE #1323: create modal css module
we use modals and each one styles things individually.  These style rules should be consolidated into a `modal` module.  This module should be concerned only with positioning and visibility (open/closed), everything else would be the responsibility of the child modules.

felixscheffer:
Resolves #1323 

Copied over from #1192


sam-irl:
@felixscheffer @n8rzz I've added the `blackBox` styling over in #804

src/assets/style/module/dialog/_dialog.less
```diff
@@ -0,0 +1,161 @@
...
+.dialog td {
+    padding: 3px;
+}
+
+.dialog.open {
+    opacity: 1;
+    pointer-events: auto;
+}
+
+.dialog-title {
```
sam-irl:
```suggestion
.dialog-bd {
```
I think this is the convention now?

src/assets/style/module/dialog/_dialog.less
```diff
@@ -0,0 +1,161 @@
...
+    padding: @spacing-2x 0 7px 0;
+    font-size: 170%;
+    text-align: center;
+}
+
+.dialog-body {
```
sam-irl:
```suggestion
.dialog-bd {
```
As above

Issue #995, opened 2018-06-04
Simulator continues running after the simulator pausing and loses focus


indianbhaji:
Browser: Chrome 66.0.3359.181 (Official Build) (64-bit)

Airport: all

Approximate play time before issue: n/a

#### Issue Description
If you pause the simulator using the pause button on the bottom, the screen is greyed out with a play button appearing in the middle. However, if the tab loses focus, you return to see that the simulator is running, instead of the grey screen and play button as before.

#### Steps to reproduce
- pause the game
- alt-tab out or go to another tab
- return to the simulator tab
- the simulator is running

n8rzz:
Hmm, I feel like we did this on purpose because the pause logic is a little weird.

erikquinn:
Resolved by #1336. Closing.

PR #1336, opened 2019-03-05
bugfix/1335 - fix inconsistent state of control buttons due to window onblur / onfocus

### LINKED ISSUE #1148: Time warp toggle should reset when window regains focus
Also see #1335.

#### Issue Description
If the window loses focus while time warp is enabled, the simulation is paused and time warp is set back to 1 (as desired). When the window regains focus, the time warp toggle displays time warp still enabled.

Reported by darren on Slack

#### Steps to reproduce
- Set time warp to 5 (button will be red)
- Switch to another window
- Switch back to openScope
- Time warp button will remain red, despite time warp being disabled

### LINKED ISSUE #1335: UI inconsistencies / UX issue due to window onblur / onfocus
Also see #1148.

Browser: chrome (any)

Airport: any

#### Issue Description
Control button UI elements become inconsistent due to window onblur / onfocus.

**Timewarp**: if timewarp (x2 / x5) is active when app loses focus, `_onWindowBlur` [resets timewarp to x1](https://github.com/openscope/openscope/blob/3763f6e966350690699bea6f00a21147c0e3490b/src/assets/scripts/client/game/GameController.js#L569-L571) without updating the button visuals to match.

**Pause**: even if app was already (manually) paused when loses focus, `_onWindowFocus` [unpauses it](https://github.com/openscope/openscope/blob/3763f6e966350690699bea6f00a21147c0e3490b/src/assets/scripts/client/game/GameController.js#L584) without updating the button visuals to match.  

Furthermore this behavior is not ideal; if the simulation was manually paused, it would be preferable that the app respect the paused state and not unpause automatically.

#### Steps to reproduce

**Timewarp**
* turn on timewarp by clicking button, note button color change indicating timewarp active
* tab out and then back in
* timewarp is now inactive, but button has not been updated

**Pause**
* pause by clicking button, note button color change indicating paused state
* tab out and then back in
* unpaused but button has not been updated

cake-pie:
Resolves #1335

The purpose of this pull request is to fix inconsistent state of control buttons when window loses and regains focus.

Also fixes #1148, #995 (duplicates)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 
 ### Bugfixes
+- [#1335](https://github.com/openscope/openscope/issues/1329) - Fix inconsistent state of control buttons when window loses and regains focus.
 
 
 ### Enhancements & Refactors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 
 ### Bugfixes
-- [#1335](https://github.com/openscope/openscope/issues/1329) - Fix inconsistent state of control buttons when window loses and regains focus.
+- [#1335](https://github.com/openscope/openscope/issues/1335) - Fix inconsistent state of control buttons when window loses and regains focus.
 
 
 ### Enhancements & Refactors
--- a/src/assets/scripts/client/game/GameController.js
+++ b/src/assets/scripts/client/game/GameController.js
@@ -566,9 +566,16 @@ class GameController {
      */
     _onWindowBlur(event) {
         this.game.focused = false;
+
         // resetting back to 1 here so when focus returns, we can reliably reset
         // `#game.delta` to 0 to prevent jumpiness
         TimeKeeper.updateSimulationRate(1);
+        // update visual state of the timewarp control button for consistency
+        const $fastForwards = $(SELECTORS.DOM_SELECTORS.FAST_FORWARDS);
+        $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_2);
+        $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_5);
+        $fastForwards.prop('title', 'Set time warp to 2');        
+
         TimeKeeper.setPause(true);
     }
 
--- a/src/assets/scripts/client/game/GameController.js
+++ b/src/assets/scripts/client/game/GameController.js
@@ -576,7 +576,7 @@ class GameController {
 
         $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_2);
         $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_5);
-        $fastForwards.prop('title', 'Set time warp to 2');        
+        $fastForwards.prop('title', 'Set time warp to 2');
         TimeKeeper.setPause(true);
     }
 
--- a/src/assets/scripts/client/game/GameController.js
+++ b/src/assets/scripts/client/game/GameController.js
@@ -588,7 +588,10 @@ class GameController {
     _onWindowFocus(event) {
         this.game.focused = true;
 
-        TimeKeeper.setPause(false);
+        // if was already manually paused when lost focus, respect that
+        if (!$('html').hasClass(SELECTORS.CLASSNAMES.PAUSED)) {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 
 ### Bugfixes
+- [#1335](https://github.com/openscope/openscope/issues/1329) - Fix inconsistent state of control buttons when window loses and regains focus.
```
felixscheffer: please add a new line after that so git will understand that you add a line instead of replacing one.

src/assets/scripts/client/game/GameController.js
```diff
@@ -566,9 +566,16 @@ class GameController {
      */
     _onWindowBlur(event) {
         this.game.focused = false;
+
         // resetting back to 1 here so when focus returns, we can reliably reset
         // `#game.delta` to 0 to prevent jumpiness
         TimeKeeper.updateSimulationRate(1);
+        // update visual state of the timewarp control button for consistency
+        const $fastForwards = $(SELECTORS.DOM_SELECTORS.FAST_FORWARDS);
+        $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_2);
```
felixscheffer: please add an empty line between function calls and assignments.

src/assets/scripts/client/game/GameController.js
```diff
@@ -566,9 +566,16 @@ class GameController {
...
+        $fastForwards.removeClass(SELECTORS.CLASSNAMES.SPEED_5);
+        $fastForwards.prop('title', 'Set time warp to 2');        
+
```
felixscheffer: remove empty line

src/assets/scripts/client/game/GameController.js
```diff
@@ -581,7 +588,10 @@ class GameController {
     _onWindowFocus(event) {
         this.game.focused = true;
 
-        TimeKeeper.setPause(false);
+        // if was already manually paused when lost focus, respect that
+        if (!$('html').hasClass(SELECTORS.CLASSNAMES.PAUSED)) {
```
n8rzz: ~Is this actually needed?  All this method is trying to do is change the `focused` and `pause` values.  Pretty sure any UI feedback would be covered by your updates in `._onWindowBlur()`.  I'm also a little wary of calling `TimeKeeper` based on the state of an html element.~

I see what you're trying to do here.  We (well I am) are a big fan of exclusive `if` statements.  Meaning, when we don't have the data we need to continue, `return`.  So with that in mind, I would reverse the logic here to something like: 
```js
if ($('html').hasClass(SELECTORS.CLASSNAMES.PAUSED)) {
    return;
}
```

Last thing here, avoid using `$('html')` to select DOM elements.  Doing so means we have to start at the very top of the DOM tree to find what we're looking for.  Instead, use one of the cached elements.  Not sure if we have one in this class for the pause controls, but if we do we should use it.

cake-pie: >  avoid using `$('html')` ...  Instead, use one of the cached elements.

Inline vars and `$()` appear to be used throughout this class, cached elements have been commented out where they were implemented. Something about tests breaking?
https://github.com/openscope/openscope/blob/3763f6e966350690699bea6f00a21147c0e3490b/src/assets/scripts/client/game/GameController.js#L78-L85

CHANGELOG.md
cake-pie: Hm? I did add rather than replace a line. Compare [before](https://raw.githubusercontent.com/openscope/openscope/3763f6e966350690699bea6f00a21147c0e3490b/CHANGELOG.md) vs [after](https://raw.githubusercontent.com/openscope/openscope/6da1f0ad6a2d3168a90cd3a206930296fb6917ec/CHANGELOG.md).  
Or is there something else I'm missing? (Besides the incorrect link -- that I will fix)

felixscheffer: My mistake. I got confused by the red minus...

src/assets/scripts/client/game/GameController.js
felixscheffer: Ideally, we would move all the button-related code to `UiController` which "owns" the buttons.

erikquinn: done

PR #1348, opened 2019-03-25
bugfix/1183 - marks a deep render on theme change

### LINKED ISSUE #1183: Changing theme should trigger a deepRender() within CanvasController
found in #1182, though likely not related to changes there.

moving from default -> classic
screen shot 2018-11-10 at 12 38 12 pm
 
moving from classic -> default
screen shot 2018-11-10 at 12 38 25 pm

At a high level, when this game option changes an event should be triggered that results in `CanvasController.deepRender()` being called immediately afterward.  

The app renders the terrain only as often as it needs to.  Which means if there is no pan or zoom change the terrain will not re-render, which is why we see this issue here.

sam-irl:
Resolves #1183 

The purpose of this pull request is to call `CanvasController._markDeepRender` in the `setTheme` method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # 6.12.0 (April 1, 2019)
 ### New Features
+- [#804](https://github.com/openscope/openscope/issues/804) - Adds a changelog in-game
 
 
 ### Bugfixes
 - [#1335](https://github.com/openscope/openscope/issues/1335) - Fix inconsistent state of control buttons when window loses and regains focus.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,12 @@
 # 6.12.0 (April 1, 2019)
 ### New Features
+- [#804](https://github.com/openscope/openscope/issues/804) - Adds a changelog in-game
```
felixscheffer: 👍

PR #1350, opened 2019-03-25
enhancement/1349 - moves cid to aircraftmodel and allows selection by it

### LINKED ISSUE #1349: CID numbers should be generated and kept by the AircraftModel
Currently, all CID numbers are purely cosmetic and generated by the StripView. In anticipation for allowing scope commands to reference aircraft by their CID, we should generate the CID in the AircraftModel, store it there, and allow aircraft to be referenced by their CID.

sam-irl:
Resolves #1349 

The purpose of this pull request is to move CID numbers from the StripView and having them being fully cosmetic to ownership by the AircraftModel

The logic assigning and removing now belongs to the `AircraftController`


src/assets/scripts/client/aircraft/StripView/StripViewModel.js
```diff
@@ -506,7 +492,7 @@ export default class StripViewModel extends BaseModel {
     _redraw() {
         this.$callsignView.text(this._callsign);
         this.$aircraftTypeView.text(this._aircraftType);
-        this.$cidView.text(this.cidString);
+        this.$cidView.text(this._cid);
```
felixscheffer: should be cidString, not cid

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -894,4 +944,46 @@ export default class AircraftController {
...
+     * Generate a unique number to represent a `CID`
+     *
+     * Should be displayed with leading zeros, so a `CID` value of `1` should be displayed as `001`
+     *
+     * @for AircraftController
+     * @method _generateCidNumber
+     * @return nextCid {number}
+     * @private
+     */
+    _generateCidNumber() {
```
felixscheffer: I think this is a good opportunity to add some tests for this method.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -894,4 +944,46 @@ export default class AircraftController {
...
+     * Remove a given `#cid` from use
+     *
+     * Used when an aircraft has landed or departed controlled airspace
+     *
+     * @for AircraftController
+     * @method _removeCidFromUse
+     * @param cid {number}
+     * @private
+     */
+    _removeCidFromUse(cid) {
```
felixscheffer: I think this is a good opportunity to add some tests for this method.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -414,6 +432,36 @@ export default class AircraftController {
...
     }
 
+    /**
+     * Finds an aircraft by its cid (NOT internal id)
+     *
+     * @method findAircraftByCid
+     * @param {string} cid
+     * @return {AircraftModel}
+     */
+    findAircraftByCid(cid) {
```
felixscheffer: Needs tests

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -414,6 +432,36 @@ export default class AircraftController {
...
+    }
+
+    /**
+     * Finds an aircraft by its transponder (squawk) code
+     *
+     * @method findAircraftByTransponderCode
+     * @param {Number} transponderCode
+     * @return {AircraftModel}
+     */
+    findAircraftByTransponderCode(transponderCode) {
```
felixscheffer: Needs tests

src/assets/scripts/client/scope/RadarTargetCollection.js
```diff
@@ -154,7 +154,8 @@ export default class RadarTargetCollection extends BaseCollection {
         const radarTargetModels = this._items;
         const results = _filter(radarTargetModels, ({ aircraftModel }) =>
             aircraftModel.transponderCode === aircraftReference ||
-            aircraftModel.callsign === aircraftReference
+            aircraftModel.callsign === aircraftReference ||
+            aircraftModel.cid === aircraftReference
```
felixscheffer: Please remove the todo comment above `findRadarTargetModelForAircraftReference`

src/assets/scripts/client/aircraft/StripView/StripViewModel.js
erikquinn: Is now natively a string, so _cid is good again.

src/assets/scripts/client/scope/RadarTargetCollection.js
erikquinn: good catch. done

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -76,6 +76,7 @@ import {
     INVALID_NUMBER,
     TIME
 } from '../constants/globalConstants';
+import { leftPad } from '../utilities/generalUtilities';
```
cake-pie: lint-diff complains:
```
/home/travis/build/openscope/openscope/src/assets/scripts/client/aircraft/AircraftModel.js
  79:10  error  'leftPad' is defined but never used  no-unused-vars
```
Added in 82b0803 but no longer used after e8c15b7, can be removed.

oobayly:
Seeing as this has been orphaned, and now that #1402 has been merged, I had a look at reworking this - you can see the changes in f6b9d47a

I took a different approach as:
* Erik didn't want the cid property added to the `AircraftModel`
* Populating a list of the 1000 possible CIDs didn't seem the best idea as alpha-numeric CIDs are mentioned as being possible

Instead, I opted for using the `EventBus` to signal to the `RadRadarTargetCollection` that a StripView has been added or removed. This maintains a soft link of the CID -> CallSign, which `findRadarTargetModelForAircraftReference` can use. This CID has also been explicitly typed as a string.

Any comments?

erikquinn:
@oobayly I don't have time quite yet to explain fully, but that isn't exactly the right approach either. There is an _extremely_ specific way these things are all interrelated to each other, and to prevent issues down the line, we'll have to match those relationships correctly. It's something that will take quite a bit of thought on my end before we can really do anything with it. Truth is that this small-seeming change is setting the foundation for the fundamental way flightplans are stored and the radar works. It's on my horizon though. I'll self-assign as a reminder to try to lay out a full gameplan to that effect, but for now I'd skip over this one since we don't want to get locked into the wrong track right at the starting line.

oobayly:
No worries.

PR #1357, opened 2019-04-03
documentation/1356 - correct departure command sequence

### LINKED ISSUE #1356: Aircraft command reference departure sequence
The departure sequence in [aircraft command reference](https://github.com/openscope/openscope/blob/develop/documentation/aircraft-commands.md) page is presently CAF-CVS-TO-TAXI. The taxiing of course must happen first.

guyndumais:
Resolves #1356 

The purpose of this pull request is to correct the Departure commands sequence.

Taxiing was as the end of the sequence as it should be the first command in the list.

### DIFF
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -3,10 +3,10 @@
 ## Table of Contents
 
 [Departure Commands](#departure-commands)
+- [Taxi](#taxi)
 - [Cleared as Filed](#cleared-as-filed)
 - [Climb via SID](#climb-via-sid)
 - [Takeoff](#takeoff)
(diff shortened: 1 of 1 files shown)


sam-irl:
I'm not sure what the conversation has been regarding this pull request, but when I originally wrote the table of contents I believe it was all alphabetical? I'm not sure if consensus has shifted away from that, however

erikquinn:
Nope, still doing alphabetical, so it's correct as currently showing on develop. Closing.

felixscheffer:
@sam-irl can you double-check that it's still in alphabetical order? I think it wasn't the last time I checked but I might be wrong.

PR #1367, opened 2019-04-23
hotfix/1366 - Remove/combine conflicting spawn patterns

### LINKED ISSUE #1366: ABQ generates initial traffic that collides
Aircraft are generated at the same position and collide at the following airports:
- [x] EGGW
- [x] EGNM
- [x] ENGM
- [x] KABQ
- [x] KATL
- [x] KAUS
- [x] KBOS
- [x] KCVG
- [x] KMCI
- [x] KPIT
- [x] KRDU
- [x] KSTL
- [x] LROP
- [x] LTBA
- [x] OMDB

erikquinn:
Resolves #1366

### Test me here: https://openscope-dev-pr-1367.herokuapp.com/


erikquinn:
@fechulo they actually stay vertically separated from each other well within the airspace. It looks goofy at the start because there are pairs of airplanes stacked at each entry, but the KAP (cape air C208s) guys won't be able to keep up with the Jets, so they can't go into the other pattern. We should probably just do without them, so I'll make that change.

PR #1373, opened 2019-05-03
feature/1344

### LINKED ISSUE #1344: Add analytics tracking of traffic rate adjustments
The traffic rate adjustment dialog added by #1192 (feature #16) adds some functionality that we'd like to see how people use in order to judge any needed traffic adjustments to certain airports, etc, and/or just generally learn how people use the sim so we can cater to that better.

n8rzz:
Resolves #1344 

adds EventTracking to the trafficRateController and concerns

### DIFF
--- a/src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
+++ b/src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
@@ -405,6 +405,25 @@ export default class SpawnPatternModel extends BaseModel {
     }
 
     /**
+     * Convenience getter used for `EventTracker`
+     *
+     * This getter *should not* be used in code for
+     * anything other than event tracking
+     *
+     * @property airportIcao
+     * @return string
+     */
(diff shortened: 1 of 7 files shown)


src/assets/scripts/client/trafficGenerator/SpawnPatternModel.js
```diff
@@ -404,6 +404,25 @@ export default class SpawnPatternModel extends BaseModel {
...
+     * @property airportIcao
+     * @return string
+     */
+    get airportIcao() {
+        if (this.isOverflight()) {
+            return 'overflight';
+        }
+
+        return this.isArrival() ?
+            this.destination :
```
erikquinn: lint-diff shows that the `:` and `?` should be at the beginning of the following line, not the end of the leading one.

erikquinn: done

PR #1382, opened 2019-05-10
refactor/1381 - Use Jeppesen chart's english names for all airports

### LINKED ISSUE #1381: Change all airport names in loadlist to the official English name

Marcel510:
Resolves #1381

### DIFF
--- a/assets/airports/airportLoadList.js
+++ b/assets/airports/airportLoadList.js
@@ -43,7 +43,7 @@ window.AIRPORT_LOAD_LIST = (function() { // eslint-disable-line wrap-iife
      *    +----------+------------------------+
      *    |   Hard   |    More than 60 AcpH   |
      *    +----------+------------------------+
-     * - `AIRPORT_NAME` is the official English name of the airport (ex: San Francisco International Airport)
+     * - `AIRPORT_NAME` is the official English name of the airport (ex: San Francisco International Airport) according to Jeppesen charts
      * - `MEETS_PREMIUM_STANDARDS` is either `true` or `false`, see the airport standards document
      *
      * @property AIRPORT_LOAD_LIST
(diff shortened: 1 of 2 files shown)


assets/airports/airportLoadList.js
```diff
@@ -43,7 +43,7 @@ window.AIRPORT_LOAD_LIST = (function() { // eslint-disable-line wrap-iife
      *    +----------+------------------------+
      *    |   Hard   |    More than 60 AcpH   |
      *    +----------+------------------------+
-     * - `AIRPORT_NAME` is the name of the airport (ex: San Francisco International Airport)
+     * - `AIRPORT_NAME` is the official English name of the airport (ex: San Francisco International Airport)
```
Fechulo: Add 'according to Jeppesen charts'

Marcel510: Screenshot_255
Well the thing with the Jeppesen charts is that it doesnt use the airports name its either the icao code (EDDM) or the city where the airport is located (MUNICH, GERMANY).

Fechulo: Ah, I see. What source did you use for the names?

Marcel510: Mainly the official website of the airport

Fechulo: Hmm, I'm not sure whether that will work for every airport in the world

Fechulo: It'd be better if we had a single source

Marcel510: the problem is there isnt really a perfect source (or I haven't found one yet) that has all the official airport names. Maybe it would make more sense to leave it like that and decide for every airport separately in the review process. What do you think?

Marcel510:
@Fechulo What should we do with this one?

assets/airports/airportLoadList.js
erikquinn: Looks like you can still use this source. Just look at the airport diagram and use the name listed under the city. For EDDM, it just says "Munich", for LFPG "Charles-de-Gaulle", EDDF "Frankfurt/Main", KMIA "Miami Int'l". Should do the trick.

PR #1073, opened 2018-09-02
refactor/1072 - Overhaul of KELP

JPmAn24:
Resolves #1072 

The purpose of this pull request is to Update and Re-introduce KELP into the game.

PR #1383, opened 2019-05-11
bugfix/1351 - Fix turn right 5 degrees not useable

### LINKED ISSUE #1351: "Turn 5 degrees right" is no longer possible
It seems that the parser for turns must be expecting incremental turns to be two digits-- this should not be the case, as <10 degree turns are quite useful as well, and need to be supported.

JPmAn24:
Resolves #1351 

The purpose of this pull request is to re-enable the `t r` or `t l` commands to use 1-digit increments

### DIFF
--- a/test/aircraft/Pilot/Pilot.spec.js
+++ b/test/aircraft/Pilot/Pilot.spec.js
@@ -1048,7 +1048,7 @@ ava('.maintainHeading() returns a success message when incremental is true and d
     t.deepEqual(result, expectedResult);
 });
 
-ava('.maintainHeading() returns a success message when incremental is true and direction is right', (t) => {
+ava('.maintainHeading() returns a success message when incremental is true, direction is right, and a 2-digit numeral is used for the increment', (t) => {
     const directionMock = 'right';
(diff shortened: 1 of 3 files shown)


test/aircraft/Pilot/Pilot.spec.js
```diff
@@ -1063,6 +1063,21 @@ ava('.maintainHeading() returns a success message when incremental is true and d
...
+        {
+            log: 'turn 5 degrees right',
+            say: 'turn 5 degrees right'
+        }
+    ];
+    const aircraftModel = new AircraftModel(ARRIVAL_AIRCRAFT_INIT_PROPS_MOCK);
+    const result = aircraftModel.pilot.maintainHeading(aircraftModel, 5, directionMock, true);
+
+    t.deepEqual(result, expectedResult);
+});
```
erikquinn: Actually, lint-diff shows there are trailing spaces on this line, please remove when able :)

JPmAn24:
@erikquinn Done! Could you also check #1073?

PR #1404, opened 2019-06-22
refactor/1403 Fix A321 Service Ceiling

### LINKED ISSUE #1403: Unrealistic Cruising Altitudes + A321 Service Ceiling
Browser: Running the sim locally

Airport: Any

Approximate play time before issue: Anytime

#### Issue Description
image
Aircraft always use their service ceiling as crusing altitude, which is unrealistic. Maybe there is a way to have dynamic cruise levels.

Also, the A321's ceiling is 41,000ft, but in the screenshot you can see an A321 crusing at FL430.

JonasStuhlberger:
Resolves #1403

The purpose of this pull request is to correct the A321 Service ceiling

Previously FL430
Now FL410

Source: https://www.skybrary.aero/index.php/A321


erikquinn:
This PR goes way out of scope, changing things beyond its purpose... If it was just changing the ceiling of the A321, I would merge it. Instead, all of the other changes must be looked over in detail.

In the future, please keep PRs limited to their purpose, and open others for other purposes rather than lumping them together like this.

JonasStuhlberger:
Okay, I looked over this and realized I made 12 commits on the wrong branch. I will create a new pull request regarding this issue. Thank you!

PR #1411, opened 2019-07-13
devops/1410

### LINKED ISSUE #1410: build:scripts:server fails to copy server.js to public/
running `npm run build` fails to copy `server/index.js` to the destination directory

### LINKED ISSUE #1412: upgrade deps and npm audit, v2

n8rzz:
Resolves #1412.

* removes server/index.js and express dependencies
* moves npm start to use npx http-server instead
* removes references to server in gulp scripts
* removes unused dependencies


n8rzz:
It _should_ fix the issue, yes. But I’m on a Mac, so I have no way to test this on windows. It always worked for me 😉

Anyone willing and able to just checkout the branch, delete `public` run the build and then start the app?

JPmAn24:
Just took a look at this, and things are looking good. I'm currently on vacation so can't check if it actually fixes the deletion of `index.js` until I get back late on Thursday. 

Thanks for doing this @n8rzz!

erikquinn:
Still trying to look for a solution here. It initially seems like this doesn't solve it, but that doesn't mean it might not still be the better approach-- it's just that the same issue of files not being copied correctly still persists. Here's a step-by-step process below which I followed to reliably see some behaviors.

When you have further input/advise/changes, it'll be easy for me to go back through them and see if we are then getting the correct result here. For now, I can't find a way to get this PR to load on localhost, while with some manual copying, I can get develop to do so. Anyway, the processes are below:

---

Process on develop:
1. Manually delete `node_modules`, and run `npm cache verify`, then `npm install`.
2. Delete `./public` folder.
2. Do `npm run build` and `npm run start`.
3. Observe the failure, stating "Error: Cannot find module './public/assets/scripts/server/index.js'"`.
4. Manually copy `./src/assets/scripts/server/index.js` to `./public/assets/scripts/server/`.
5. Do `npm run start`.
6. Observe the server running and the page being served.
7. Stop the server (Ctrl+C), re-run `npm run build` and `npm run start`.
8. Observe the failure message again.

Process on devops/1410:
1. Manually delete `node_modules`, and run `npm cache verify`, then `npm install`.
2. Delete `./public` folder.
3. Do `npm run build` and `npm run start`.
4. Observe the server start, but no page be served at `localhost:3003`. The `public` folder contains no file named `index.js`, including in any subfolders.
5. Observe that navigating to `./public/index.html` and opening the file shows the loading screen but never finishes load.
6. Copying `./src/assets/scripts/client/index.js` to `./public/assets/scripts/client/` or `./public/assets/scripts/server/` do not yield a page which loads any more fully.
6. Unsure how to proceed from here.

n8rzz:
Hmm 🤔 

## Previous state

* A small `Express` server was used to serve all assets during development
* a `gulp` task existed to simply copy that simple server from `src` to `public`

## New state

* we use an npm package called `http-server` which does, essentially, the same thing: acts a simple server to server up assets during development
* we never actually install this dep, instead we use `npx`.  this means we can run a thing without having to actually install it
*` http-server` defaults to serving a `public` dir, so you should be able to go to `localhost:3003` and be done
* on a mac, this fires up right away just like it did before.

This whole issue stemmed from a windows issue in the first place.  I wonder if this introduces a new, different windows issue?  What sort of output do you see after `npm run start`?

Screen Shot 2019-07-18 at 9 45 27 PM

wrksx:
Waiting for the new server, if anyone needs the fix for this, change tools/tasks/scriptTasks.js:36 to this
```js
const buildServer = () => gulp.src('**/*.js', {cwd: OPTIONS.DIR.SRC_SCRIPTS_SERVER})
```
Instead of
```js
const buildServer = () => gulp.src([path.join(OPTIONS.DIR.SRC_SCRIPTS_SERVER, '**/*.js')])
```
Worked for me on windows 8.1 (using cygwin).

wrksx:
I just tried it here on windows 8.1 + cygwin and it didn't work as expected.
It starts the server, and I can see HTTP requests logged on the server output.

Error using Chrome browser:
```
ERR_INVALID_REDIRECT
```

Error using Firefox browser: 
```
Corrupted Content Error

The site at http://127.0.0.1:3003/ has experienced a network protocol violation that cannot be repaired.

The page you are trying to view cannot be shown because an error in the data transmission was detected.
```

Error using wget command: 
```
$ wget http://127.0.0.1:3003/
--2019-08-06 15:02:02--  http://127.0.0.1:3003/
Connecting to 127.0.0.1:3003... connected.
HTTP request sent, awaiting response... 302 Found
Location: // [following]
http://: Invalid host name.
```

wrksx:
So this issue is known and referenced here: https://github.com/http-party/http-server/issues/525
If i understand correctly, the temporarily recommended fix is to revert back to http-server@0.9.0.

I tested manually by running `npx http-server@0.9.0 -p 3003 -c-1` and it worked.

I believe it's not a good idea to use npx http-server without explicitly depending on it. I think it's a 
 better practice to use the devDependencies to specify we depend on `http-server@0.9.0` that way the version is fixed. The npm start script would still work fine and use the locally installed version.

I tested with `"http-server": "0.9.0"` in the `devDependencies` and it worked.

n8rzz:
In most cases I would agree with you, this should live in `devDependencies`.  However, in this case, I think it makes sense to use `npx` for these reasons:
* this server is only needed to run the app locally
* it's a convenience, a developer could use any local server to run the app
* it reduces additional dependencies (because `npx`) and speeds up installs
* one of my goals here is to remove the local server (and it's dependencies) entirely

I like calling out a specific version, that's generally a good practice anyway.  I'd lean towards updating the command in this PR like you posted above referencing the version.

erikquinn:
The initial suggestion from @wrksx seems to resolve the issue. Recommend we go with that (posted PR #1415), and close this or otherwise update it once that fix goes in.

erikquinn:
Seems to no longer be needed. Closing.

PR #1423, opened 2019-08-18
enhancement/471 - Rename RouteModel to RouteStringModel

### LINKED ISSUE #471: Rename RouteModel to RouteStringModel
the current name is misleading and this class has taken on a different shape that what was originally intended.  `RouteStringModel` would be a much more appropriate name for this class.

- [ ] rename class
- [ ] update references

JPmAn24:
Resolves #471 

The purpose of this pull request is to rename RouteModel to RouteStringModel

### DIFF
--- a/test/aircraft/FlightManagementSystem/RouteModel.spec.js
+++ b/test/aircraft/FlightManagementSystem/RouteModel.spec.js
@@ -5,7 +5,7 @@ import _isArray from 'lodash/isArray';
 import _isEmpty from 'lodash/isEmpty';
 import _map from 'lodash/map';
 import LegModel from '../../../src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel';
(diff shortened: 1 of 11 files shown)


test/aircraft/FlightManagementSystem/RouteModel.spec.js
```diff
@@ -5,7 +5,7 @@ import _isArray from 'lodash/isArray';
 import _isEmpty from 'lodash/isEmpty';
 import _map from 'lodash/map';
 import LegModel from '../../../src/assets/scripts/client/aircraft/FlightManagementSystem/LegModel';
-import RouteModel from '../../../src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel';
+import RouteStringModel from '../../../src/assets/scripts/client/aircraft/FlightManagementSystem/RouteStringModel';
```
n8rzz: This filename should also be updated

Issue #1427, opened 2019-08-31
Initial spawn rate artificially high


oobayly:
Browser:
Version 76.0.3809.132 (Official Build) (64-bit)

Airport:
EGCC - Manchester

Approximate play time before issue:
Immediate

#### Issue Description
The initial spawn rate is artificially high, within the first few minutes there will be about 13 arrivals, After that the rate drops markedly. Rather than having an initial "stampede" of arrivals, followed by a relative trickle, could the initial spawning be done so it ramps up?
For reference, EGCC will have an average of 23 aircraft movements per hour, but can (according to schedules) have 50+ movements per hour during busy periods.

#### Steps to reproduce
Load EGCC and wait a few minutes

PR #1429, opened 2019-08-31
refactor/1428 - KCVG Arrivals Colliding

### LINKED ISSUE #1428: KCVG arrivals colliding
Browser: Google Chrome

Airport: KCVG

Approximate play time before issue: Immediately

#### Issue Description
The arrivals on a few different patterns collide on entry into airspace. For example, the BRUSH3 and GAVNN4 arrivals have aircraft spawn on top of each other.

Suggested fix: Consolidate spawnPatterns on top of eachother into one.

#### Steps to reproduce
- Load KCVG
- Wait for first arrivals to arrive

JPmAn24:
Resolves #1428 

The purpose of this pull request is to adjust/consolidate arrivals so that they don't collide at KCVG

### DIFF
--- a/assets/airports/kcvg.json
+++ b/assets/airports/kcvg.json
@@ -1282,7 +1282,7 @@
             "altitude": 34000,
             "speed": 280,
             "method": "random",
-            "rate": 1.5,
+            "rate": 1.7,
             "airlines": [
                 ["aal", 1],
                 ["aay", 4],
@@ -1297,7 +1297,7 @@
                 ["gjs", 5],
                 ["gti", 8],
                 ["jia", 10],
-                ["n/fastGA", 7],
+                ["n/fastGA", 14],
                 ["pac", 5],
                 ["rpa", 8],
                 ["skw", 3],
@@ -1313,7 +1313,7 @@
             "altitude": 31000,
             "speed": 280,
             "method": "random",
-            "rate": 2,
+            "rate": 2.2,
             "airlines": [
                 ["aal", 1],
                 ["aay", 4],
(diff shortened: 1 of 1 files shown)


erikquinn:
Did they actually conflict? Or were they just stacked on top of each other at different altitudes?

JPmAn24:
@erikquinn Yes they did conflict.

erikquinn:
Checking on my phone I show that they are all separated. The only thing that's weird is the way they spawn stacks everywhere at the beginning. And that's only a result of the way the prespawn logic currently works (and will be improved later).

@fechulo it's up to you guys what you want to do, whether it's keeping the low arrivals in there so the tiny aircraft aren't on the big jet arrivals, or removing them so we don't start with stacks everywhere. Just wanted to point out that it isn't exactly a "bug" being fixed here, it's more of an airport design change.

Fechulo:
@erikquinn I guess I just assumed I had originally added conflicting spawnPatterns when I made the airport (something which I did do for a number of airports) and din't bother check what the actual problem is. Apparently I actually made sure the spawnPatterns didn't conflict and, other than the awkward overlapping, a conflict outside controlled airspace shouldn't be possible the way it's currently done. @JPmAn24 If you can show that there is actual separation infringements caused by the current spawnPatterns, then we can merge this. Otherwise, I'd rather keep the old code.

erikquinn:
image

Finally on a real computer 😌

The screenshot is what I see. As I fast forward, they stay at those altitudes of 120 and FL200, so they never do conflict. So it sounds like we'll keep them in the file. When #1427 is addressed, the weird stacks should no longer be a thing.

Closing.

JPmAn24:
@erikquinn Thanks. Now for me to test #1427.

PR #1430, opened 2019-09-01
bugfix/1420 - Prespawned aircraft spawn with heading 360

### LINKED ISSUE #1420: Prespawned aircraft spawn with heading 360
Airport: EDDF, EDDL, KATL, KBOS, KLAS (probably every airport)

Screenshot_305
The last aircraft always spawns with heading 360.
Green line: How the aircraft should fly
Red line: What the aircraft does instead

wrksx:
Resolves #1420

Fix a condition that was used to find the next waypoint from a prespawn position.
The non-strict condition resulted in computing a heading from and to the same point, thus returning an incorrect arbitrary value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
-# 6.14.0 (September 1, 2019)
+# 6.14.0 (October 1, 2019)
 ### New Features
 
 
 ### Bugfixes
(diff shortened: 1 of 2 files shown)


erikquinn:
Thanks @wrksx!

PR #1433, opened 2019-09-03
bugfix/1432 - Aircraft exiting hold on their own (on develop)

### LINKED ISSUE #1432: Aircraft exiting hold after entry in beta
Browser:
Chrome: Version 75.0.3770.100 (Official Build) (64-bit)
OpenScope version: devel (yes I know
Airport:
EGCC, LSZH

Immediate

#### Issue Description

Aircraft enter a hold as expected, but never return back to the fix to complete the circuit, and continue on (eventually leaving the area). Holding works as expected in v.6.13
I've noticed that a lot of work regarding wind has been done between v.6.13 and devel, but I haven't found out when the regression was caused as I've only started perusing the code.

#### Steps to reproduce

Allow an aircraft to enter a procedure hold, or explicitly command a hold.

oobayly:
Resolves #1432

Fixes issue where aircraft never turned back to the hold fix, thereby
exiting the hold. Uses the ground track instead of heading so wind
doesn't affect the tests.

### DIFF
--- a/src/assets/scripts/client/constants/aircraftConstants.js
+++ b/src/assets/scripts/client/constants/aircraftConstants.js
@@ -193,7 +193,7 @@ export const PERFORMANCE = {
     MAXIMUM_DISTANCE_CONSIDERED_ESTABLISHED_ON_APPROACH_COURSE_NM: 0.0822894, // appx. 500 feet
 
     /**
-     * Maximum angular differce from the approach course heading to consider the aircraft close
+     * Maximum angular difference from the approach course heading to consider the aircraft close
      * to be "established on the approach course", which is an important condition for applying
      * rules of separation.
      *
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/constants/aircraftConstants.js
```diff
@@ -203,6 +203,16 @@ export const PERFORMANCE = {
      */
     MAXIMUM_ANGLE_CONSIDERED_ESTABLISHED_ON_APPROACH_COURSE: 0.0872665, // appx. 5 degrees
 
+    /**
+     * Maximum angular difference from the hold inbound heading to consider the aircraft close
```
wrksx: Isn't that the outbound heading you've been dealing with?

oobayly: Ah, for the love of $DEITY, You wouldn't believe how many times I tried proof reading a diff that changes so few lines...

n8rzz:
**approved!** (from me) Though, I would love to see a test or two added for this

erikquinn:
image

For the tests, I find it helpful to click the red circled link from this page, which will show you which parts of your new code are not tested:

image

For anywhere that shows that coverage _fell_ as a result of your changes (red down arrow), you can click on the file and it will show the specific changes which are not covered by any tests. Tests are added in the `test` folder with `.spec.js` file extensions, and you can generally follow the patterns you see in order to get them to work. The basic idea of them is to lay out how something is supposed to work, what results it should produce, etc. That way, if someone comes along and changes your work later, or breaks something else that would break your work, the tests sound the alarm for us rather than releasing the changes and waiting for users to start coming in with bug reports. Generally, we try to keep increasing our test coverage as much as possible, and we try not to add any untested code to the repo.

Maybe I'll take a look in a bit and see if I can add some quick ones to cover the changes you made. Just thought you'd like an overview of how it all works since you mentioned that you weren't too familiar with the process. 😃

erikquinn:
All done! This test was actually quite tricky, you can feel free to have a look though if you like:

https://github.com/openscope/openscope/pull/1433/files#diff-62659f25f18808e5f03c855d16cd6c68

PR #1435, opened 2019-09-03
refactor/1434 - Update EINN to AIRAC 1909

### LINKED ISSUE #1434: Update EINN procedures
Update EINN procedures as per the current AD (03 September 2019)

oobayly:
Resolves #1434

Refactor EINN existing procedures to be in line with the current AD. I haven't added any new routes to use the unused procedures, but it does make the airport usable again.


Fechulo:
Hey @oobayly Thanks for your work on bringing back Shannon, can't wait to see it back in the sim. For the airport to be brought back, it must meet all of the [standard requirements](https://github.com/openscope/openscope/blob/develop/documentation/airport-file-standards.md). The only thing currently missing is an airport guide. You can take a look at some [examples](https://github.com/openscope/openscope/tree/develop/documentation/airport-guides) for _'inspiration'_. When you think the PR is ready to be merged, request a review either from me or the airport-reviewers team.

oobayly:
Thanks for the feedback
* Just to confirm, the top one is the ICAO name. I assumed the short one as that is what was used by Navigraph.
image
* I'm actually working a local test branch that implements tiered airspace, the current code checks to see if the aircraft is below the ceiling (ignores floor) and within the bounds of the perimeter airspace (first item), so the way I've left it does work. Regardless, I'll move the interior ones into `_airspace`
* Coastline, I agree. The other option I had was to clip the coastline to the video map bounds, but that seemed even stranger...
* I was getting some strange results when trying to get the aircraft to turn at 500ft. I added an invisible waypoint (_RWY2401DME) 1NM from the end of RW24. The climb rate being 550ft / NM
 I've just removed the BEVES fix from the SID and tried getting a screenshot of the issue and it all looks fine now. Strange...
* The turn point (600ft) for RW06 is pretty much on SHA, so I used that rather than adding another waypoint. I can do so if desired though. It looks like I can remove the OL points though, seeing as the 
BEVES point is no longer needed.

Completed;
* Fix guide typos
* Extend centrelines and add Coonagh runway to the video map
* Update ENS -> EKNIS (I kept seeing that and for some reason never did it...)
* Fixed speeds
* Fixed MOMIN typo
* Removed the hold at DERAG and ELPOM, and the associated upper hold limits

Fechulo:
Great to hear someone's working on the tiered airspace! It should be easy to bring the other airspace sections back, once that's fully working.

Regarding the coastline, I agree it would look even stranger if we clip it to the airspace. Only other option is to have the entire Irish coastline on the map, which might be a bit much... I'm happy to leave it as is, I'll leave it up to you

Regarding departures off of RWY 06, even though it might look like the SID passes on top of SHA on the charts, in practice it probably doesn't. I would prefer if there was an invisible fix, just like you now have with RWY 24. This fix should be crossed at or above 600ft.

I see you renamed ENS to EKNIS, however they're two different fixes with different coordinates. ENS was a physical NDB stationed somewhere north of Shannon (It's probably still there but unused for whatever reason). EKNIS was a waypoint created to replace the now defunct ENS, but its coordinates are (very slightly) different. I don't think the difference is large enough to even notice looking at the scope, but I would rather use the new the coordinates: `N52.905147329005445, W8.927856434613213`

oobayly:
Serves me right for being lazy! I hadn't considered that SHA was slightly off the runway centreline. Also I don't know why I didn't notice the altitude limit was wrong either.

I was tempted to add the entire coastline (and still am slightly), but am also concerned whether it's an unnecessary performance hit. trying to draw all those extra lines.

oobayly:
Hopefully that should be all the changes made as requested.

oobayly:
Hopefully we'll be able to dispense with the coastline soon as there's a PR to add water to the terrain (it's currently a fairly simple change, but I foresee it getting more complicated). Many thanks for reviewing...

PR #1437, opened 2019-09-09
enhancement/1436 - Update airport maps format and render all layers

### LINKED ISSUE #1436: Render all map layers for airport
Currently, the maps object in the aiports file has implicit support for multiple "layers", but the `CanvasController` only renders the `base` layer.

It's a trivial change to iterate through all the layers and render each of them on the canvas. The benefit for doing this is that if part of the videomap is updated, diffs will be a little less overwhelming.
``` js
for (let mapKey in airportModel.maps) {
    const mapList = airportModel.maps[mapKey];
    if (!mapList || !mapList.length) {
        continue;
    }
    for (let i = 0; i < mapList.length; i++) {
        const mapItem = mapList[i];
        cc.moveTo(
            CanvasStageModel.translateKilometersToPixels(mapItem[0]),
            -CanvasStageModel.translateKilometersToPixels(mapItem[1])
        );
        cc.lineTo(
            CanvasStageModel.translateKilometersToPixels(mapItem[2]),
            -CanvasStageModel.translateKilometersToPixels(mapItem[3])
        );
    }
}
```

However, while this is the simplest method, is it the best? For a bit of future proofing It could be preferable to format the layer similarly to the restricted values:
``` json
"maps": {
  "base": {
    "name": "Base Map",
    "lines": [
      ["51.0", "-10.0", "53.0", "-8"]
    ]
  },
  "centrelines": {
    "name": "Extended Centrelines",
    "lines": [
      ["51.0", "-10.0", "53.0", "-8"]
    ]
  },
  "coastline": {
    "name": "Coastlines",
    "lines": [
      ["51.0", "-10.0", "53.0", "-8"]
    ]
  }
}
```
This way, if desired in the future the UI would be able to specify which layers to display, along with friendly names. Again, it would be relatively trivial to refactor all the existing airports to use this type schema, although it might break tools that people may have used.

oobayly:
Resolves #1436

The purpose of this pull request is to render all map layers in the aiport.maps object.

airports.map object has been updated to be an array of objects templated as:
```json
{
  "name": "<Layer name>",
  "hidden": false,
  "lines": [
    [51, -10, 53, -8]
  ]
}
```

* Add a `MapLayerModel` that handles the building of the layer
* Update `CanvasController` to render each layer
* Include a `hidden` boolean property to prevent the layer from being rendered


src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+     * @param map {object}
+     * @param airportPosition {StaticPositionModel}
+     * @param magneticNorth {number}
+     */
+    constructor(map, airportPosition, magneticNorth) {
+        super();
+
+        if (!map || !airportPosition || !_isNumber(magneticNorth)) {
+            // eslint-disable-next-line max-len
+            throw new TypeError('Invalid parameter, expected map, airportPosition and magneticNorth to be defined');
```
n8rzz: you can break this up:
```js
throw new TypeError(
    'Invalid parameter, expected map, airportPosition and magneticNorth to be defined
');
```

then you can remove the `eslint-disable`

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+
+        /**
+         * A flag indicating whether the layer should be hidden
+         *
+         * @for MapLayerModel
+         * @property hidden
+         * @type {boolean}
+         * @default false
+         */
+        this.hidden = false;
```
n8rzz: booleans should be prefixed with `is` or `has`.  so, in this case: `this.hidden` -> `this.isHidden`

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+     * @method _init
+     * @param map {object}
+     * @param airportPosition {StaticPositionModel}
+     * @param magneticNorth {number}
+     * @private
+     */
+    _init(map, airportPosition, magneticNorth) {
+        this.hidden = map.hidden === true;
+        this.name = map.name;
+        this._buildMapLines(map.lines, airportPosition, magneticNorth);
```
n8rzz: missing empty line between assignments and method call

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+
+        return this;
+    }
+
+    /**
+     * Create the array of map lines
+     * @for MapLayerModel
+     * @method _buildMapLines
+     * @param map {object}
```
n8rzz: `map` -> `lines`

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+     * @param airportPosition {StaticPositionModel}
+     * @param magneticNorth {number}
+     * @private
+     */
+    _buildMapLines(lines, airportPosition, magneticNorth) {
+        _forEach(lines, (line) => {
```
n8rzz: could we use `map` here instead?  probably not because `lines` is an object?

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
...
+            this.lines.push(lineVerticesRelativePositions);
+        });
+    }
+
+    /**
+     * @for MapLayerModel
+     * @method reset
+     */
+    reset() {
+        this.hidden = true;
```
n8rzz: this should probably be `false`

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1908,16 +1908,22 @@ export default class CanvasController {
...
-                -CanvasStageModel.translateKilometersToPixels(mapItem[1])
-            );
-            cc.lineTo(
-                CanvasStageModel.translateKilometersToPixels(mapItem[2]),
-                -CanvasStageModel.translateKilometersToPixels(mapItem[3])
-            );
+        for (let map of airportModel.maps) {
+            if (!map || !map.lines.length || map.hidden) {
+                continue;
+            }
```
n8rzz: missing empty line after closing `}`

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1908,16 +1908,22 @@ export default class CanvasController {
...
+            const lines = map.lines;
+            for (let i = 0; i < lines.length; i++) {
```
n8rzz: missing empty line above `for`

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1908,16 +1908,22 @@ export default class CanvasController {
...
-            );
-            cc.lineTo(
-                CanvasStageModel.translateKilometersToPixels(mapItem[2]),
-                -CanvasStageModel.translateKilometersToPixels(mapItem[3])
-            );
+        for (let map of airportModel.maps) {
+            if (!map || !map.lines.length || map.hidden) {
+                continue;
+            }
+            const lines = map.lines;
```
n8rzz: you could use destructuring here:
```js
const { lines } = map;
```

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1908,16 +1908,22 @@ export default class CanvasController {
...
+            for (let i = 0; i < lines.length; i++) {
+                const mapItem = lines[i];
```
n8rzz: missing empty line after assignment

src/assets/scripts/client/airport/MapLayerModel.js
```diff
@@ -0,0 +1,118 @@
+import _forEach from 'lodash/forEach';
+import _isEqual from 'lodash/isEqual';
+import _isNumber from 'lodash/isNumber';
+import _map from 'lodash/map';
+import BaseModel from '../base/BaseModel';
+import DynamicPositionModel from '../base/DynamicPositionModel';
+import StaticPositionModel from '../base/StaticPositionModel';
+
+/**
+ * A collection of lines that represent a map layer.
```
n8rzz: **Soooooo** glad to see this split out!

what, exactly, _is_ a `map layer`?   could you add a little bit more here about: 
* what this is for
* why you would use it
* what consumes this

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -162,13 +163,13 @@ export default class AirportModel {
         this._runwayCollection = null;
 
         /**
-         * Dictionary of polys that make up any airport video maps
+         * List of map layers that make up any airport video maps
          *
          * @property maps
-         * @type {object}
+         * @type {array}
```
n8rzz: * this is correct, but incomplete. `array` -> `array<MapModel>`
* I'd encourage you to rename this property as well: `map` -> `mapModelList`

In the future, it may make sense to add an additional layer here known as a `Collection`.  So, in a possible future state, this would be instantiated as `this.maps = new MapLayerCollection(...args);`.  Probably isn't needed now, but its a strategy we've used elsewhere to encapsulate domain logic.

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: Copy-pasta from the original!

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: I hadn't spotted that convention, the question is, should the json also use isHidden?

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: Again, I didn't know about that convention - I copied the style from AirspaceModel where there's no empty line.

I guess you mean this?
```js
    _init(map, airportPosition, magneticNorth) {
        this.isHidden = map.hidden === true;
        this.name = map.name;
        
        this._buildMapLines(map.lines, airportPosition, magneticNorth);

        return this;
    }

```

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: Yup, renamed the variable. It should also be typed as an array.

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: Absolutely.

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: The param type was wrong, it should have been an array, so I think `lines` correctly describes the contents
```json
"lines": [
    ["N45.00374744", "W063.74988584", "N45.00715629", "W063.74650952"],
    ["N45.00715629", "W063.74650952", "N45.00715649", "W063.74650992"]
]
```

src/assets/scripts/client/canvas/CanvasController.js
oobayly: Thanks for pointing that out. I'd spotted the usage elsewhere and hadn't looked into the usage and reasoning.

src/assets/scripts/client/airport/MapLayerModel.js
oobayly: I'll look at other models to figure out the way that info should be formatted

src/assets/scripts/client/airport/AirportModel.js
oobayly: I did wonder, but again say array being used elsewhere so assumed that was the convention. I did consider a `MapLayerCollection`, but wanted to get the core idea together first.

oobayly:
Many thanks for all your feedback, and for explaining the code styling in detail, as well as the various errors.

src/assets/scripts/client/airport/AirportModel.js
oobayly: In fact, the name layer is pretty much redundant. The model is already contained in an object called `maps`, so perhaps MapModel is better.

erikquinn: I agree the layer is not needed in the name. `MapModel` and `.mapCollection`

src/assets/scripts/client/airport/MapLayerModel.js
erikquinn: Yep, they should match each other directly.

src/assets/scripts/client/airport/MapLayerModel.js
erikquinn: Correct. Any time you change between assignments, method/function calls, or returns, there should be a blank line. Otherwise, no blank lines. So your example is correct.

src/assets/scripts/client/airport/MapLayerModel.js
n8rzz: oh, perfect.  then we can use `.map()` here instead of `_forEach`

oobayly: Ah, that's what you meant. I had thought you were referring to the variable name...

src/assets/scripts/client/airport/AirportModel.js
oobayly: As suggested, I've added MapCollection too

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,140 @@
...
+        return this._items;
+    }
+
+    /**
+     * @for MapCollection
+     * @method getVisibleMapLines
+     * @return {array<array<object>>}
+     */
+    getVisibleMapLines() {
+        const filtered = this._items.filter((map) => {
```
n8rzz: this can be simplified by using `reduce`:
```js
return this._items.reduce((sum, map) => {
    if (map.isHidden || map.lines.length === 0) {
        return sum;
    }

    return [
        ...map.lines,
        ...sum,
    ];
}, []);
```

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,140 @@
...
+        this._buildMapModels(mapJson);
+    }
+
+    /**
+     * Tear down the instance and destroy any instance property values
+     *
+     * @for MapCollection
+     * @method destroy
+     */
+    destroy() {
```
n8rzz: you'll want to reset all class properties in this method

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,140 @@
...
+         * @default #_items.length
+         */
+
+        /**
+         * @property _airportPositionModel
+         * @type {StaticPositionModel}
+         * @default null
+         * @private
+         */
+        this._airportPositionModel = null;
```
n8rzz: you may not want to keep these here.  instead, accept them as params (like you have now), but then just pass these off to the new `MapModel` instances

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,140 @@
...
+     * @return {array<MapModel>}
+     */
+    get maps() {
+        return this._items;
+    }
+
+    /**
+     * @for MapCollection
+     * @method getVisibleMapLines
+     * @return {array<array<object>>}
```
n8rzz: does this actually return a 2d array?  or is it actually `array<MapModel>`?

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1895,7 +1895,7 @@ export default class CanvasController {
     _drawVideoMap(cc) {
         const airportModel = AirportController.airport_get();
 
-        if (!_has(airportModel, 'maps') || !this._shouldDrawVideoMap) {
+        if (!_has(airportModel, 'mapCollection') || !this._shouldDrawVideoMap) {
```
n8rzz: we may want to create a method on the `airportModel` to handle this.  basically move this logic there within a method, so you could do: 
```js
if (!airportModel.hasMapCollection()) {
    return;
}
```

src/assets/scripts/client/airport/MapCollection.js
oobayly: Now with the airportPositionModel and magneticNorth variables moved to being params, this is essentially empty.

Or should this include _items property too? I took the usage from RunwayCollection which doesn't destroy _items.

src/assets/scripts/client/airport/MapCollection.js
oobayly: As is, it does return a 2d array . The rationale was that it returns all the lines that are to be displayed, but I wanted to avoid the overhead of merging the array every time the canvas is redrawn. I've only just realised that the videomap isn't collected every time an aircraft position changes.

The other reason I don't actually use it was because if the UI is ever updated to allow the user to select which maps are rendered, I wasn't sure if this would be done by filtering the MapModels in the MapCollection, or in the CanvasController.

Using this method, and doing the reduction in the MapsCollection would certainly make the CanvasController cleaner

src/assets/scripts/client/canvas/CanvasController.js
oobayly: Makes sense

src/assets/scripts/client/airport/MapCollection.js
oobayly: Clever.

n8rzz: haha, perhaps.  It just means 1 loop instead of 2 😉

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+     * @method _buildMapModels
+     * @param mapJson {object}
+     * @param airportPositionModel {StaticPositionModel}
+     * @param magneticNorth {number}
+     */
+    _buildMapModels(mapJson, airportPositionModel, magneticNorth) {
+        // Need to use forEach, as the value being passed can be either
+        // an array of map objects, or a dictionary of key -> object[]
+        _forEach(mapJson, (map, key) => {
+            if (!_has(map, 'lines')) {
```
erikquinn: Another way to do this is actually `if (!('lines' in map)) {` or `if ('lines' in map === false) {`. Kinda neat. No change needed at all, just thought I'd point it out.

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+                // Handle the existing map dictionary format too
+                map = {
+                    name: `Legacy-${key}`,
+                    lines: map
```
erikquinn: Clever! Typically, I strongly prefer to change all of the (currently active in the load list) airports to conform to the new standard, and consciously do _not_ support the old ones, for the sake of keeping them as consistent as possible. We used to have some crazy stuff going on in three or four airports, and moving to a mindset of consistency has made it a lot easier for everyone to share the work of updating the airports over time.

Just a thought, this is great for now. Maybe I can update the airports and remove this in a few days.

src/assets/scripts/client/airport/MapModel.js
```diff
@@ -0,0 +1,135 @@
...
+
+        /**
+         * A flag indicating whether the layer should be hidden
+         *
+         * @for MapModel
+         * @property isHidden
+         * @type {boolean}
+         * @default false
+         */
+        this.isHidden = false;
```
erikquinn: So here's where I'd like us to make a small fundamental change... Instead of setting whether or not the map is "hidden" in the definition of the map _itself_, which maps are shown/hidden should be determined in a singular place in the airport file, called `defaultMaps`. This would be an array of strings which match the map names.

Then, during the initialization of the MapModel, we will set the `.isHidden` property based on whether or not `airportModel.defaultMaps.includes(mapModel.name)`. This could be passed down during instantiation if needed, or accessed at the appropriate time.

This way, when the maps are super long, it's easy for us to go into the airport file and quickly change which maps to display on initial load. When turning maps on/off during the simulation, the maps can still be hidden/unhidden as necessary.

src/assets/scripts/client/airport/MapModel.js
```diff
@@ -0,0 +1,135 @@
...
+        this.name = map.name;
+
+        this._buildMapLines(map.lines, airportPosition, magneticNorth);
+
+        return this;
+    }
+
+    /**
+     * Create the array of map lines
+     * @for MapModel
```
erikquinn: Extra line between these

src/assets/scripts/client/airport/MapCollection.js
oobayly: I agree, this was more a stopgap so I didn't have to change multiple airport files before a format was agreed upon. Especially in light of your other comment.

I'm happy to go through and update all the files, it's not too onerous and I'd assumed it was part of the branch.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -264,6 +265,17 @@ export default class AirportModel {
...
 
+    /**
+     * A flag indicating whether the airport has a valid
+     * `MapCollection`
+     *
+     * @property hasMapCollection
+     * @return {boolean}
+     */
+    get hasMapCollection() {
+      return this.mapCollection != null;
```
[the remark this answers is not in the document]
oobayly: The problem is that I'm used to obsessively sanitising bad input, and that's flowed over into the code here! You're right, it should be validating the input when building the model

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * A flag indicating whether the `MapCollection` has any maps.
+     *
+     * @for MapCollection
+     * @property hasMaps
+     * @return {boolean}
+     */
+    get hasMaps() {
+        return this.length !== 0;
+    }
```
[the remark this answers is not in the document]
oobayly: Obsessive run-time validation strikes again...

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+     * @param mapJson {object}
+     * @param airportPositionModel {StaticPositionModel}
+     * @param magneticNorth {number}
+     */
+    constructor(mapJson, airportPositionModel, magneticNorth) {
+        super();
+
+        if (!_isArray(mapJson) && !_isObject(mapJson)) {
+            throw new TypeError(
+                `Invalid parameter passed to MapCollection. Expected an array or object but found ${typeof mapJson}`
```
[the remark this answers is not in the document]
oobayly: That's to permit the backwards compatibility, which we both agree can be dispensed of.

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+     * A flag indicating whether the `MapCollection` has any visible maps.
+     *
+     * @for MapCollection
+     * @property hasMaps
+     * @return {boolean}
+     */
+    get hasVisibleMaps() {
+        // for means we can return on finding the first match, rather
+        // than walking the whole array
+        for (let i = 0; i < this.length; i++) {
```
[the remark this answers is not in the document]
oobayly: Fair point

src/assets/scripts/client/airport/MapCollection.js
oobayly: Funnily enough, I think that was how the original check was done, but I changed to it to _has as I'd seen it used elsewhere and thought that that was the way to go. It'll be removed with the backwards compatibility support anyway,

src/assets/scripts/client/airport/MapModel.js
oobayly: > This way, when the maps are super long
Yup, that struck me when having to fold the JSON. That should be relatively simple to implement and test for.

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -162,13 +163,13 @@ export default class AirportModel {
         this._runwayCollection = null;
 
         /**
-         * Dictionary of polys that make up any airport video maps
+         * Collection of all `MapModel` objects that make up the airport's video map
```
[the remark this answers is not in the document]
oobayly: I'll have you know the department of redundancy department came up with that one...

src/assets/scripts/client/airport/MapCollection.js
n8rzz: Cool.

We've tried to encapsulate as much logic and _"work"_ as possible within these models and collections as possible.  That way the controllers can (in theory) be lighter.  They can concentrate on operating on the data, and not have to worry about assembling any of it.

In some places we are better about this than in others, but thats the general intent: MVC.

n8rzz:
> Not a problem. It's a bit of a learning curve, getting the code style and flow to conform, but it's definitely worth starring off with a simple enhancement, even if it has no obvious affect at the moment.

We are certainly more on the picky side of things, but it's with good reason.  This is a medium sized very complex app.  Consistency is incredibly important so that things can be predictable.  For new things, or new refactors, we are especially harsh because we want to maintain (or bring things up to) [standards](https://github.com/thenerdery/javascript-standards).

That said, we aren't afraid of pushback either here or on slack.  So if you have questions, please don't hesitate to poke us 😉

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * @for MapCollection
+     * @method _buildMapModels
+     * @param mapJson {object}
+     * @param defaultMaps {array<string>}
+     * @param airportPositionModel {StaticPositionModel}
+     * @param magneticNorth {number}
+     */
+    _buildMapModels(mapJson, defaultMaps, airportPositionModel, magneticNorth) {
+        this._items = mapJson.map((item) => {
```
erikquinn: `return mapJson.map(.....` instead of setting this._items

Because generally where we see "build", "generate", or "calculate" in method names, they are only assembling or calculating the answer and returning it, rather than actually using the answer.

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,180 @@
...
+     *
+     * @for MapCollection
+     * @method _init
+     * @param mapJson {object}
+     * @param defaultMaps {array<string>}
+     * @param airportPositionModel {StaticPositionModel}
+     * @param magneticNorth {number}
+     */
+    _init(mapJson, defaultMaps, airportPositionModel, magneticNorth) {
+        this._buildMapModels(mapJson, defaultMaps, airportPositionModel, magneticNorth);
```
erikquinn: `this._items = this._buildMaps(...`

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -433,36 +434,11 @@ export default class AirportModel {
...
-                    ...airportPositionAndDeclination
-                );
-                const lineVerticesRelativePositions = [...startPosition, ...endPosition];
-
-                outputMap.push(lineVerticesRelativePositions);
-            });
-        });
+    buildAirportMaps(maps, defaultMaps) {
+        this.mapCollection = new MapCollection(maps, defaultMaps, this.positionModel, this.magneticNorth);
     }
```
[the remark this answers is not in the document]
oobayly: Sure, it makes sense. I've removed the associated test in AirportModel too, but that's not an issue as the same tests were issued for the MapCollection.

oobayly:
All done. I'm almost disappointed that that's it. **Almost!**

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,173 @@
...
+        this._buildMapModels(mapJson, airportPositionModel, magneticNorth);
+    }
+
+    /**
+     * Tear down the instance and destroy any instance property values
+     *
+     * @for MapCollection
+     * @method destroy
+     */
+    destroy() {
```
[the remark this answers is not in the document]
erikquinn: We don't reuse the map collection, so no it's not really needed. I only mean that if you were to implement one, to call it `reset()` instead of `destroy()`. You can remove or rename, either way.

erikquinn:
@oobayly So is this ready, or are you still working?

oobayly:
This should be ready to merge, providing you're happy that I've dealt with the changes correctly.

src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -0,0 +1,181 @@
+import _isArray from 'lodash/isArray';
+import BaseCollection from '../base/BaseCollection';
+import MapModel from './MapModel';
```
erikquinn: We generally try to sort these by the file path instead of the object. When we import multiple things from a given file, it saves us from having to reshuffle them all around. But otherwise the order here is good-- lodash, then classes, then constants and such.

erikquinn:
Thanks @oobayly! 😃

PR #1442, opened 2019-09-19
feature/420 - Include water in terrain files

### LINKED ISSUE #420: Include water in terrain files
We can show hills and such, but being able to also show shorelines would be sweet, and give a lot more context to the terrain you see in certain airports (GCRR, EIDW, KSFO, many others).

oobayly:
Resolves #420 

The purpose of this pull request is to add support for rendering water contours in airport terrain files.

* Add theme colours for 0ft contours
* Fix rendering bug where `CanvasController._drawTerrain` returned before ``cc.restore()` was called twice.
* Only warn on terrain below sea level
* Update ENGM (Oslo) to include water

Edit: I've kept this to the bare minimum for the moment, even though I've seen stuff that could/should  be refactored.


erikquinn:
A bit bright though, I'll dim the color set for the default theme.

oobayly:
> Absolutely brilliant. Never would have thought at first glance this could be done with **LITERALLY SIX LINES OF CODE** changed. Hats off to your restraint to not want to tear things apart too, I'm sure you passed some ugly stuff looking through the canvas controller

You possibly haven't seen the python that I use to generate the polygons in QGIS. Sure it works, but it possible it would make a grown man cry :scream:

PR #1444, opened 2019-09-21
Enhancement/1431 - Add support for hold parameters for fixes

### LINKED ISSUE #1431: Add support for hold parameters for fixes
At the moment, holds _appears_ to be based on the reciprocal of the entry bearing into the hold. This mean that aircraft will often leave the hold at an unhelpful heading, especially if the hold is an IF.

It would mean having to add the hold parameters to the airport definition. The way I could see this being done wold be to:
1. Add a `HoldModel` that provides the following properties: leg length, inbound heading, turn direction, max & min altitudes, max & min speeds
2. The values could be specified in a similar manner to how fixes are specified in SIDs & STARs:
Hold DERAG, right turns, 1 minute, 237°, Max 220KT, MAX FL140, MHA 3000
`"TR|L1min|H237|S220-|A30+|A140-"`
3. Add a `holds` collection that references the name of the fix and the `HoldModel`
`["DERAG", "TR|L1min|H237|S220-|A30+|A140-"]`
4. **or** Extend the `fixes` so that it *optionally* provides a `HoldModel` property
`"DERAG": ["N52.86288437239589", "W8.58032227172346", "TR|L1min|S220-|A30+|A140-"]`

This way, when a hold is specified, either by the `AirportCommander`, or the `AircraftModel`:
1. A check is done as to whether the `inboundHeading` property is undefined
2. If it's undefined, find the fix and if a `HoldModel` exists for it use the values.
3. Otherwise, use the current fallback method of using the `bearingToHoldFix` value.

An additional bonus is that we could then validate if a hold is requested that is above or below the limits, speed restrictions, etc.

Possible Issues:
* Can a fix have contain multiple holds, depending on the STAR?. I've been playing around with refactoring EINN as I know it and it's a simple airport to start on, but it may be different for bigger airports.

oobayly:
Resolves #1431 

The purpose of this pull request is to add hold parameters for fixes in airports. This adds an array of objects that contain the inboundHeading, legLength and turnDirection properties.

It's pretty hacky at the moment, I don't really like the way it gets HoldModel from the current airport when constructing a FixModel. People may also take issue with null values being returned by the argumentParser (even though they are correctly handled in runHold). Aircraft may also end up doing a 270 if entering a hold from the opposite direction of the hold, I don't know what the expected behaviour would be in real life.

But other that that, it does seem to work. Aircraft end up holding in a neat stack. It also allows the fix's hold parameters to be overridden if the player specifies the leg length or turn direction in the hold command.


erikquinn:
Some investigation should be done to see if this is a safe approach-- it's quite possible the same fix could be held over with two different published holding patterns (maybe right turns at odd altitudes and left turns at even altitudes.... or 1 minute legs from 8000 through 12000, and then 20 mile legs from FL220-FL280... you'd be surprised some of the whacky cases we've encountered while trying to support airports from various countries). It _may_ be necessary to define the holds in the procedures individually (though this could have some super annoying duplication). Or maybe a combination-- have a defined hold for a given fix, and STARs could have custom implementations which would override the default hold parameters.

Not that we have to go that deep into this anyway, but just want to take a good think through how we approach adding support for these hold parameters.

oobayly:
> Some investigation should be done to see if this is a safe approach

Definitely, this was more of a proof of concept.

>  it's quite possible the same fix could be held over with two different published holding patterns

I wondered about this too. I found that for EGGW the LOREL and ASKEY fixes are in the same position (the ASKEY hold is only used if the Barkway VOR is u/s). As you say other countries may not do this.

> maybe right turns at odd altitudes and left turns at even altitudes

Would this happen? It'd certainly keep flight crew on their toes if they had to change hold direction if their assigned altitude was changed by 1,000ft!

> or 1 minute legs from 8000 through 12000, and then 20 mile legs from FL220-FL280

I hadn't really considered this, but I can see which it could happen.

> you'd be surprised some of the whacky cases we've encountered while trying to support airports from various countries

I don't doubt it. I've stuck with airports that I know and have easy access to ADs (NAT or Eurocontrol) mainly. I'll have a peruse through STARs for various supported countries and see what outliers I can find. I also know a couple of ATPL pilots that might be able to help me here.

> It may be necessary to define the holds in the procedures individually

Another way could be to have the hold defined by both Fix and an optional Procedure name[s]. The waypoint could be built using the `holdParameters` for a null. If however the waypoint is created as part of a procedure, the `holdParameters` could be overwritten for that specif one. But I can see that getting messy very fast!

> Not that we have to go that deep into this anyway, but just want to take a good think through how we approach adding support for these hold parameters.

Completely agree, I in no way expected this to get merged as is, but it was a way of getting feedback. I could have done it on Slack, but this way it's kept in one place.

erikquinn:
>>maybe right turns at odd altitudes and left turns at even altitudes
>
>Would this happen? It'd certainly keep flight crew on their toes if they had to change hold direction if their assigned altitude was changed by 1,000ft!

Hah! No, I don't really think so-- I only meant that there might be some weird ass place where (for example) they run eastbound traffic just south of the holding pattern at low altitude and westbound traffic just north of the holding pattern at high altitude... Then the chart would say `holding at and below 13000: make right turns, holding at and above 14000: make left turns`, and ATC would only hold aircraft at 9k/11k/13k/14k/16k/etc, and the whole thing would just work. I'm not saying there isn't likely a much better way for them to do it, but I am saying that we've encountered even stupider edge cases than this. I've grown to try to consider every ridiculous thing people could try to do so we can consider supporting them before we go down a path that blocks out entire airports due to their weird ways of being (looking at you, LOWW...)

>Another way could be to have the hold defined by both Fix and an optional Procedure name[s]. The waypoint could be built using the holdParameters for a null procedure name. If however the waypoint is created as part of a procedure, the holdParameters could be overwritten for that specific one. But I can see that getting messy very fast!

I'll be honest, I'm not following a word of that 😬 

>Completely agree, I in no way expected this to get merged as is, but it was a way of getting feedback. I could have done it on Slack, but this way it's kept in one place.

Oh, certainly! I appreciate good documentation just as much as the next guy 👍

oobayly:
> I'll be honest, I'm not following a word of that grimacing

I don't blame you, I think I rewrote that several times and I'm still struggling to put it into words, so here goes...

The fix ABBOT could have multiple hold definitions:
* Right, 1nm
* Left, 1 min when part of procedures ABBOT1C & ABBOT1D
* Right 5 NM when part of the [ficticious] ABBOT1E procedure

``` json
    "holds": [
        {
            "fixName": "ABBOT",
            "turnDirection": "right",
            "legLength": "1min",
            "inboundHeading": 265
        },
        {
            "fixName": "ABBOT",
            "procedures": ["ABBOT1C", "ABBOT1D"],
            "turnDirection": "left",
            "legLength": "1min",
            "inboundHeading": 250
        },
        {
            "fixName": "ABBOT",
            "procedures": ["ABBOT1E"],
            "turnDirection": "right",
            "legLength": "5nm",
            "inboundHeading": 265
        },
```

Lets say the a Waypoint get created at ABBOT, as part of the ABBOT1C STAR:
1. Use `HoldCollection.findHoldByFix` to find the HoldModel for the Fix where no procedures are specified
`holdModels.find((hold) => hold.fixName == fixName && hold.procedures == null)`
2. This will set the default holdParameters to **"right 1min on the 265 radial"**
3. Now use `HoldCollection.findHoldByFixAndProcedure`, and if a match is found set the holdParameters using `WaypointModel.setHoldParameters`
4. Now holdParameters = **"left 1min on the 250 radial"**

Now, if we do a reroute (rr ABBOT hold ABBOT)
1. Use `HoldCollection.findHoldByFix` to find the HoldModel for the Fix where no procedures are specified
2. This will set the default holdParameters to **"right 1min on the 265 radial"**

Obvious issues with this is
1. Gets a bit confusing when rerouting. This will wipe out the procedure and the player may be left wondering why the rerouted aircraft is holding differently to the aircraft hold as part of the ABBOT1C STAR
2. Specifying min/max altitudes for the HoldModel would be easy. Selecting them correctly in the game does not sound fun!

I'm not sure if this has helped.

erikquinn:
Okay, yeah it does. Honestly though, I'm not so sure this is the best way to go about it-- defining the holds in a pile and then saying which one belongs with which procedures, etc. It seems only natural to instead include the holds as a part of the procedure definition for each STAR. If there are any holds that are actually predefined that are not on a STAR (I'm not familiar with this being the case, but I expect there _are_ charted holds just on jet airways some places), then they would be the only thing I would include in a `holds` that isn't within a procedure.

You seem to have put quite a bit of thought into this, could you think through what problems we might encounter if we set it up like the below? Should yield the same result, and I think would be more intuitive (though more tedious) to enter in airport files, and also easier to implement in the simulator code.

```javascript
...
"holds": [
    "ABBOT": { // used when the ABBOT WaypointModel is not generated by a ProcedureModel
        "turnDirection": "right",
        "legLength": "1min",
        "inboundHeading": 265
    }
],
"stars": {
    "ABBOT1C": {
        "icao": "ABBOT1C",
        "name": "Abbot One Charlie",
        "entryPoints": {
            "IDESI": [["IDESI", "A180"]]
        },
        "body": [["LAPRA", "A120|S250-"], ["@ABBOT", "A80+|A140-|S220-"]],
        "rwy": {
            "EGGW08": [],
            "EGGW26": []
        },
        "holds": {
            "ABBOT": { // used when the ABBOT WaypointModel is generated by "this" (ProcedureModel)
                "turnDirection": "left",
                "legLength": "1min",
                "inboundHeading": 250
            }
        },
        "draw": [
            ["IDESI", "LAPRA", "ABBOT"]
        ]
    },
    "ABBOT1D": {
        "icao": "ABBOT1D",
        "name": "Abbot One Delta",
        "entryPoints": {
            "LOGAN": ["LOGAN"]
        },
        "body": [["CLN", "S250-"], ["@ABBOT", "A80+|A140-|S220-"]],
        "rwy": {
            "EGGW08": [],
            "EGGW26": []
        },
        "holds": {
            "ABBOT": { // used when the ABBOT WaypointModel is generated by "this" (ProcedureModel)
                "turnDirection": "left",
                "legLength": "1min",
                "inboundHeading": 250
            }
        },
        "draw": [
            ["LOGAN", "CLN", "ABBOT"]
        ]
    }
},
...
```

Also as I think it through, both approaches have the added benefit of being able to actively separate the definition of a specific holding pattern from whether or not the aircraft should, by default, enter the holding pattern. In the STAR, you can say `, "ABBOT", ...`, or `, "@ABBOT", ...`, and from _that alone_, the aircraft will decide whether or not to mark true `isHoldWaypoint`. And even if it isn't a hold waypoint, it will still have the holding parameters defined.

oobayly:
>  It seems only natural to instead include the holds as a part of the procedure definition for each STAR.

I can certainly see the logic in doing it that way. My first thought was that it might end up with a lot of duplication, but I decided to follow the my own advice* and check how much it would happen. Taking KSFO as a busy airport it turns out that STARS don't share a lot of hold points, which I suppose makes sense. _Other  countries may be less sensible..._

I can't really see any issues with your approach either. It still won't handle the possibility of having multiple hold definitions for a fix as part of the same STAR.

One upshot of this discussion is that I've created a local branch that adds specifying the hold radial when using the hold command. While we hammer out the details on this, I might create a PR for that.

\* Ignore what you feel the situation is, back your claims up with numbers :grin:

oobayly:
Actually, I'm going to back-pedal a little. Having a look at at EGCC - Manchester, using a separate hold definition for every STAR would be quite a lot of duplication:
* DALEY x6 - DALEY2A, DALEY2B, DALEY4D, DALEY2E, DALEY1F, DALEY1G
* DAYNE x2 - DAYNE1B, DAYNE2A
* MIRSI x4 - MIRSI1A, MIRSI2B, MIRSI2C, MIRSI2D
* ROSUN x12 (6 used in openScope) - ROSUN** (I got bored of typing :grin: )
* I guess that's how we roll in the UK!

Granted, not all of those STARS are used openScope, but that amount of duplication increases the likelihood of introducing errors when it comes to updates

erikquinn:
Hm, yeah that is a lot. My two cents is that they shouldn't have 50 different stars into the same GD airport, but it's not like they care what I think.

So here's how I see it-- storing the holds in a HoldCollection is too inconvenient to move forward with. Updating the holds in the airport file for each and every star that holds there is also inconvenient. And since one has to lose, I say we make it the airport file (as usual). The cleaner we can keep the operations of the simulator itself, the faster we can add features with fewer bugs. And though it's not a popular opinion, I think we should do what's best for the simulator, even if it makes maintaining the airport files more difficult. If it becomes too difficult, that airport will just get dropped when an update breaks it, and that's how it goes sometimes. But we've got to keep our eye on the prize, which is moving the simulator itself forward.

Let's move ahead with defining the holds in the procedures. It will be in the procedure model somehow to generate waypoints with the hold parameters. If the RouteModel generates the waypoint directly (because it's a direct leg or airway leg), then there should be no hold parameters stored.

I can help with updating airport files to make this happen, but I do think it's the course we should take here.

oobayly:
> Hm, yeah that is a lot. My two cents is that they shouldn't have 50 different stars into the same GD airport, but it's not like they care what I think.

I bet the guys at Jeppesen say exactly the same thing.

> The cleaner we can keep the operations of the simulator itself, the faster we can add features with fewer bugs.

I wish I could do that in my day-job, I think I'm just inured to the pain of having to deal with copious amounts of backward compatibility (and not having to upset anyone)!

Ok, so from my understanding:
* Make the hold a Key-Value object, as opposed to an array
* Add a _holdCollection property to the `ProcedureModel`
* ...
* Profit

erikquinn:
Exactly, thanks. :Shipit:

erikquinn:
Let's also please add an optional `speed` parameter to the published holds. If it's not defined, then there is no speed restriction. Last addition to this PR, I promise. It's just the last step to being able to actually implement published holding patterns perfectly. After that, we'll be doing everything correctly except for holding pattern entries (which are hard!).

Aircraft should cross the holding fix at the specified speed.

image

oobayly:
So, I've implemented the speeds, though it may not be the best method of doing so.

* `WaypointModel.speedMaximum` has been set to a getter.
* If `isHoldWaypoint` is true **and** a speed restriction is set for the `holdParameters`, it'll return that speed
otherwise it falls back to return the value in _speedMaximum (current behaviour)
* I've added an artificially low speed restriction (150KIAS) to the EGGW ABBOT2D STAR to demonstrate the functionality.

oobayly:
So, this is how it's ended up:

* Each procedure has an optional `holds` property.
* When the `ProcedureModel` creates a route, it fetches the optional `holdParameters` for each Fix in the route. If available, it applies the parameters over the `DEFAULT_HOLD_PARAMETERS` for `WaypointModel`
* As there will be an amount of duplication when defining holdParameters in the airport.json, they are defined in a string (not the original JSON object format in the PR), similar to how the `WaypointModel` restrictions are defined:
`"<radial>|<turn>|<length>|<speedMaximum>"`
    eg. `"270|right|1min|S220-"` or `"341|left|3nm|S200-"`
This should make it easier for airport maintainers to update the `holds` during AIRAC cycles as a multi-line JSON object would make batch replacements tedious (if not error prone).
* The `HoldModel` strictly validates the string, throwing an error telling the developer which part is invalid. eg:
`Invalid legLength parameter for Fix 'ABBOT': 2km is not valid`
* A maximum speed restriction is also present. This is used by the `WaypointModel.speedMaximum` property. This means that the FMS will automatically decelerate the aircraft to the maximum hold speed, providing the `WaypointModel.isHoldWaypoint` is true
* The hold command now uses a _"patching"_ system, so only the parameters explicitly provided by the user will override the default hold paramters for that waypoint.

So, for the example @erikquinn gave the hold parameters for the procedure would be:
``` json
{
  "holds": {
    "PAATS": "231|left|4nm|S210-"
  }
}
```

src/assets/scripts/client/airport/AirportModel.js
```diff
@@ -371,6 +381,7 @@ export default class AirportModel {
         this.initial_alt = _get(data, 'initial_alt', DEFAULT_INITIAL_ALTITUDE_FT);
         this.rangeRings = _get(data, 'rangeRings', DEFAULT_RANGE_RINGS);
         this._runwayCollection = new RunwayCollection(data.runways, this._positionModel);
+        this.holdCollection = new HoldCollection(data.holds);
```
erikquinn: Trying to figure out if this is something we even want to do. TBD.

If we did keep generic holds that aren't part of a procedure though, they would belong in the NavigationLibrary. We are trying to move EVERYTHING that isn't explicitly about the airfield out of the AirportModel so we can be able to support multiple airports in the same session.

assets/airports/eggw.json
```diff
@@ -637,6 +637,9 @@
         "WOD"  : ["N51d27m10.02", "W000d52m43.68"],
         "XAMAN": ["N51d47m05", "E002d13m27"]
     },
+    "_holds": {
+        "HEMEL": "270|right|1min|S150-"
+    },
```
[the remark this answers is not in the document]
oobayly: Damn, I thought I removed that. Thanks

src/assets/scripts/client/airport/AirportModel.js
oobayly: That makes sense. I hadn't realised that that was an aim. I'm happy to remove it, as I saw the comments on Slack regarding whether a controller would ever order "hold as published"

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -308,19 +308,27 @@ export default class AircraftCommander {
...
+        }
         if (radial !== null) {
-            inboundHeading = radians_normalize(degreesToRadians(radial) + Math.PI);
+            // Radial is given as the outbound course, so it needs to be inverted
+            holdParameters.inboundHeading = radians_normalize(degreesToRadians(radial) + Math.PI);
+        } else {
+            // As radial has not been explicitly requested, we need to pass a "fallback"
+            // inboundHeading, as it's possible that a default inboundHeading doesn't exist for
+            // the `WaypointModel`s _holdParameters (eg. if the procedure JSON has no holds)
+            holdParameters.fallbackHeading = fixModel.positionModel.bearingFromPosition(aircraft.positionModel);
```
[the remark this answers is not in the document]
oobayly: Sure, the reason I implemented it like this was because I wanted the behaviour to replicate the existing behaviour (ie not breaking things).

`EZY97GP rr bpk..gwe16 hold gwe16`
image


I'll have a look at trying to get it to work as you describe, making sure that the readback will be also be correct if the hold fix requested is later on in the procedure.

src/assets/scripts/client/airport/AirportModel.js
erikquinn: Yeah I mean it's basically done so it seems silly to toss it, but I know we can at least plan to move it. :)

oobayly: It's entirely up to you. It's basically there as a result from original concept being being changed the current behaviour.

src/assets/scripts/client/aircraft/AircraftCommander.js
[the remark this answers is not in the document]
oobayly: So, this behaviour is corrected:
1. `AircraftCommander.runHold` calculates the `fallbackHeading` to use as the heading from the aircraft's current position to the holding fix (this is the existing behaviour)
2. `RouteModel.activateHoldForWaypointName` then calculates the inbound course to the holding fix from the preceding fix and uses this as `fallbackHeading`
If there is no preceding waypoint, then it uses the `fallbackHeading` from (1)
3. This is then passed all the way down to `WaypointModel.setHoldParameters`, where it is used if no default `inboundHeading` value is present.

src/assets/scripts/client/airport/AirportModel.js
erikquinn: Despite the encouraging talk on slack about them probably never being used, since it's in question, and it's already done, I say we keep it around. But it will need to move out of the AirportModel, and into the NavigationLibrary instead.

Some sparse internet resources do show that they _exist_ like this, even though I (as a controller) can attest that on a day-to-day basis, they generally aren't even known about, let alone used "as published" if they aren't on a STAR chart (where the overwhelming majority of holding is done in my experience). But since some fixes that aren't on arrivals definintely have flight-checked holding patterns that appear in the Low IFR Charts, it's best to keep the code alive just in case we need those holds. Creates some confusion about where to put holds (STAR vs directly in the airport `holds` key), but that can be sorted out with a little bit of thinking... 🤔 

https://www.boldmethod.com/learn-to-fly/maneuvers/what-you-should-know-about-holding/
https://nfdc.faa.gov/webContent/content8260/VA_EWOOD_REV5.pdf

image

erikquinn: Despite the encouraging talk on slack about them probably never being used, since it's in question, and it's already done, I say we keep it around. But it will need to move out of the AirportModel, and into the NavigationLibrary instead.

Some sparse internet resources do show that they _exist_ like this, even though I (as a controller) can attest that on a day-to-day basis, they generally aren't even known about, let alone used "as published" if they aren't on a STAR chart (where the overwhelming majority of holding is done in my experience). But since some fixes that aren't on arrivals definintely have flight-checked holding patterns that appear in the Low IFR Charts, it's best to keep the code alive just in case we need those holds. Creates some confusion about where to put holds (STAR vs directly in the airport `holds` key), but that can be sorted out with a little bit of thinking... 🤔 

https://www.boldmethod.com/learn-to-fly/maneuvers/what-you-should-know-about-holding/
https://nfdc.faa.gov/webContent/content8260/VA_EWOOD_REV5.pdf

image

oobayly: Moving to the NavigationLibrary makes perfect sense and takes very little work, as it's a static reference.

It still means that those holds still have to be stored separately from those in the STARs. Put it out to a committee? :grin:

erikquinn: 👍

oobayly:
Hopefully fb4963c doesn't overstep the remit of this PR, but seeing as the speed restriction parsing was being duplicated I felt it was valid:

* As `HoldModel` uses a similar speed restriction parser to `WaypointModel`, move the parser logic into `navigationUtilities`. These parsers return a tuple containing the speed/altitude and limit char:
`"S250-" -> [250, '-']`
* Update `WaypointModel` to use `parseAltitudeRestriction` and `parseSpeedRestriction`
* `WaypointModel` now throws if given an value that is invalid
eg. S250@, S1000-
* Update tests

**Note**: The parses also do some basic bounds validation
* `parseSpeedRestriction` rejects speed values less than 100kts, and more than 999kts
* `parseAltitudeRestriction` rejects altitudes above 99,999 ft

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -308,20 +308,31 @@ export default class AircraftCommander {
             return [false, `unable to hold at unknown fix ${fixName}`];
```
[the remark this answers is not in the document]
oobayly: Ah, I never considered the message being returned. Thanks.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -754,17 +754,23 @@ export default class Pilot {
...
+            return [success, responseValue];
         }
 
+        // When successful, the responseValue contains the actual holdParameters used by the
+        // `WaypointModel`. This means that we can send partial holdParameters, to patch
+        // the `WaypointModel`s _holdParameters property
+        holdParameters = responseValue;
+
+        const radialText = heading_to_string(holdParameters.inboundHeading + Math.PI);
+        const cardinalDirectionFromFix = getRadioCardinalDirectionNameForHeading(holdParameters.inboundHeading);
```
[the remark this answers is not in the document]
oobayly: How on earth did I miss that? :man_facepalming:

`Fms.activateHoldForWaypointName` didn't have a `fallbackInboundHeading` parameter so it wasn't passing it on to the `RouteModel`

PR #1445, opened 2019-09-22
bugfix/23 - Ensure focus is on the command input

### LINKED ISSUE #23: Command bar cursor not at end of line in MS Edge
The cursor should be at the end of the line, but in Microsoft Edge, it is at the far left instead.

Originally reported in zlsa/atc#489.

Also, per duplicate issue #1438, clicking aircraft will sometimes not put the command bar in focus, meaning the first keystroke is not shown in the command bar.

oobayly:
Resolves #23 

The purpose of this pull request is to ensure that the command input textbox has focus after an aircraft is selected.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -307,6 +307,9 @@ export default class InputController {
         this.input.callsign = aircraftModel.callsign;
         this.input.command = '';
         this.$commandInput.val(`${aircraftModel.callsign} `);
+        if (!this.$commandInput.is(':focus')) {
+            this.$commandInput.focus();
+        }
         this._eventBus.trigger(EVENT.SELECT_STRIP_VIEW_FROM_DATA_BLOCK, aircraftModel);
     };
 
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/InputController.js
```diff
@@ -307,6 +307,9 @@ export default class InputController {
         this.input.callsign = aircraftModel.callsign;
         this.input.command = '';
         this.$commandInput.val(`${aircraftModel.callsign} `);
+        if (!this.$commandInput.is(':focus')) {
```
n8rzz: There should be an empty line above this `if`, and also an empty line after the closing `}`

oobayly: Indeed it should.

oobayly:
I could also go with comparing it to document.activeElement, but it seems a bit clunky combining both vanilla js and jQuery.

n8rzz:
> I could also go with comparing it to document.activeElement, but it seems a bit clunky combining both vanilla js and jQuery.

nah, we already have jquery here we may as well keep to the native solutions

PR #1447, opened 2019-09-22
bugfix/1446 - Normalise aircraft heading in "say heading" read-back

### LINKED ISSUE #1446: Normalise aircraft heading in "say heading" read-back
Browser:
All

Airport:
All (Use EGGW for an example)

Immediate

#### Issue Description
When issuing the "say heading" command, the heading isn't normalised and may be given as a negative number.

#### Steps to reproduce
When an aircraft on the ABBOT2B STAR enters the airspace, request a heading read-back. The heading will be given as a negative number.

oobayly:
Resolves #1446 

The purpose of this pull request is to ensure that heading returned by the `runSayHeading` and `runSayAssignedHeading` read-backs are normalised to 0° and 360°

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - [#1440](https://github.com/openscope/openscope/issues/1440) - Add missing "b738"-fleet to TUI Airways
 - [#23](https://github.com/openscope/openscope/issues/23) - Ensure focus remains on the text input box in MS Edge
 - [#1448](https://github.com/openscope/openscope/issues/1448) - Lower spawn altitudes of KSFO arrivals so can comply with STAR restrictions
+- [#1446](https://github.com/openscope/openscope/issues/1446) - Ensure `sh`/`sah` commands return headings within 001-360
 
 
 ### Enhancements & Refactors
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -12,7 +12,6 @@ import {
     FLIGHT_PHASE
 } from '../constants/aircraftConstants';
 import { EVENT } from '../constants/eventNames';
-import { degrees_normalize } from '../math/circle';
 import { round } from '../math/core';
 import { AIRCRAFT_COMMAND_MAP } from '../parsers/aircraftCommandParser/aircraftCommandMap';
 import { speech_say } from '../speech';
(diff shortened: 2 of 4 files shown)


oobayly:
Yeah, I had a look to see if there were test that needed modifying, but there are no AircraftCommander tests (to get some hints from), nor any AircraftController (upon which the AircraftCommander relies) tests.

I was going to say that the tests for degrees_normalize _should_ catch errors, but I've just seen that there are none of those either (proving the assumptions adage!)

n8rzz:
> Yeah, I had a look to see if there were test that needed modifying, but there are no AircraftCommander tests (to get some hints from), nor any AircraftController (upon which the AircraftCommander relies) tests.
> 
> I was going to say that the tests for degrees_normalize _should_ catch errors, but I've just seen that there are none of those either (proving the assumptions adage!)

I was afraid of that.  We use `ava` for testing, which is node based and *does not have a DOM*.  At least, not without a bunch of work.  Because we don't have a DOM within our test env, pretty much anything that imports jquery is not testable.  So that's likely the reason this isn't tested at the moment.

oobayly:
Um, I think we've got our PRs mixed up. The input focus & jQuery one is #1445 

I've cobbed together some test for this one, but am concerned that a badly written test is worse than none at all. `runSayAssignedHeading` relies up the aircaft mcp being set which is currently beyond my comprehension.

```js
import ava from 'ava';
import AircraftCommander from '../../src/assets/scripts/client/aircraft/AircraftCommander';
import AircraftModel from '../../src/assets/scripts/client/aircraft/AircraftModel';
import {
    AIRCRAFT_MOCK_WITH_HEADING,
    AIRCRAFT_MOCK_WITH_HEADING_NEGATIVE,
    RUN_SAY_HEADING_RESULT
} from './_mocks/aircraftCommanderMocks';

ava('.runSayHeading() returns correct when heading is normalized', (t) => {
    const commander = new AircraftCommander();
    const aircraft = new AircraftModel(AIRCRAFT_MOCK_WITH_HEADING);
    const result = commander.runSayHeading(aircraft);

    t.deepEqual(result, RUN_SAY_HEADING_RESULT);
});

ava('.runSayHeading() returns correct when heading is negative', (t) => {
    const commander = new AircraftCommander();
    const aircraft = new AircraftModel(AIRCRAFT_MOCK_WITH_HEADING_NEGATIVE);
    const result = commander.runSayHeading(aircraft);

    t.deepEqual(result, RUN_SAY_HEADING_RESULT);
});
```

n8rzz:
Ha, I tend to take the other side of that coin; any test is better than no test.  Of course, any in this case would typically mean simple happy path test.

I think your example looks good for happy path, I wouldn't have any concerns adding that in.

oobayly:
Started to write my own getRadioHeading method, and then realised that if the "fly heading" command reads back correctly, then I should probably use that!

PR #1449, opened 2019-09-25
enhancement/1402 - Prevent assignment of reserved or nondiscrete beacon codes

### LINKED ISSUE #1402: Prevent assignment of reserved or nondiscrete beacon codes
Browser: Running the sim locally

Airport: LOWW (WIP Airport, but fully reliable, so that shouldn't be a problem)

Approximate play time before issue: Anytime 

#### Issue Description
When I was testing spawnpatterns at my WIP airport LOWW Vienna, an Austrian Airbus spawned with the squawk code 7600. I'm guessing 7500, 7600 and 7700 are blocked from appearing, but it doesn't seem to work properly

https://files.slack.com/files-pri/T1LML0QF2-FKQHNHK6Y/image.png

Important: No, I did not tell the aircraft to squawk that code.

Thank you.

oobayly:
Resolves #1402 

The purpose of this pull request is to fix the following:
1. Fix invalid or duplicate transponder codes being added into `_transponderCodesInUse` due to incorrect recursion
2. Fix issue with transponder codes not being compared correctly to the reserved list (string vs number)
3. Add region specific restricted transponder codes, eg. 1202 is reserved for VFR in the USA but not in the UK. ~~**Incomplete**~~

This slightly duplicates the work done in #1350 but as that's been orphaned it might we worth revisiting that later - implementing the Cid lookup as suggested by Erik.

### DIFF
--- a/src/assets/scripts/client/utilities/transponderUtilities.js
+++ b/src/assets/scripts/client/utilities/transponderUtilities.js
@@ -21,6 +21,8 @@ const MAX_TRANSPONDER_CODE = 4095;
  * The codes listed should still be assignable at the
  * controler's discretion
  *
+ * Source: https://en.wikipedia.org/wiki/List_of_transponder_codes
+ *
  * @property TRANSPONDER_CODES
  * @type {array<object>}
  * @final
@@ -48,7 +50,9 @@ const TRANSPONDER_CODES = [
         // Belgium
         prefix: /^eb/,
         reserved: [
-            /^00(4[1-6]|5[0-7])$/,
+            // Assigned for VFR traffic under Flight Information Services (BXL FIC)
+            // 0041–0057
+            /^00(4[1-7]|5[0-7])$/,
             // For testing stations
             '7777'
         ]
@@ -76,7 +80,11 @@ const TRANSPONDER_CODES = [
         prefix: /^eg/,
         reserved: [
             // Parachute drop
-            '0033'
+            '0033',
+            // Sudden military climb out from low-level operations
+            '7001',
+            // Aerobatic & displays
+            '7004'
         ]
     },
     {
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/utilities/transponderUtilities.js
```diff
@@ -0,0 +1,209 @@
...
+            // Parachute drop
+            '0033'
+        ]
+    },
+    {
+        // USA
+        prefix: /^k/,
+        reserved: [
+            // Military intercept
+            '0000',
```
erikquinn: FAA considers all transponder codes ending in `00` to be nondiscrete and does not assign them to specific aircraft. So `0100, 0200, 1800, 5600`, etc are all no good.

oobayly: This is checked for `isDiscreteTransponderCode`, so any non-discrete codes could be simply removed from the list of reserved items. In fact that check should be first anyway, so so it fails quickly on non-discrete codes before doing the more complex checks.

It would certainly simplify the list, but would mean they wouldn't be validated as reserved in `isReserved`. That in itself isn't a problem as the only place it's called is in `isDiscreteTransponderCode`.

src/assets/scripts/client/utilities/transponderUtilities.js
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * The highest decimal value allowed for a 4-digit
+ * octal transponder code
+ *
+ * @property MAX_TRANSPONDER_CODE
+ * @type {number}
+ * @final
+ */
+const MAX_TRANSPONDER_CODE = 4095;
```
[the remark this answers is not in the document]
oobayly: Ah, I can see that the choice in variable name wasn't clever! `MAX_TRANSPONDER_CODE_DECIMAL_VALUE` might be a better option as it's an upper limit and not a count.

I opted for the conversion from decimal to octal as it requires a single RNG call. I'll double check, but I'm pretty sure the upper limit for the function is inclusive, hence the 4095.

PR #1452, opened 2019-09-27
enhancement/1451 - Add ability to specify the radial for a hold

### LINKED ISSUE #1451: Add ability to specify the radial for a hold
As the title suggests, this would allow the player to specify the radial (inbound heading) that the aircraft should use when holding.

The `radial` option is purely optional, and defaults to the current behaviour if no radial is provided.

`WZZ36EX hold abbot 265`

Readback: WZZ36EX, hold east of ABBOT on the 265 radial, right turns, 1min legs

oobayly:
Resolves #1451.

The purpose of this pull request is to add ability to specify the radial for a hold. While we're hammering out the ~finer~ coarser points of #1431 I _accidentally_ implemented this as I needed a method of commanding a hold direction easily.

* Add the optional `radial` parameter (a 3-digit decimal number) to the hold command which sets the inboundHeading when executing a hold
* Use `WZZ36EX hold abbot 265`
With readback _WZZ36EX, hold east of ABBOT on the 265 radial, right turns, 1min legs_
* Updates tests to deal with the extra parameter
* Update documentation

Entering ABBOT hold from different directions
image

Both holding at ABBOT
image

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -12,6 +12,7 @@ import {
     FLIGHT_PHASE
 } from '../constants/aircraftConstants';
 import { EVENT } from '../constants/eventNames';
+import { radians_normalize } from '../math/circle';
 import { round } from '../math/core';
 import { AIRCRAFT_COMMAND_MAP } from '../parsers/aircraftCommandParser/aircraftCommandMap';
 import { speech_say } from '../speech';
@@ -300,19 +301,22 @@ export default class AircraftCommander {
      * @return {array} [success of operation, readback]
      */
     runHold(aircraft, data) {
-        const [turnDirection, legLength, fixName, inboundHeading] = data;
+        const [turnDirection, legLength, fixName, radial] = data;
         const fixModel = NavigationLibrary.findFixByName(fixName);
 
         if (!fixModel) {
             return [false, `unable to hold at unknown fix ${fixName}`];
         }
 
+        // Radial is given as the outbound course, so it needs to be inverted
+        const inboundHeading = radial === null ?
+            fixModel.positionModel.bearingFromPosition(aircraft.positionModel) :
+            radians_normalize(degreesToRadians(radial) + Math.PI);
+
         const holdParameters = {
             turnDirection,
             legLength,
-            inboundHeading: inboundHeading === null ?
-                fixModel.positionModel.bearingFromPosition(aircraft.positionModel) :
-                degreesToRadians(inboundHeading)
+            inboundHeading: inboundHeading
         };
 
         return aircraft.pilot.initiateHoldingPattern(fixName, holdParameters);
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -308,15 +308,17 @@ export default class AircraftCommander {
             return [false, `unable to hold at unknown fix ${fixName}`];
         }
 
+        let inboundHeading = fixModel.positionModel.bearingFromPosition(aircraft.positionModel);
+
         // Radial is given as the outbound course, so it needs to be inverted
-        const inboundHeading = radial === null ?
-            fixModel.positionModel.bearingFromPosition(aircraft.positionModel) :
-            radians_normalize(degreesToRadians(radial) + Math.PI);
+        if (radial !== null) {
+            inboundHeading = radians_normalize(degreesToRadians(radial) + Math.PI);
+        }
 
         const holdParameters = {
             turnDirection,
             legLength,
-            inboundHeading: inboundHeading
+            inboundHeading
         };
 
         return aircraft.pilot.initiateHoldingPattern(fixName, holdParameters);
--- a/src/assets/scripts/client/aircraft/AircraftCommander.js
+++ b/src/assets/scripts/client/aircraft/AircraftCommander.js
@@ -21,7 +21,7 @@ import {
     radio_heading,
     radio_altitude
 } from '../utilities/radioUtilities';
-import { heading_to_string, radiansToDegrees } from '../utilities/unitConverters';
+import { heading_to_string, radiansToDegrees, degreesToRadians } from '../utilities/unitConverters';
 
 /**
  *
@@ -300,9 +300,7 @@ export default class AircraftCommander {
      * @return {array} [success of operation, readback]
      */
     runHold(aircraft, data) {
-        const turnDirection = data[0];
-        const legLength = data[1];
-        const fixName = data[2];
+        const [turnDirection, legLength, fixName, inboundHeading] = data;
         const fixModel = NavigationLibrary.findFixByName(fixName);
 
         if (!fixModel) {
@@ -312,7 +310,9 @@ export default class AircraftCommander {
         const holdParameters = {
             turnDirection,
             legLength,
-            inboundHeading: fixModel.positionModel.bearingFromPosition(aircraft.positionModel)
+            inboundHeading: inboundHeading === null ?
+                fixModel.positionModel.bearingFromPosition(aircraft.positionModel) :
+                degreesToRadians(inboundHeading)
         };
 
         return aircraft.pilot.initiateHoldingPattern(fixName, holdParameters);
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -757,6 +757,7 @@ export default class Pilot {
      * @return {array} [success of operation, readback]
      */
     initiateHoldingPattern(fixName, holdParameters) {
+        const radial = heading_to_string(holdParameters.inboundHeading);
         const cardinalDirectionFromFix = getRadioCardinalDirectionNameForHeading(holdParameters.inboundHeading);
         const problematicResponse = this._fms.activateHoldForWaypointName(fixName, holdParameters);
 
@@ -765,11 +766,13 @@ export default class Pilot {
         }
 
         const holdParametersReadback = `${holdParameters.turnDirection} turns, ${holdParameters.legLength} legs`;
+        const radialReadbackLog = `on the ${radial} radial`;
+        const radialReadbackSay = `on the ${radio_heading(radial)} radial`;
 
         // force lower-case in verbal readback to get speech synthesis to pronounce the fix instead of spelling it
         return [true, {
-            log: `hold ${cardinalDirectionFromFix} of ${fixName.toUpperCase()}, ${holdParametersReadback}`,
-            say: `hold ${cardinalDirectionFromFix} of ${fixName.toLowerCase()}, ${holdParametersReadback}`
+            log: `hold ${cardinalDirectionFromFix} of ${fixName.toUpperCase()} ${radialReadbackLog}, ${holdParametersReadback}`,
+            say: `hold ${cardinalDirectionFromFix} of ${fixName.toLowerCase()} ${radialReadbackSay}, ${holdParametersReadback}`
         }];
     }
 
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -757,7 +757,7 @@ export default class Pilot {
      * @return {array} [success of operation, readback]
      */
     initiateHoldingPattern(fixName, holdParameters) {
-        const radial = heading_to_string(holdParameters.inboundHeading);
+        const radial = heading_to_string(holdParameters.inboundHeading + Math.PI);
         const cardinalDirectionFromFix = getRadioCardinalDirectionNameForHeading(holdParameters.inboundHeading);
         const problematicResponse = this._fms.activateHoldForWaypointName(fixName, holdParameters);
 
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -757,7 +757,7 @@ export default class Pilot {
      * @return {array} [success of operation, readback]
      */
     initiateHoldingPattern(fixName, holdParameters) {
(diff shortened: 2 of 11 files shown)


erikquinn:
As mentioned in the issue, we specify the holding direction by the bearing FROM the holding fix, not by the inbound heading.

So specifying 265 should result in a holding pattern to the west of ABBOT, along the ABBOT-R265, with an inbound heading of 085.

erikquinn:
Quick note-- please verify that when an aircraft is established in a holding pattern, clearing it direct to the fix it's holding over will clear the hold, and go direct to that fix to rejoin the rest of the route.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -300,19 +301,22 @@ export default class AircraftCommander {
...
         const fixModel = NavigationLibrary.findFixByName(fixName);
 
         if (!fixModel) {
             return [false, `unable to hold at unknown fix ${fixName}`];
         }
 
+        // Radial is given as the outbound course, so it needs to be inverted
+        const inboundHeading = radial === null ?
+            fixModel.positionModel.bearingFromPosition(aircraft.positionModel) :
+            radians_normalize(degreesToRadians(radial) + Math.PI);
```
erikquinn:
```
let inhoundHeading = fixModel.positionModel.bearingFromPosition(aircraft.positionModel);

if (radial) {
    inhoundHeading = radians_normalize(degreesToRadians(radial) + Math.PI);
}
```

I'm not saying this is any better or worse, your solution works fine, we just tend to avoid ternaries unless they're small and clearly advantageous.

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -757,6 +757,7 @@ export default class Pilot {
      * @return {array} [success of operation, readback]
      */
     initiateHoldingPattern(fixName, holdParameters) {
+        const radial = heading_to_string(holdParameters.inboundHeading + Math.PI);
         const cardinalDirectionFromFix = getRadioCardinalDirectionNameForHeading(holdParameters.inboundHeading);
```
erikquinn: I assume this rounds the value to the nearest integer? Don't want a response of `holding east of ABBOT on the 4.174926391036391846291739 radial, right turns, 10 mile legs` :laughing:

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -300,19 +301,22 @@ export default class AircraftCommander {
...
+
         const holdParameters = {
             turnDirection,
             legLength,
-            inboundHeading: fixModel.positionModel.bearingFromPosition(aircraft.positionModel)
+            inboundHeading: inboundHeading
```
erikquinn: Can actually just do `inboundHeading,` here, since the key and variable name are the same :wink:

src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -757,6 +757,7 @@ export default class Pilot {
      * @return {array} [success of operation, readback]
      */
     initiateHoldingPattern(fixName, holdParameters) {
+        const radial = heading_to_string(holdParameters.inboundHeading + Math.PI);
```
erikquinn: Can we call this `radialText` or something similar? I had to look a few times since I assumed this was simply the radial, when actually we're taking the radial value and making a string out of it. I'd name it `radialString` but then @n8rzz would slap me for using types in names :joy:

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandParserMessages.js
```diff
@@ -27,6 +27,7 @@ export const ERROR_MESSAGE = {
     THREE_ARG_LENGTH: `${INVALID_ARG_LENGTH}. Expected exactly three arguments`,
     ZERO_OR_ONE_ARG_LENGTH: `${INVALID_ARG_LENGTH}. Expected zero or one argument`,
     ZERO_TO_THREE_ARG_LENGTH: `${INVALID_ARG_LENGTH}. Expected zero to three arguments`,
+    ZERO_TO_FOUR_ARG_LENGTH: `${INVALID_ARG_LENGTH}. Expected zero to four arguments`,
```
erikquinn: Never woulda thought we'd need this lol :+1:

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
```diff
@@ -168,6 +168,23 @@ export const fixValidator = (args = []) => {
...
+/**
+ * Returns true if the value is exactly a 3 digit decimal number
+ *
+ * @function isValidCourseString
+ * @param value {string}
+ * @returns {boolean}
+ */
+export const isValidCourseString = (value) => {
+    // Can't rely on parseInt/convertStringToNumber as it'll parse 1min => 1
+    if (!/^[0-9]{3}$/.test(value)) {
```
erikquinn: You regex wizards blow my mind understanding this garbage...

That said, please move it to the REGEX const so the rest of us lemmings can reuse your wizardry.

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
```diff
@@ -168,6 +168,23 @@ export const fixValidator = (args = []) => {
...
+        return false;
+    }
+
+    const course = convertStringToNumber(value);
+    return course >= 0 && course <= 360;
```
erikquinn: Not sure if it matters in this instance, but `0` for a heading is never right to display to the user. If we're converting between radians and degrees we may end up here which is fine. Just want to verify/reiterate that only 001 through 360 should be what the user sees anywhere.

src/assets/scripts/client/aircraft/AircraftCommander.js
n8rzz: @erikquinn fun fact: adding `js` after the three bat ticks of a code block will give you syntax highlighting.  I've edited your comment to use that 😉

src/assets/scripts/client/aircraft/Pilot/Pilot.js
n8rzz: ☝️ maybe not slap, just say I was _'disappointed in you'_ 😉

src/assets/scripts/client/parsers/aircraftCommandParser/aircraftCommandParserMessages.js
n8rzz: me either.  glad it's proving to be flexible though

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
```diff
@@ -168,6 +168,23 @@ export const fixValidator = (args = []) => {
     }
 };
 
+/**
+ * Returns true if the value is exactly a 3 digit decimal number
```
n8rzz: I'd modify this slightly to: `Tests if value is exactly a 3 digit decimal number`

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
n8rzz: missing empty line above `return`

test/parsers/aircraftCommandParser/argumentParser.spec.js
```diff
@@ -98,53 +97,71 @@ ava('.headingParser() parses three digit heading as a generic heading', t => {
 
 ava('.findHoldCommandByType() returns a turnDirection when passed a variation of left or right', (t) => {
     const argsMock = ['dumba', 'l', '3nm'];
-    t.true(findHoldCommandByType('turnDirection', argsMock) === 'left');
+    t.is(findHoldCommandByType('turnDirection', argsMock), 'left');
```
n8rzz: 👍 on the `t.is`

src/assets/scripts/client/aircraft/AircraftCommander.js
oobayly: I first went with that, but opted for the ternary so I could declare it const. Plus it doesn't have to calculate bearingFromPosition unless needed.

Because you never know if you'll be requesting holds in a tight loop :grin:

src/assets/scripts/client/aircraft/Pilot/Pilot.js
oobayly: _"Readback correct, but plus or minus a degree will suffice"_

But yes, it does

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
oobayly: Yup, all the readbacks are normalised to 001 -> 360. I left this as allowing 000, as it's for validating user input, much in the same way `fh 000` is an acceptable command

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
oobayly: Good point

oobayly:
> Quick note-- please verify that when an aircraft is established in a holding pattern, clearing it direct to the fix it's holding over will clear the hold, and go direct to that fix to rejoin the rest of the route.

Yes, um, kind of. But under certain scenarios it doesn't. I wonder if some of the hold logic relies up the inboundHeading being the same (or similar) to the leg course. Holding at an unexpected side of the Fix seems to raise some interesting issues.

I foresee a lot more work with this one...

Definitely not ready for prime time :disappointed:

src/assets/scripts/client/aircraft/Pilot/Pilot.js
erikquinn: Everybody knows that's worse!

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
erikquinn: `fh 0` or `fh 000` is not an acceptable command, it should definitely assume that to be a typo and reject a zero-heading.

That said, if it's what it was already doing before these changes, you don't have to go change that if you don't want to-- only if it's a new behavior.

oobayly:
I found that the waypoint the aircaft flies to depends on the direction it's heading when the `proceed direct` command is issued. At first I thought it was a regression caused by this PR, but after testing on the develop branch I found the same issues.

A picture is better than a 1,000 words. These are all on EGGW - Luton, with aircraft on the LOREL2D STAR. BPK -> BKY -> BUSTA -> LOREL

This one works:
1. `RYR59MR HOLD BPK 265`
image
2. `RYR59MR PD BPK`
image

This one doesn't work:

1. `WZZ41B HOLD BPK 160`
image
2. I then wait until he's on the outbound leg of the hold and order `WZZ41B PD BPK` Instead of turning right to go to BPK, he turns left direct to BKY
image

This one works, even though at first glance it doesn't appear to:

1. Because I'm a stupid controller I order `RYR24FL HOLD BPK 010`. This means that RYR24FL's inbound leg to the fix will be the opposite direction that he original arrived from
image
2. After the 360 turn to enter the hold, and with almost 50% of the 1st circuit complete, he's starting to settle down. The next outbound leg is now correctly offset to the West of BPK
image
3. I wait for RYR24FL to get 50% through the next circuit.
image
4. Now, remembering that I'm an idiot (and should be fired immediately) I decide to order `RYR24FL PD BPK`
image

To the untrained eye (me), I **assumed** that the he wasn't _"going direct to that fix [BPK] to rejoin the rest of the route."_, but was going direct to BKY. In actual fact it's working perfectly well, and is dealing with the 180° turn correctly.

oobayly:
I'm putting this as a separate comment, these are some screenshots from dev.openscope.io:

1. `EZY4565  HOLD BPK`
image
2. `EZY4565 PD BPK` before he's established on the outbound leg
image
3. `WZZ1851 PD BPK` after he's established on the outbound leg
image

oobayly:
I'll try to explain what was going on. I misunderstood the proceed direct behaviour and thought at first it was a regression causes by this PR. What I found was happening was this:

1. Ordering `proceed direct` when on the inbound leg of the hold worked as expected. The aircraft would proceed to the hold fix and then rejoin the route. I was :smile: - it worked.
2. Ordering `proceed direct` while the aircraft was still on the outbound leg resulted in the aircraft bypassing the hold fix. I was :disappointed: - I thought I had broken something.

I then went on a wild goose chase trying to work out what I'd done, only to realise that this happened on the develop branch too. Only then did I realise that this was expected behaviour. It had nothing to do with the course, but with the proximity to the Fix.

**TLDR**; `proceed direct` works as expected:
* Flies to the hold fix first, if it's far enough away to make the turn
* Files to the next waypoint on the route if it's too close to make the turn to the hold fix.

erikquinn:
Let's also please add an optional `speed` parameter to the published holds. If it's not defined, then there is no speed restriction. Last addition to this PR, I promise. It's just the last step to being able to actually implement published holding patterns perfectly. After that, we'll be doing everything correctly except for holding pattern entries (which are hard!).

Aircraft should cross the holding fix at the specified speed.

image

src/assets/scripts/client/parsers/aircraftCommandParser/argumentValidators.js
oobayly: Ah, `fh 000` may not be an acceptable command, but it is currently accepted by the sim as a valid command.

I'm happy change it so 000 isn't valid for the `hold` command if you like. I imagine you'd prefer the runHeading behaviour is fixed outside of this PR.

oobayly:
> Let's also please add an optional speed parameter to the published holds.

Shouldn't this be in #1444 - Enhancement/1431 (Add support for hold parameters for fixes) ? It probably doesn't help you're being spammed with PRs from me.

erikquinn:
Did not realize there were two different PRs... Mind slightly blown, will look again when I get some free time

erikquinn:
Okay, yes belongs in the other. Will drop it there.

PR #1454, opened 2019-09-29
hotfix/1453 - Fix EINN maps property and update terrain

### LINKED ISSUE #1453: Fix EINN maps property
When #1435 and #1437 were merged into develop, then maps property for einn.json was reverted back to the old format and cannot be displayed by the sim.

This may relate to #1443 , although only EINN is affected here

* Update that `maps` property in einn.json to use the new array format
* While I'm in there, update the terrain to include water polygons ala #1442

oobayly:
Resolves #1453.

* Update that maps property in einn.json to use the new array format
* While I'm in there, update the terrain to include water polygons ala #1442


Fechulo:
PS: The water looks superb :clap:

Fechulo:
PPS: @erikquinn Any chance of this getting merged as a 'late hotfix' before the next update?

erikquinn:
tbh, the geography (except certain important elements thereof) are almost always in its own video map, rather than being combined with other markings.

When we have real video maps, I would (violently) oppose altering them. Since these are (I assume) manually created (by us or by vatsim people), I don't have an opinion, it's up to you.

We'll go next scheduled release, unless there's some other genuine bugfix we need to push out.

erikquinn:
Oh wait, it was actually broken though? Like it doesn't load correctly?

oobayly:
> However, I do have a concern. I noticed you removed the coastline from the video map. Presumably this is because the video map coastline and the real coastline didn't match up perfectly?

Exactly this - they didn't match up as they were from two different sources. I was happy to remove the coastline from the video map as I was the person who added it in the first place.

> There are many airports (American ones specifically) where it might be very hard (or just too time-consuming) to remove the coastline from the video map once a coastline is added to the terrain.

Yup, I've noticed this. The European maps don't tend to have the coastline on the videomap, so it's not such an issue. Editing the maps isn't such a chore as I've put together a QGIS plugin that provides importing-export `airport.json` files as well as terrain generation, but... \/

> When we have real video maps, I would (violently) oppose altering them.

Sure, this is why I've held off creating a PR to update the terrain files, as the mismatched lines can look a little dodgy, and I felt whole deleting of videomap lines would be highly frowned upon!

> Oh wait, it was actually broken though? Like it doesn't load correctly?

Unfortunately. It worked on develop, but for whatever reason when the changes were merged into master the file got reverted. I'm not sure if that was my fault, seeing as I was responsible for both PRs.

PR #1459, opened 2019-10-01
enhancement/1458 - handle distance based holds

### LINKED ISSUE #1458: Implement distance-based holds
The hold command currently supports multiple formats for the leg length (1min vs 4nm), but only the time-based leg length is handled.

A _quick and dirty_ resolution is to calculate the leg length duration by estimating the time to complete a hold leg by dividing the leg length by then aircraft's ground speed.

oobayly:
Resolves #1458.

The purpose of this pull request is to implement holds where the leg length is a distance.

As the sim uses a timer to determine when the return leg of a hold should be initiated, this estimates the time based on the length of the leg (as specified by the player) and the aircraft's current ground speed.

Edit: (Sorry Erik, but this is another hold-based PR. I can't help myself)

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -1989,16 +1989,6 @@ export default class AircraftModel {
         const isPastFix = offset[1] < 1 && offset[2] < 2;
         const isTimerSet = holdParameters.timer !== INVALID_NUMBER;
         const isTimerExpired = isTimerSet && gameTime > holdParameters.timer;
-        let holdLegDurationInSeconds;
-
-        if (legLength.indexOf('min') !== -1) {
-            const holdLegDurationInMinutes = legLength.replace('min', '');
-            holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
-        } else {
-            // Leg is a distance, use the ground speed to determine the duration
-            const holdLegDistance = legLength.replace('nm', '');
-            holdLegDurationInSeconds = (holdLegDistance / this.groundSpeed) * TIME.ONE_HOUR_IN_SECONDS;
-        }
 
         if (isPastFix && !this._isEstablishedOnHoldingPattern) {
             this._isEstablishedOnHoldingPattern = true;
@@ -2011,6 +2001,17 @@ export default class AircraftModel {
         let nextTargetHeading = outboundHeading;
 
         if (abs(groundTrack - outboundHeading) < PERFORMANCE.MAXIMUM_ANGLE_CONSIDERED_ESTABLISHED_ON_HOLD_COURSE && !isTimerSet) {
+            let holdLegDurationInSeconds;
+
+            if (legLength.indexOf('min') !== -1) {
+                const holdLegDurationInMinutes = legLength.replace('min', '');
+                holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
+            } else {
+                // Leg is a distance, use the ground speed to determine the duration
+                const holdLegDistance = legLength.replace('nm', '');
+                holdLegDurationInSeconds = (holdLegDistance / this.groundSpeed) * TIME.ONE_HOUR_IN_SECONDS;
+            }
+
             currentWaypoint.setHoldTimer(gameTime + holdLegDurationInSeconds);
         }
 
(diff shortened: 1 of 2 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1981,17 +1981,27 @@ export default class AircraftModel {
...
         const offset = getOffset(this, waypointRelativePosition, inboundHeading);
-        const holdLegDurationInMinutes = holdParameters.legLength.replace('min', '');
-        const holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
         const gameTime = TimeKeeper.accumulatedDeltaTime;
         const isPastFix = offset[1] < 1 && offset[2] < 2;
         const isTimerSet = holdParameters.timer !== INVALID_NUMBER;
         const isTimerExpired = isTimerSet && gameTime > holdParameters.timer;
 
+        let holdLegDurationInSeconds;
+        if (legLength.indexOf('min') !== -1) {
```
n8rzz: * `let` assignment should immediately follow `const` without an empty line
* missing empty line above `if`

oobayly: Every time. You'd think is learn...

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1981,16 +1981,24 @@ export default class AircraftModel {
...
         const isTimerExpired = isTimerSet && gameTime > holdParameters.timer;
+        let holdLegDurationInSeconds;
+
+        if (legLength.indexOf('min') !== -1) {
+            const holdLegDurationInMinutes = legLength.replace('min', '');
+            holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
+        } else {
+            // Leg is a distance, use the ground speed to determine the duration
+            const holdLegDistance = legLength.replace('nm', '');
+            holdLegDurationInSeconds = (holdLegDistance / this.groundSpeed) * TIME.ONE_HOUR_IN_SECONDS;
```
erikquinn: What if some bozo enters `8000m` for the leg length? Do we validate that somewhere to be either `min` or `nm`?

erikquinn: We want the groundspeed when the aircraft is established on the outbound leg, or else the distance will be wrong. Is this when this is calculated?

oobayly: > What if some bozo enters `8000m` for the leg length? Do we validate that somewhere to be either `min` or `nm`?

Good shout, but yes, that's validated for in the `argumentParser.isLegLengthArg` function

> We want the groundspeed when the aircraft is established on the outbound leg, or else the distance will be wrong. Is this when this is calculated?

Good question. `_calculateTargetedHeadingHold` appears to be called every tick. To verify this I watched the value for `this.groundSpeed`, the `holdLegDurationInSeconds` is calculated for an up-to-date GS value. So even if I order the hold well before the fix is reached, it will use the GS when the a/c is `MAXIMUM_ANGLE_CONSIDERED_ESTABLISHED_ON_HOLD_COURSE`.

For example if the aircraft is long before the fix (cruising at 250KIAS) and I order 165KIAS (about 180KTAS) and a leg length of 3NM (aiming for 60s leg length). The `holdLegDurationInSeconds` is calculated to be about 62s.

In fact, to make thiings more efficient, `holdLegDurationInSeconds` only needs to be calculated if `currentWaypoint.setHoldTimer` needs to be called (as that's the only place it's used).

``` js
if (abs(groundTrack - outboundHeading) < PERFORMANCE.MAXIMUM_ANGLE_CONSIDERED_ESTABLISHED_ON_HOLD_COURSE && !isTimerSet) {
    let holdLegDurationInSeconds;

    if (legLength.indexOf('min') !== -1) {
        const holdLegDurationInMinutes = legLength.replace('min', '');
        holdLegDurationInSeconds = holdLegDurationInMinutes * TIME.ONE_MINUTE_IN_SECONDS;
    } else {
        // Leg is a distance, use the ground speed to determine the duration
        const holdLegDistance = legLength.replace('nm', '');
        holdLegDurationInSeconds = (holdLegDistance / this.groundSpeed) * TIME.ONE_HOUR_IN_SECONDS;
    }

    console.log(isPastFix, this._isEstablishedOnHoldingPattern, this.groundSpeed, holdLegDurationInSeconds);
    currentWaypoint.setHoldTimer(gameTime + holdLegDurationInSeconds);
}
```

oobayly: Both holding at BPK @ 165 KIAS. One with a leg length of 3NM, the other with a leg length of 1min
image

PR #1462, opened 2019-10-14
enhancement/1021 - add dialog box with stats (game events)

### LINKED ISSUE #1021: Add stats
Some simple stats would be helpful, for example a counter with arrivals/departures handled, operational errors, etc. Most of this is already in place with events, we would just need to provide a way for users to see this data in a non-ugly way.

Related: #1020

grandsbor:
Resolves #1021.
Blocked by #1191.

Show to the user event counts that affect score.


grandsbor:
What is left to be done except for technical code review:
- create icon (currently copied settings icon, will accept any help/ideas)
- test in various browsers (currently only Ya.Browser (Chrome.based) on Linux)

oobayly:
Looks good, there've been more than a few times that I've looked at the score and seen a negative number, but not known what I did wrong, so this'll be useful...

I did notice a couple of things:
* The dialog doesn't close when the escape key is pressed, unlike the other dialogs (apart from the Change Log dialog)
* I couldn't get it do display separation losses, but that could be down to the game engine, not your work.

erikquinn:
Feel free to finish work on this, however due to the increasing number of buttons on the command bar, we will have to address #1191 before this can be merged.

Marking as blocked by #1191.

erikquinn:
Some tricky conflicts, and a more recent PR added functionality to show the history of all score changes. Given that the scoring will be removed anyway, will just be closing this PR and the corresponding issue.

PR #1464, opened 2019-10-17
hotfix/1463 - Fix travis build notifications

### LINKED ISSUE #1463: Fix travis build notifications

erikquinn:
Resolves #1463.
Blocks #1454 (a high priority hotfix).

Fix travis build notifications.

---

We have been migrated to the new travis-ci.com from the older travis-ci.org, and some things work differently under the new one. For one, Travis is now a GitHub App installed for the org, rather than using webhooks (or whatever other stuff I know nothing about that the legacy system used). Since making the switch, new pushes to all branches have only been running lint-diff, and not even the build/tests themselves 😬.

It's worth noting that, due to the way travis differentiates between a "PR Build" and a "Branch Build", it's apparently not currently possible to have both:
- slack notifications display the branch name that a build was run on, AND...
    - PR Build: `%{branch}` is the _target_ branch, eg `develop` or `master`
    - Branch Build: `%{branch}` is the branch the build is run on, eg `feature/123`
- ...have all builds on a given PR run coveralls or lint-diff on the _entire_ diff range of that PR
    - PR Build: $TRAVIS_COMMIT_RANGE includes all commits the PR has that the target branch does not have
    - Branch Build: $TRAVIS_COMMIT_RANGE includes only the commits _since the commit for which travis ran a build_ (so commit, push, commit, push, the first build may fail some tests, but the next build will pass since it only tests the second commit and not both)

As such, I elected to give up the former to keep the latter. See the below references on this:
- https://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/
- https://github.com/gajus/gitdown/issues/36

Example of the notification for the most recent push of this PR:
https://openscopeatc.slack.com/archives/C3L3A2MD0/p1571419579006700

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,6 @@
 sudo: false
 language: node_js
-node_js:
-- '11'
-
-# must specify this for allow_failures to work on env key
-# https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail
-env:
-
+node_js: '11.3'
 git:
   depth: 3
 cache:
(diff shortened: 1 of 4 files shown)


erikquinn:
image

PR #1465, opened 2019-10-18
feature/1368 - very simple voice control

### LINKED ISSUE #1368: Voice Control
Add an option "Control with voice" which allows the user to speak rather than type the commands. Eg. saying "Juliet Bravo Uniform One Three Niner, cleared ILS runway Two Three Left approach" instead of typing "JBU139 ils 23l"

manuel3108:
Resolves #1368 

provide a working and stable voice control at some point.

# current status
## WIP
By any means, this is not working great. But I think it's worth it showing the progress I have done in the last week.

## how it works
It's quite similar to the aircraft and airlines stuff, here is a small [format preview](https://github.com/openscope/openscope/blob/feature/1368/documentation/voice-recognition-format.md)
During the build process all files located in [assets/voice_recognition](https://github.com/openscope/openscope/tree/feature/1368/assets/voice_recognition) are assembled to one json and analyzed during runtime.
Every time a *.json is added and you build your game, the new commands will get available. So don't hesitate to add some new command's and share them witch everyone.

## usage example
- select an aircraft (so that the callsign is in the command field). **Trust me, don't try to get the callsign with voice recognition, you wont have any luck :D**
- click the mic icon in the icon bar
- (allow the page to listen to your mic (only on first time))
- say something like `turn right to 180` or `descent to 100` or `reduce speed to 200` (**but say `one-hundred-eighty` and not `one eight zero`**)
- ideally the command is understood and translated into aircraft-commands, and you only need to press enter

# compatibility
Apparently only in Chrome, [didn't worked in Firefox for me](https://hacks.mozilla.org/2016/01/firefox-and-the-web-speech-api/)

# tbd
- [ ] icon (only ripped atm)
- [ ] more commands
- [ ] aircraft callsign

propably more.... 

Please share your experiences and ask me if its not working for you


Hawaiicoder:
What you said about having to say "one-hundered-eighty"... It's fine, but I think I speak for most people in saying that it would be nice to say "one eight zero"... Just a thought.

PR #1466, opened 2019-10-21
feature/45 - Implement *T tool

### LINKED ISSUE #45: Add `*T` measuring
Using two clicks, the heading and distance between two points should be calculable and displayed to the user to aid in vectoring. In the STARS system, this is accessed via [*T (click) (click)], and in ERAM it's the "RNG BRNG" button, and click click. Need to figure out which to emulate or how else to most easily expose this functionality to the user.

Edit: this should include the ability to snap to aircraft or to a static position when no aircraft is within a specified distance. Includes aircraft to aircraft, aircraft to point, point to aircraft, and point to point.

Originally reported in zlsa/atc#740.

oobayly:
Resolves #45

The purpose of this pull request is to implement a *T (or equivalent) tool to measure distances, bearings on the scope.

Currently, it's initialted by holding down the Shift, although a scope command or button could be used to do the same.
* The tool will snap to the nearest aircraft or fix, or if none is found nearby it defaults to the cursor position
* Multiple ponts are supported, and can be an Aircraft, Fix, or Arbitrary point, and any combination of the above
* The path is updated as the mouse is moved to provide immediate vector information
* If the first point is an aircraft, the current ground speed will be used to estimate timings and turn radii

Isssues:
* There is currently no option to disable snapping. A 2nd keyboard modifier could be used.
* When drawing _"hairpin"_ paths, the tool will struggle to draw accurate arcs as the fillet (turn) radius may be too large.
* Tests are needed, as usual...

I've explain as much as possible in the comments, but I'm sure there'll be plenty of WFTs!

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -274,6 +274,24 @@ export default class InputController {
     }
 
     /**
+     * Resets the measuring tool, clearing existing paths
+     *
+     * @for InputController
+     * @method _resetMeasuring
+     * @private
+     */
+    _resetMeasuring() {
+        const { hasPaths } = MeasureTool;
+
+        MeasureTool.reset();
+
+        // Mark for shallow render so the feedback is immediate
+        if (hasPaths) {
+            this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+        }
+    }
+
+    /**
      * Starts the measuring tool
      *
      * @for InputController
@@ -281,7 +299,7 @@ export default class InputController {
      * @private
      */
     _startMeasuring() {
-        MeasureTool.isMeasuring = true;
+        MeasureTool.startNewPath();
     }
 
     /**
@@ -292,10 +310,7 @@ export default class InputController {
      * @private
      */
     _stopMeasuring() {
-        MeasureTool.reset();
-
-        // Mark for shallow render so the feedback is immediate
-        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+        MeasureTool.endPath();
     }
 
     /**
@@ -319,7 +334,7 @@ export default class InputController {
      * @param event {jquery Event}
      */
     _onMouseClickAndDrag(event) {
-        if (MeasureTool.isMeasuring && MeasureTool.hasStarted) {
+        if (MeasureTool.hasStarted) {
             this._addMeasurePoint(event, true);
 
             return this;
@@ -585,6 +600,9 @@ export default class InputController {
                 break;
             case KEY_CODES.ESCAPE:
             case LEGACY_KEY_CODES.ESCAPE:
+                // TODO: Probably should have its own cancel button
+                this._resetMeasuring();
+
                 UiController.closeAllDialogs();
 
                 const hasCallsign = _includes(currentCommandInputValue, this.input.callsign);
@@ -989,9 +1007,11 @@ export default class InputController {
     _onRightMousePress(event) {
         if (MeasureTool.isMeasuring) {
             this._removeLastMeasurePoint();
-        } else {
-            this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
+
+            return;
         }
+
+        this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
     }
 
     /**
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -10,6 +10,8 @@ import UiController from './ui/UiController';
 import AircraftCommandParser from './parsers/aircraftCommandParser/AircraftCommandParser';
 import ScopeCommandModel from './parsers/scopeCommandParser/ScopeCommandModel';
 import EventTracker from './EventTracker';
+import MeasureTool from './measurement/MeasureTool';
+import FixCollection from './navigationLibrary/FixCollection';
 import { clamp } from './math/core';
 import { EVENT } from './constants/eventNames';
 import { GAME_OPTION_NAMES } from './constants/gameOptionConstants';
@@ -61,6 +63,7 @@ export default class InputController {
         this._mouseDelta = [0, 0];
         this._mouseDownScreenPosition = [0, 0];
         this.input.isMouseDown = false;
+        this.input.measureTool = new MeasureTool();
         this.commandBarContext = COMMAND_CONTEXT.AIRCRAFT;
 
         this._init();
@@ -85,6 +88,7 @@ export default class InputController {
      */
     setupHandlers() {
         this.onKeydownHandler = this._onKeydown.bind(this);
+        this.onKeyupHandler = this._onKeyup.bind(this);
         this.onCommandInputChangeHandler = this._onCommandInputChange.bind(this);
         this.onMouseScrollHandler = this._onMouseScroll.bind(this);
         this.onMouseClickAndDragHandler = this._onMouseClickAndDrag.bind(this);
@@ -102,6 +106,7 @@ export default class InputController {
      */
     enable() {
         this.$window.on('keydown', this.onKeydownHandler);
+        this.$window.on('keyup', this.onKeyupHandler);
         this.$commandInput.on('input', this.onCommandInputChangeHandler);
         // TODO: these are non-standard events and will be deprecated soon. this should be moved
         // over to the `wheel` event. This should also be moved over to `.on()` instead of `.bind()`
@@ -127,6 +132,7 @@ export default class InputController {
      */
     disable() {
         this.$window.off('keydown', this.onKeydownHandler);
+        this.$window.off('keyup', this.onKeyupHandler);
         this.$commandInput.off('input', this.onCommandInputChangeHandler);
         // uncomment only after `.on()` for this event has been implemented.
         // this.$commandInput.off('DOMMouseScroll mousewheel', this.onMouseScrollHandler);
@@ -161,6 +167,7 @@ export default class InputController {
         this._mouseDelta = [0, 0];
         this._mouseDownScreenPosition = [0, 0];
         this.input.isMouseDown = false;
+        this.input.measureTool = null;
 
         return this;
     }
@@ -207,6 +214,91 @@ export default class InputController {
     }
 
     /**
+     * Adds a point to the measuring tool
+     *
+     * @for InputController
+     * @method _addMeasurePoint
+     * @param event {jquery Event}
+     * @param doReplaceLast {boolean} Indicates whether this will replace the last point
+     * @private
+     */
+    _addMeasurePoint(event, doReplaceLast = false) {
+        const currentMousePosition = CanvasStageModel.translateMousePositionToCanvasPosition(
+            event.pageX, event.pageY
+        );
+        const { x, y } = currentMousePosition;
+        const { measureTool } = this.input;
+        const [aircraftModel, distanceFromAircraft] = this._findClosestAircraftAndDistanceToMousePosition(x, y);
+        const [fixModel, distanceFromFix] = this._findClosestFixAndDistanceToMousePosition(x, y);
+        let distance;
+        let modelToUse;
+
+        // Which model is closest
+        if (distanceFromFix < distanceFromAircraft) {
+            distance = distanceFromFix;
+            modelToUse = fixModel;
+        } else {
+            distance = distanceFromAircraft;
+            modelToUse = aircraftModel;
+        }
+
+        if (distance > CanvasStageModel.translatePixelsToKilometers(20)) {
+            // Nothing's near the cursor, so we'll use the cursor point
+            modelToUse = this._translatePointToKilometers(x, y);
+        }
+
+        if (!measureTool.hasStarted) {
+            measureTool.startMeasuring(modelToUse);
+        } else if (doReplaceLast) {
+            measureTool.updateLastPoint(modelToUse);
+        } else {
+            measureTool.addPoint(modelToUse);
+        }
+
+        // Mark for shallow render so the draw motion is smooth
+        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+    }
+
+    /**
+     * Removes the last point in the measuring tool
+     *
+     * @for InputController
+     * @method _removeLastMeasurePoint
+     * @private
+     */
+    _removeLastMeasurePoint() {
+        this.input.measureTool.removeLastPoint();
+
+        // Mark for shallow render so the feedback is immediate
+        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+    }
+
+    /**
+     * Starts the measuring tool
+     *
+     * @for InputController
+     * @method _startMeasuring
+     * @private
+     */
+    _startMeasuring() {
+        this.input.measureTool.isMeasuring = true;
+    }
+
+    /**
+     * Stops the measuring tool
+     *
+     * @for InputController
+     * @method _stopMeasuring
+     * @private
+     */
+    _stopMeasuring() {
+        this.input.measureTool.reset();
+
+        // Mark for shallow render so the feedback is immediate
+        this._eventBus.trigger(EVENT.MARK_SHALLOW_RENDER);
+    }
+
+    /**
      * @for InputController
      * @method _onMouseScroll
      * @param event {jquery Event}
@@ -227,6 +319,13 @@ export default class InputController {
      * @param event {jquery Event}
      */
     _onMouseClickAndDrag(event) {
+        const { measureTool } = this.input;
+
+        if (measureTool.isMeasuring && measureTool.hasStarted) {
+            this._addMeasurePoint(event, true);
+            return this;
+        }
+
         if (!this.input.isMouseDown) {
             return this;
         }
@@ -351,6 +450,11 @@ export default class InputController {
 
         // TODO: this swtich can be simplified, there is a lot of repetition here
         switch (code) {
+            case KEY_CODES.SHIFT_LEFT:
+            case KEY_CODES.SHIFT_RIGHT:
+                this._startMeasuring();
+
+                break;
             case KEY_CODES.BAT_TICK:
             case LEGACY_KEY_CODES.BAT_TICK:
                 this.$commandInput.val(`${currentCommandInputValue}\` `);
@@ -499,6 +603,31 @@ export default class InputController {
                 break;
             default:
                 this.$commandInput.focus();
+        }
+    }
+
+
+    /**
+     * @for InputController
+     * @method _onKeydown
+     * @param event {jquery Event}
+     * @private
+     */
+    _onKeyup(event) {
+        let { code } = event.originalEvent;
+
+        if (code === undefined) {
+            // fallback for legacy browsers like IE/Edge
+            code = event.originalEvent.keyCode;
+        }
+
+        switch (code) {
+            case KEY_CODES.SHIFT_LEFT:
+            case KEY_CODES.SHIFT_RIGHT:
+                this._stopMeasuring();
+
+                break;
+            default:
         }
     }
 
@@ -828,10 +957,27 @@ export default class InputController {
      * @private
      */
     _findClosestAircraftAndDistanceToMousePosition(x, y) {
-        return this._aircraftController.aircraft_get_nearest([
-            CanvasStageModel.translatePixelsToKilometers(x - CanvasStageModel._panX),
-            CanvasStageModel.translatePixelsToKilometers(y + CanvasStageModel._panY)
-        ]);
+        return this._aircraftController.aircraft_get_nearest(
+            this._translatePointToKilometers(x, y)
+        );
+    }
+
+    /**
+     * Facade for `FixCollection.getNearest`
+     *
+     * Accepts current mouse position in canvas coordinates x, y
+     *
+     * @for InputController
+     * @method _findClosestFixAndDistanceToMousePosition
+     * @param x {number}
+     * @param y {number}
+     * @returns [FixModel, number]
+     * @private
+     */
+    _findClosestFixAndDistanceToMousePosition(x, y) {
+        return FixCollection.getNearestFix(
+            this._translatePointToKilometers(x, y)
+        );
     }
 
     /**
@@ -842,7 +988,11 @@ export default class InputController {
      * @private
      */
     _onRightMousePress(event) {
-        this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
+        if (this.input.measureTool.isMeasuring) {
+            this._removeLastMeasurePoint();
+        } else {
+            this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
+        }
     }
 
     /**
@@ -858,6 +1008,12 @@ export default class InputController {
      * @private
      */
     _onLeftMouseButtonPress(event) {
+        if (this.input.measureTool.isMeasuring) {
+            this._addMeasurePoint(event);
+
+            return;
+        }
+
         const currentMousePosition = CanvasStageModel.translateMousePositionToCanvasPosition(event.pageX, event.pageY);
         const [aircraftModel, distanceFromPosition] = this._findClosestAircraftAndDistanceToMousePosition(
             currentMousePosition.x,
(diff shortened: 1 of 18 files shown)


erikquinn:
> If the first point is an aircraft, the current ground speed will be used to estimate timings and turn radii

Actually, we should actively _not_ include any turn rounding. Straight lines from point to point to point only. If someone can provide a reliable reference showing that some places do round these lines, we can consider it as an option, though I may be inclined to fight it. :wink:

So for now, straight lines are perfect, which makes the work notably less difficult in terms of drawing.

oobayly:
> If someone can provide a reliable reference showing that some places do round these lines

The reason I went for arcing the lines was that I **vaguelly** remember seeing that in a VatSim ATC video, so I **assumed** that that was the way to go. Plus, it was a challenge to see if I could get it to work :grin:

To be fair, the all (but the first) arcs are cosmetic, as the distances are are calculated on vertex-to-vertex, but the errors would usually be fairly small relative to the length of the path. **I'll disable them for the time being**.

However, I'd argue that the initial turn could be a valuable feature as it calculates the turn to be made to fly to the point, and therefore the actual time to reach the point. Even if can be turned on in via the settings - just a thought...

I also noticed the build failed because of the _"dreaded prop"_, which actually raises something I'd considered, which was to make the `MeasureTool` a static object as it should never be used more than once.

image

src/assets/scripts/client/InputController.js
```diff
@@ -206,6 +213,91 @@ export default class InputController {
...
+    /**
+     * Adds a point to the measuring tool
+     *
+     * @for InputController
+     * @method _addMeasurePoint
+     * @param event {jquery Event}
+     * @param doReplaceLast {boolean} Indicates whether this will replace the last point
+     * @private
+     */
+    _addMeasurePoint(event, doReplaceLast = false) {
```
n8rzz: `doReplaceLast` -> `shouldReplaceLastPoint`

src/assets/scripts/client/InputController.js
```diff
@@ -227,6 +319,13 @@ export default class InputController {
      * @param event {jquery Event}
      */
     _onMouseClickAndDrag(event) {
+        const { measureTool } = this.input;
+
+        if (measureTool.isMeasuring && measureTool.hasStarted) {
+            this._addMeasurePoint(event, true);
+            return this;
```
n8rzz: missing empty line above `return`

src/assets/scripts/client/InputController.js
```diff
@@ -502,6 +606,31 @@ export default class InputController {
...
+    /**
+     * @for InputController
+     * @method _onKeydown
+     * @param event {jquery Event}
+     * @private
+     */
+    _onKeyup(event) {
+        let { code } = event.originalEvent;
+
+        if (code === undefined) {
```
n8rzz: you should actually look for:
```js
if (typeof code === 'undefined') { }
```

or, if you want to use coercion:
```js
if (code == null) { }
```

src/assets/scripts/client/InputController.js
```diff
@@ -842,7 +988,11 @@ export default class InputController {
      * @private
      */
     _onRightMousePress(event) {
-        this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
+        if (this.input.measureTool.isMeasuring) {
```
n8rzz:
```js
if (!this.input.measureTool.isMeasuring) {
    this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);

    return;
}

this._removeLastMeasurePoint();
```

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1217,6 +1218,144 @@ export default class CanvasController {
...
+     * @param pathInfo {object} as returned by `MeasureTools.getPointsAndLabels()`
+     * @private
+     */
+    _drawMeasureToolLabels(cc, pathInfo) {
+        let leg = pathInfo.firstLeg.next;
+        const values = [];
+
+        // This way the points are translated only once
+        while (leg != null) {
+            const p = CanvasStageModel.translatePostionModelToPreciseCanvasPosition(leg.midPoint);
```
n8rzz: does `p` stand for `position` or `point`?

src/assets/scripts/client/measurement/MeasureLegModel.js
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * The turn radius (in km)
+     *
+     * @for MeasureLegModel
+     * @property _radius
+     * @type {number}
+     */
+    get radius() {
+        const response = this._radius;
```
n8rzz: what does `response` mean here?  feels like a strange word in relation to `radius` an measuring.

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /** Gets the aircraft at the beginning of the path
+     *
+     * @for MeasureTool
+     * @property aircraft
+     * @type {AircraftModel|null}
+     */
+    get aircraft() {
+        if (this.hasStarted && this._points[0] instanceof AircraftModel) {
```
n8rzz: please reverse this logic to return `null` early when we don't have what we need

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+     * Removes the last point from the path
+     *
+     * @for MeasureTool
+     * @method removeLastPoint
+     */
+    removeLastPoint() {
+        // The "last" point is actually the 2nd last item in _points
+        // This means the path will end at the cursor position when redrawn
+        const { length } = this._points;
+        if (length > 2) {
```
n8rzz: missing empty line above `if`

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+     *
+     * The startValue can be an object with a `#relativePosition` property,
+     * or a array of numbers that represent a `relativePosition`
+     *
+     * @for MeasureTool
+     * @method startMeasuring
+     * @param startValue {AircraftModel|FixModel|array<number>}
+     */
+    startMeasuring(startValue) {
+        this._points = [startValue];
```
n8rzz: is `_points` an empty array before this assignment?  is so, you should use `.push()` here.  this is actually mutating `_points` instead of updating

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+     *
+     * The value can be an object with a `#relativePosition` property,
+     * or a array of numbers that represent a `relativePosition`
+     *
+     * @for MeasureTool
+     * @method updateEndPoint
+     * @param value {array<number>}
+     */
+    updateLastPoint(value) {
+        if (this.hasLegs) {
```
n8rzz: reverse the logic to return early:
```js
if (!this.hasLegs) {
    this.addPoint(value);

    return;
}

this._points[this._points.length - 1] = value;
```

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Get the text label that displays the distance and
+     * optional duration and bearings
+     *
+     * @param distance {number}
+     * @param duration {number}
+     * @param bearing {number|null}
+     * @private
+     */
+    _getLabel(distance, duration, bearing = null) {
```
n8rzz: `_getLabel` -> `_buildLabel`

`get` has a very specific meaning and we aren't actually _*getting*_ anything here, we're calculating/building up.

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,363 @@
...
+            center,
+            exitPoint
+        };
+    }
+
+    /**
+     * Get the text label that displays the distance and
+     * optional duration and bearings
+     *
+     * @param distance {number}
```
n8rzz: what are the units of these numbers?

src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -127,13 +148,7 @@ class FixCollection extends BaseCollection {
...
      */
     findRealFixes() {
-        const realFixList = _map(this._items, (item) => {
-            if (item.name.indexOf('_') !== 0) {
-                return item;
-            }
-        });
-
-        return _compact(realFixList);
+        return this._items.filter((fix) => fix.isRealFix);
```
n8rzz: 😌

oobayly:
> This looks really, really good! Most of my feedback are nitpicks. The meat of this submission is really, really well put together!

Thanks, that's much appreciated.

> One thing I'd like to see a little more of is tests. Specifically, for the new Measure classes and/or anything thats new in a class that doesn't import jquery.

Definitely. I'd held off creating any as yet, as quite often a lot changes during the lifetime of a PR! For example, I've:
* In order to facilitate _"death to the prop"_, I've changed the `MeasureTool` to be a static object, although also I considered using the `EventBus` to pass information from the `InputController` to the `CanvasController`
* I've also added a _"Measure path style"_ setting that allows the user to control the render style.

src/assets/scripts/client/InputController.js
oobayly: Copy pasta from `_onKeyDown` :grin: I'll fix both in that case...

src/assets/scripts/client/InputController.js
oobayly: I know it's much of a muchness, but would it be better to put the `_removeLastMeasurePoint` call inside the if? That way if more tests are ever added, the default behaviour will be the original `_markMousePressed`

src/assets/scripts/client/canvas/CanvasController.js
oobayly: Yeah, they're control positions in on the Canvas. Fixed

src/assets/scripts/client/measurement/MeasureLegModel.js
oobayly: Don't judge me, but this is a side effect from my Basic days when I didn't like to name a local variable the same as the method it's declared in. :duck:
It sometimes leaks through...

src/assets/scripts/client/measurement/MeasureTool.js
oobayly: I don't know how you spot these, I always think I've done it.

src/assets/scripts/client/measurement/MeasureTool.js
oobayly: Same goes for `getPathInfo` and `_getInitialTurnParameters` then...

src/assets/scripts/client/measurement/MeasureTool.js
oobayly: Ugh, one of my aerodynamics lecturers would dock a mark from coursework every time we left out a unit. Can't believe I left that out!

src/assets/scripts/client/navigationLibrary/FixCollection.js
oobayly: I had to google that emoji to make sure it was a good thing...

src/assets/scripts/client/measurement/MeasureTool.js
oobayly: Yes it should always be empty, but it was my way of making sure it was (which is a ~good~ bad way of ensuring bugs never manifest themselves)...

src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -101,6 +100,28 @@ class FixCollection extends BaseCollection {
...
 
+    /**
+     * Returns the nearest fix to the specified position
+     *
+     * @for FixCollection
+     * @method getNearestFix
+     * @param position {StaticPositionModel}
+     * @param allowRnavFixes {boolean} A flag indicating whether RNAV fixes should be used
+     */
+    getNearestFix(position, allowRnavFixes = false) {
```
erikquinn: `allowRnavFixes` --> `hiddenFixes`

(RNAV fixes are totally game, just not the fake ones we create to make certain procedures work, so I don't want to confuse readers of the code when they see that we aren't including "RNAV fixes", when really we _are_ 😉)

src/assets/scripts/client/InputController.js
```diff
@@ -342,15 +435,20 @@ export default class InputController {
...
+        if (code == null) {
             // fallback for legacy browsers like IE/Edge
             code = event.originalEvent.keyCode;
         }
 
         // TODO: this swtich can be simplified, there is a lot of repetition here
         switch (code) {
+            case KEY_CODES.SHIFT_LEFT:
+            case KEY_CODES.SHIFT_RIGHT:
+                this._startMeasuring();
```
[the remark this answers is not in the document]
oobayly: Funny, I started off with Ctrl, but switched to Shift after re-reading the original issue.

src/assets/scripts/client/InputController.js
```diff
@@ -828,10 +951,27 @@ export default class InputController {
...
+     * @for InputController
+     * @method _findClosestFixAndDistanceToMousePosition
+     * @param x {number}
+     * @param y {number}
+     * @returns [FixModel, number]
+     * @private
+     */
+    _findClosestFixAndDistanceToMousePosition(x, y) {
+        return FixCollection.getNearestFix(
+            this._translatePointToKilometers(x, y)
```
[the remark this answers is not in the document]
oobayly: My bad, I though I was replicating the behaviour of `AircraftController.aircraft_get_nearest`, obviously I wasn't...

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1217,6 +1219,142 @@ export default class CanvasController {
...
+        if (initialTurn !== null) {
+            const {
+                isRHT, center, entryAngle, exitAngle, turnRadius
+            } = initialTurn;
+            const position = CanvasStageModel.translatePostionModelToPreciseCanvasPosition(center);
+            const radius = CanvasStageModel.translateKilometersToPixels(turnRadius);
+
+            // The angles calculated in the `MeasureTool` are magnetic, and have to be shifted CCW 90°
+            cc.arc(position.x, position.y, radius, entryAngle - Math.PI / 2, exitAngle - Math.PI / 2, !isRHT);
+        }
```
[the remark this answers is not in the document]
oobayly: Yeah, this isn't entirely obvious. The way `MeasureTool.buildPathInfo` works is that it sets the radius to 0 if the corner shouldn't be arced (either from the style, or if the corner is too sharp to fillet). This has the same affect as using a simple lineTo. The reason it's done like that is twofold:
* Its simplified the draw logic by not having two loops, at the possible small expense of using the `arcTo` function (although I'd like to think it uses `lineTo` internally if the `radius === 0`)
* A path can have a mixture of arcs and points

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -16,7 +16,8 @@ export const GAME_OPTION_NAMES = {
     SOFT_CEILING: 'softCeiling',
     DRAW_ILS_DISTANCE_SEPARATOR: 'drawIlsDistanceSeparator',
     MOUSE_CLICK_DRAG: 'mouseClickDrag',
-    RANGE_RINGS: 'rangeRings'
+    RANGE_RINGS: 'rangeRings',
+    MEASURE_TOOL_PATH: 'measureToolPath'
```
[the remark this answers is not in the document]
oobayly: I remember this, I couldn't work out how the ordering is done, becuase `DRAW_ILS_DISTANCE_SEPARATOR` comes after `SOFT_CEILING`, so I thought they were somehow grouped by usage (which in hindsight is a little daft!).

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,513 @@
...
+     */
+    removeLastPoint() {
+        this._throwIfNotMeasuring();
+
+        // The "last" point is actually the 2nd last item in _points
+        // This means the path will end at the cursor position when redrawn
+        const { length } = this._points;
+
+        if (length > 2) {
+            this._points.splice(length - 2, 1);
```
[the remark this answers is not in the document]
oobayly: Again, it's not very obvious. If I used `pop()`:
1. the last point would be removed as expected
2. but if the mouse isn't moved,  `updateLastPoint` would be not be called
3. meaning that the line will contine to end at the last corner, and not under the mouse
4. I found that behaviour a little clumsy.

With the current usage, this removes the last corner so the lines continues to end under the mouse, even if the mouse hasn't been moved. It just felt more intuitive that way, even though the code looks a tad strange.

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,513 @@
...
+     * @method _buildInitialTurnParameters
+     * @returns {object}
+     * @private
+     */
+    _buildInitialTurnParameters() {
+        if (!this._hasStyle(MEASURE_TOOL_STYLE.INITIAL_TURN)) {
+            return null;
+        }
+
+        const { aircraft } = this;
```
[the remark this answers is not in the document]
oobayly: Yup, the tool takes a `FixModel`, `AircraftModel` or `relativePosition` as any point. The `aircraft` property is simply an alias for the first point (providing it's an instance of `AircraftModel`).

This could possibly be seen as a feature. If the draw style is `Arc to next fix`, starting from a fix (or cursor position) and ending in an aircraft will result in a straight line. So you can still get "as the crow flies" distance.

src/assets/scripts/client/navigationLibrary/FixCollection.js
oobayly: Hi hadn't considered that. :+1:

src/assets/scripts/client/InputController.js
```diff
@@ -206,6 +211,88 @@ export default class InputController {
...
+     * @method _addMeasurePoint
+     * @param event {jquery Event}
+     * @param shouldReplaceLastPoint {boolean} Indicates whether this will replace the last point
+     * @private
+     */
+    _addMeasurePoint(event, shouldReplaceLastPoint = false) {
+        const currentMousePosition = CanvasStageModel.translateMousePositionToCanvasPosition(
+            event.pageX, event.pageY
+        );
+        const { x, y } = currentMousePosition;
```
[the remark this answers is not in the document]
oobayly: Sorry, I'm not following. Are you saying you want to move the `findClosest` logic to the MeasureTool?

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,513 @@
...
+import { TIME } from '../constants/globalConstants';
+
+/**
+ * Style enumeration of available render styles
+ */
+export const MEASURE_TOOL_STYLE = {
+    STRAIGHT: 0,
+    INITIAL_TURN: 1,
+    ARCED: 2
+};
```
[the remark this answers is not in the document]
oobayly: I can see what you mean, but my only concern is that we'll be using string like `'straight'` in various places. Using the constants means they're only defined one.

I'm happy to go to string constants instead of bitwise operators, but wouldn't it make sense to keep `MEASURE_TOOL_STYLE` constant and import it into `gameOptionsConsts`?

``` js
            {
                displayLabel: 'Straight lines only',
                value: MEASURE_TOOL_STYLE.STRAIGHT
            },
```

src/assets/scripts/client/constants/gameOptionConstants.js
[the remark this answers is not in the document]
erikquinn: Definitely!

src/assets/scripts/client/measurement/MeasureTool.js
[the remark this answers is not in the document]
erikquinn: Fair point. And yes, that's a perfect solution then.

src/assets/scripts/client/measurement/MeasureTool.js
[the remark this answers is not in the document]
erikquinn: Oh god, I'm sorry. You even had a comment to this effect and I didn't notice it, or the fact that were splicing the second-to-last element, not the last. Disregard everything I said and carry on.

src/assets/scripts/client/measurement/MeasureTool.js
```diff
@@ -0,0 +1,513 @@
...
+        /**
+         * The style of how the path is rendered
+         *
+         * @for MeasureTool
+         * @property style
+         * @type {number}
+         * @default MEASURE_TOOL_STYLE.STRAIGHT
+         * @private
+         */
+        this.style = MEASURE_TOOL_STYLE.STRAIGHT;
```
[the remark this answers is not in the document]
oobayly: Will do.

oobayly:
This was easier that could have been expected, possibly because some part of me thought this might be requested.

1. Use Ctrl to start a new path
2. Start clicking (use Shift to snap)
3. Release Ctrl to end that path
4. Go to # 1
5. Press Esc to clear the tool

erikquinn:
@oobayly Is this working properly? It may be just me because I'm on a laptop with a bad trackpad, but I am having maybe a 5% success rate getting the lines to come up using ctrl. I thought it wasn't working at all until I had it work once or twice, but was very difficult to reliably create them. Just trying to figure out if it's me, or if you can confirm that something is up. 🤔

oobayly:
Strange, I had done most of the development for this on my Linux lapop, and I've just tried the deployment on the same machine and it works as expected.

However I also tried it on my Windows PC & TV with a wireless keyboard and it's struggling. Attaching a keydown event handler shows in the event being repeatedly called even though the key isn't being released. Can I ask what OS and browser version you're using?

erikquinn:
Twas on Win10, Chrome 77.0.3865.120

Updated to 78.0.3904.87 (latest) just now, and same result.

Maybe it has to do with it being a modifier? Not sure why it's behave differently than shift, but looking at changing to "on click, if Ctrl is currently depressed" rather than "on Ctrl keydown" if that's even how it's working (I no longer remember :grimacing:)

oobayly:
I've done all the dev work on Linux and hadn't noticed this on Windows. I've also noticed that  the keydown behaviour is different on  Chrome on Linux than it is on Windows 10. Using this website - https://howtodoinjava.com/jquery/jquery-difference-between-keypress-and-keydown-events/  and holding down Ctrl:
• On Linux, the keydown event is fired only once
• On Windows, the keydown event is fired repeatedly
It may be an OS thing (Edge does the same), fun, fun fun...

oobayly:
What was happening was that the repeated keypress events being fired were calling `startNewPath` every time, and that was ending the last path and starting a new one every event.

Instead of trying to handle different browser event behaviour, `MeasureTool.startNewPath` returns early if a path is already being measured.

I've just tried the deployment on my Windows box (after clearing the cache) and it works as expected (or at least the way I expect it to work)

erikquinn:
Looks like a travis issue 🤔

erikquinn:
Does seem to be working again for me on the laptop!

src/assets/scripts/client/canvas/CanvasController.js
```diff
@@ -1217,6 +1219,149 @@ export default class CanvasController {
...
+        // Label backgrounds
+        cc.fillStyle = this.theme.SCOPE.RUNWAY_EXTENDED_CENTERLINE;
+
+        values.forEach((item) => {
+            const { x, y, labels } = item;
+            const height = 10 + (12 * labels.length);
+            const width = 150;
+
+            cc.fillRect(x, y, width, height);
+        });
```
erikquinn: Want to try to get the text box fill to always match the size of its content so it isn't taking up extra space that it doesn't need to (and thus blocking things behind it).

image

oobayly: I actually moved the colours to the themes, but stupidly forgot to push... 🤦‍♂️

I'll merge your changes and do the measurements (I actually removed the text box background for the default theme to march the style of the aircraft labels).

I also slightly changed the behaviour of start measuring to throw off it's called a second time so it can be tested for.

src/assets/scripts/client/InputController.js
```diff
@@ -1006,7 +1006,7 @@ export default class InputController {
      */
     _onRightMousePress(event) {
         if (MeasureTool.isMeasuring) {
-            this._removeLastMeasurePoint();
+            this._removePreviousMeasurePoint();
```
oobayly: That's much nicer behaviour, I wonder why I didn't think of that.

oobayly:
There were some changes I forgot to push up. They were mainly some code and test improvements, but also moved the colours to the themes. I used the colours you suggested for the Default theme, and used what I felt were optimal for the Classic theme.

And the labels are now sized to the text (that omission was down to laziness...)

PR #1468, opened 2019-10-26
feature/1191 - Consolidate command bar buttons

### LINKED ISSUE #1191: Consolidate command bar buttons

manuel3108:
Resolves #1191 

The purpose of this pull request is to improve the controls by grouping buttons


src/assets/style/module/controls/_controls.less
```diff
@@ -1,46 +1,113 @@
...
+    margin-left: 0;
+}
+
+.dropup-content {
+    display: none;
+    position: absolute;
+    background-color: @dark-gray;
+    box-shadow: 0 8px 16px 0 rgba(0, 0, 0, 0.2);
+    bottom: @command-bar-height;
+    float: right;
```
n8rzz: float won't apply on an absolutely positioned element

src/templates/_footer.hbs
```diff
@@ -1,59 +1,98 @@
...
-        <li class="control pause-toggle" title="Pause simulation">
-            <img src="assets/images/icons/pause.png" />
-        </li>
-        <li class="control toggle-speech" title="Enable speech synthesis (experimental)">
-            <img src="assets/images/icons/speech.png" />
-        </li>
-        <li class="control toggle-traffic" title="Adjust traffic rate">
-            <img src="assets/images/icons/traffic.png" />
+
+    <ul>
```
n8rzz: this should still use a class name.  directly styling elements has a different specificity that styling classnames (https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), and we prefer to style with classnames.

src/templates/_footer.hbs
```diff
@@ -1,59 +1,98 @@
...
+        <li class="dropup">
```
n8rzz: hmm, we may want to generalize this name a bit.  this item doesn't actually include a `dropup` element, it's just a menu item

manuel3108: jup, agree. Do you think that it's better now?

n8rzz:
It might make a little more sense to have this menu item last in the list

Screen Shot 2019-11-02 at 12 33 24 PM

manuel3108:
I have done the reorder. Where do you want to have the airpot-guide button?
I put it beside the switch-airport button, because it doesn't fit in any of those subcategories in my opinion.

package-lock.json
```diff
@@ -1889,7 +1889,7 @@
         },
```
n8rzz: hmm, interesting that these changed from `https` -> `http`.  

could you revert these changes by running: `git checkout origin/master package-lock.json` and then commit the result?

manuel3108: Done, didn't see that one!
I think you mean `git checkout origin/develop package-lock.json`

n8rzz: whoops!  yes, I did mean `develop`.  thank you!

PR #1422, opened 2019-08-18
bugfix/39 - Compass rose misaligned

wrksx:
Resolves #39
Resolves #1316

- Show and hide the compass marks around the view when selecting and deselecting an aircraft
- Align the compass rose to the selected aircraft

PR #1520, opened 2020-01-30
bugfix/1316 Mark deep render when aircraft is (de)selected

### LINKED ISSUE #1316: Mark deep render when aircraft is (de)selected
Playing v6.10 prod.

The compass around the screen does not appear when selecting an aircraft. Instead, i have to scroll to get it to render. Same goes for de-selecting an aircraft.

Mathew-Estafanous:
Resolves #1316

The compass now (dis)appears once the aircraft is (de)selected instead of having the player scroll the mouse wheel for it to initially appear.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -209,7 +209,7 @@ export default class InputController {
         this.$commandInput.val('');
 
         this._eventBus.trigger(EVENT.DESELECT_ACTIVE_STRIP_VIEW, {});
-        this._eventBus.trigger(EVENT.SELECT_AIRCRAFT);
+        this._eventBus.trigger(EVENT.TOGGLE_COMPASS);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Mathew-Estafanous:
> Since it's being used for both select and de-select, I'd encourage renaming `_onSelectAircraft` to `_onToggleAircraftSelection`. I'd also encourage creating a new event name, `SELECT_AIRCRAFT` should be used for just that, _selecting_ an aircraft and I feel like this is actually different concern.

Thanks for the suggestions. I sent an update with the changes you suggested. Would you say they are more appropriate now?

Mathew-Estafanous:
We could do that but I saw that other methods did the same thing by just instantly calling the _makeDeepRender() method.

n8rzz:
> Can't we actually just do a `trigger(EVENT.MARK_DEEP_RENDER)`? Or is there not an event for that. Seems like that would be much more straightforward, but I'm not sure if it's currently an event. I know we have `EVENT.MARK_DIRTY_CANVAS`, though that may have since been renamed, and I'm not sure if it marked a shallow or deep render.

Triggering that action directly is certainly an option, but I'd argue that is a little bit harder to track.  This way there is a pretty clear line between even triggered, why it was triggered, and resulting callback after event.

cake-pie:
Obsoleted by #1422, see [#1316 (comment)](https://github.com/openscope/openscope/issues/1316#issuecomment-650576877)

PR #1522, opened 2020-01-31
enhancement/1343 Regenerate inbound traffic when arrival rate changes

### LINKED ISSUE #1343: Regenerate inbound arrival traffic when changing spawn rate
Continuation of #16.

When we change the spawn rate for arrivals (this doesn't apply to departures), it will often take a long time for that change to be noticed, due to the fact that the spawn points are often very _very_ far away from the airspace boundary. So even though they're spawning at the new rate, the preexisting aircraft are what we'll be receiving at the border for the next 30 minutes or so before the new wave of arrivals reaches us.

Instead, we should clear out some of the arrivals that have not yet reached our airspace, and regenerate them at the _new_ spawn rate, so the desired effect is achieved sooner. There are two conflicting desires here though:
- The change in traffic should take effect quickly
- The change in traffic should not be too jarring to the user, as airplanes almost to the border are deleted, and new ones added, etc.

To achieve a balance between the two concepts, maybe we can have a small buffer, such as 10-30 miles from the airspace boundary, aircraft will stay in place, and only aircraft further out will be reset? Something like that to find a happy medium.

Mathew-Estafanous:
Resolves #1343 

[WIP] Trying to work on resetting the arrival aircraft once the arrival rate is reset.

What Works:
   - Deleting all the aircraft outside the airspace once the rate is changed.
   - The Settings panel has the option to turn off/on feature [does not completely work]

One issue that I would love help with is how to set `this.canResetArrivals` to false when the option "No" is chosen. For some reason, the event `EVENT.RESET_ARRIVALS_OPTION` does not seem to call the `onArrivalSettingsChange()` even though I set it to do exactly that.

[Also I'm pretty new so if you any criticism about what needs to be fixed then I'll gladly listen.]

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -22,6 +22,7 @@ import { isEmptyOrNotArray } from '../utilities/validatorUtilities';
 import { FLIGHT_CATEGORY } from '../constants/aircraftConstants';
 import { EVENT, AIRCRAFT_EVENT } from '../constants/eventNames';
 import { INVALID_INDEX } from '../constants/globalConstants';
+import { THEME } from '../constants/themes';
 
 // Temporary const declaration here to attach to the window AND use as internal property
 const aircraft = {};
@@ -144,6 +145,7 @@ export default class AircraftController {
      */
     _setupHandlers() {
         this._onRemoveAircraftHandler = this.aircraft_remove.bind(this);
+        this._onRemoveOutsideAircraft = this.aircraft_remove_all_outside.bind(this);
 
         return this;
     }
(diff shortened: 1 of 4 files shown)


Mathew-Estafanous:
I would love help with one main issue that I can't seem to figure out. When I want to respawn the arrivals, I was thinking about spawning arrivals the same way they are initially spawned but that meant I would need the new "arrival rate" that was set in the _TrafficRateController.js_ file. Because this rate is private in the _TrafficRateControlle_r, I was thinking about making an EVENT and calling it in the _buildPreSpawnAircraft.js_ file. The problem I faced is when I called the EventBus.trigger method in _buildPreSpawnAircraft_  it would give me an error. 

`const rate = EventBus.trigger(EVENT.GET_ARRIVAL_TRAFFIC_RATE);`
`const entrailDistance = (spawnSpeed / spawnPatternJson.rate) * _calculateRateOffset(rate);`
**Error: "EVENT" is not defined**

Does anyone have any ideas on what I could possibly do or am I going about this completely wrong?

erikquinn:
First real quick-- the game option is not needed; we should remove all aircraft who are more than a fixed distance away from the airspace as described in the original issue writeup #1343.

So the lint error about `EVENT` not being defined would indicate that you haven't imported it to the file where you're using it. Just look at the top of that file, and follow the pattern of the other imports to import EVENT.

However, we do _not_ want to use an event to "get" anything, only to declare that something has happened which should cause multiple simultaneous actions in multiple different classes.

The issue you're facing lies with a deeper problem: the only class which uses prespawned aircraft is the `SpawnPatternModel` itself, when it populates its own `SpawnPatternModel.preSpawnAircraftList`... There is no scenario where any other class should even be _allowed_ to manipulate the spawning of aircraft. For that reason, I think we ought to either:
(a) have all prespawn code moved to the SpawnPatternModel, or
(b) remove all prespawn _and regular spawn_ code from anywhere other than the AircraftController and SpawnPatternModel.

All this has always been a bit messy (with small improvements over time) and we may have to clean things up a bit before we can get a good solution. I would say for sure though that the regeneration of traffic should be triggered from the AircraftController. It removes the requisite aircraft and tells the SpawnPatternModel to generate a new prespawn list, which it then uses to populate the new arrivals/departures.

PR #1541, opened 2020-03-10
refactor/1537 - Make KJFK traffic more realistic

### LINKED ISSUE #1537: Make KFJK traffic more realistic
Suggested by u/Slott2Boomr on Reddit

Fechulo:
Resolves #1537 

### [Test me here!](https://openscope-dev-pr-1541.herokuapp.com/)


assets/airports/kjfk.json
```diff
@@ -31,217 +31,247 @@
...
-                [40.24996028, -74.33294528],
-                [40.15420583, -73.99997306],
-                [40.13753194, -73.89581917],
-                [40.06392917, -73.79860389]
+                ["N40.03355", "W73.71679"],
+                ["N40.06299", "W73.67482"],
+                ["N40.09182", "W73.63765"],
+                ["N40.21134", "W73.46444"],
+                ["N40.23725", "W73.43697"],
+                ["N40.40881", "W73.25296"],
```
erikquinn: I don't really see a benefit to this change-- it trades away 3 digits of precision to add two `"` and a `N`. Using straight up decimal degrees like `40.73953111, -74.26430806` is the easiest of all accepted forms for the sim to accept coordinate data, so we shouldn't discourage the use of that form.

Please revert changes to this section.

erikquinn: I do see these changes to other sections too-- any chance this is from the qgsopenscope plugin's fixes feature?

erikquinn: Of course, with 5 digits after the decimal, it looks like the highest possible error of that rounding would be maybe about a meter off (negligible for anything other than airport layout markings). Is this being done because it's your intent to move all the airports to a consistent coordinate format?

Fechulo: It is indeed the output from qgsopenscope. I'll change it back in a sec. @oobayly maybe you could change the coordinate format used in the plugin?

Fechulo: @erikquinn I never thought to ask this, is there a particular format that would be most efficient for performance or other reasons that we should be trying to stick to?

erikquinn: straight numbers, `[34.612813545, -83.341892057], though really it's negligible since it converts them when importing the data from the file. It's not like it causes a calculation every frame, so it doesn't make an appreciable difference what format the coordinates are given in.

PR #1553, opened 2020-03-30
Enable with tab key

nickajeglin:
Expressly note that you have to hit tab first. Right now this is only in the upper level commands file.

The purpose of this pull request is to improve documentation.


erikquinn:
Per the contribution guidelines, please create an issue that describes what the problem/improvement is that this pull request will fix, and then include `Resolves #xxxx` in your initial comment in this PR.

The PR should then be titled `documentation/xxxx - Improve command reference documentation` or similar.

PR #1592, opened 2020-06-02
enhancement/1581 - Add Ability to Clear Scratchpad

### LINKED ISSUE #1581: Add way to clear scratchpad
Add a case where setting a scratchpad to something like "---" would clear the aircraft's scratchpad.

j-651:
Ideally resolves #1581

The purpose of this pull request is to allow clearing an aircraft's scratchpad. Setting the scratchpad to `---` would clear it.

Not sure if the syntax is 100% accurate as this is my first 'code' commit to this repo. Hopefully it's at least a starting point!

### DIFF
--- a/src/assets/scripts/client/scope/ScopeModel.js
+++ b/src/assets/scripts/client/scope/ScopeModel.js
@@ -301,6 +301,10 @@ export default class ScopeModel {
         if (scratchPadText.length > 3) {
             return [false, 'ERR: SCRATCHPAD MAX 3 CHAR'];
         }
+        
+        if (scratchPadText === '---') {
+            return radarTargetModel.setScratchpad('');
+        }
 
         return radarTargetModel.setScratchpad(scratchPadText.toUpperCase());
     }


j-651:
Just tested this and it does work, removing the scratchpad.

image

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -301,6 +301,10 @@ export default class ScopeModel {
         if (scratchPadText.length > 3) {
             return [false, 'ERR: SCRATCHPAD MAX 3 CHAR'];
         }
+        
+        if (scratchPadText === '---') {
+            return radarTargetModel.setScratchpad('');
+        }
```
erikquinn, line 307: Let's change this to use a single period mark to clear the scratchpad. This is how it is done in vSTARS, and thus will be easier for those users to adjust to, while being no more complicated than the `---` (but with the added benefit of being fewer keystrokes).

j-651:
```suggestion
        if (scratchPadText === '.') {
            return radarTargetModel.setScratchpad('');
        }
```

Would this be the desired implementation then?

I'll check on the docs and update them as necessary. Also going to work on moving this to an openscope branch as I don't think it was.

j-651:
Only documentation I can find with reference to the scratch pad is the [Scope Command Reference](https://github.com/openscope/openscope/blob/2e61441e92916af172d4bd4a518987a9198471b4/documentation/scope-commands.md). Will update.

Do any of the tests need to be modified?

src/assets/scripts/client/scope/ScopeModel.js
erikquinn: yep, exactly.

erikquinn:
Ideally, yes, we would want to add a new test. If you find the tests for the scratchpad, you should be able to follow the patterns to add a new test that checks that when the `.setScratchpad()` method is called with `.`, the scratchpad is set to the expected value. You can always hit up `#code-help` in slack and somebody can help you out 😃 

slack.openscope.io

j-651:
Replaced by #1602

PR #1602, opened 2020-06-15
enhancement/1581 - Add Ability to Reset Scratchpad

### LINKED ISSUE #1581: Add way to clear scratchpad

j-651:
Resolves #1581.
Replaces #1592.

Add the ability to clear/reset an aircraft's scratchpad. Typing `. [CID]` (scope command) will reset the aircraft's scratchpad to its destination or first fix as automated on spawn.

### DIFF
--- a/src/assets/scripts/client/scope/RadarTargetModel.js
+++ b/src/assets/scripts/client/scope/RadarTargetModel.js
@@ -539,8 +539,7 @@ export default class RadarTargetModel {
      *
      * @for RadarTargetModel
      * @method setDefaultScratchpad
-     * @private
-     * @chainable
+     * @return {array} [success of operation, system's response]
      */
     setDefaultScratchpad() {
         if (this.aircraftModel.isDeparture()) {
--- a/src/assets/scripts/client/scope/RadarTargetModel.js
+++ b/src/assets/scripts/client/scope/RadarTargetModel.js
@@ -172,8 +172,7 @@ export default class RadarTargetModel {
         this._theme = theme;
 
         this._init(aircraftModel)
-            .enable()
-            .setDefaultScratchpad(aircraftModel);
+            .enable();
     }
 
     /**
@@ -268,6 +267,8 @@ export default class RadarTargetModel {
         this._dataBlockLeaderLength = this._theme.DATA_BLOCK.LEADER_LENGTH;
         this._routeString = aircraftModel.fms.getRouteString();
 
+        this.setDefaultScratchpad();
+
         return this;
     }
 
@@ -276,6 +277,7 @@ export default class RadarTargetModel {
     *
     * @for RadarTargetModel
     * @method enable
+    * @chainable
     */
     enable() {
         this._eventBus.on(EVENT.SET_THEME, this._setTheme);
@@ -288,9 +290,12 @@ export default class RadarTargetModel {
     *
     * @for RadarTargetModel
     * @method disable
+    * @chainable
     */
     disable() {
         this._eventBus.off(EVENT.SET_THEME, this._setTheme);
+
+        return this;
     }
 
     /**
@@ -298,6 +303,7 @@ export default class RadarTargetModel {
     *
     * @for RadarTargetModel
     * @method reset
+    * @chainable
     */
     reset() {
         this.aircraftModel = null;
@@ -310,6 +316,8 @@ export default class RadarTargetModel {
         this._interimAltitude = INVALID_NUMBER;
         this._isUnderOurControl = true;
         this._routeString = '';
+
+        return this;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -544,6 +526,26 @@ export default class RadarTargetModel {
...
+     * @chainable
+     */
+    setDefaultScratchpad(aircraftModel) {
+        if (aircraftModel.isDeparture()) {
+            this.scratchPadText = aircraftModel.fms.getFlightPlanEntry();
+
+            return [true, 'RESET SCRATCHPAD'];
+        }
+
+        this.scratchPadText = aircraftModel.destination.substr(1, 3);
```
erikquinn, line 544: We really don't need to _pass_ the aircraftModel like this, we should instead just access it with `this.aircraftModel`, as used in other methods of this class.

This change is the reason some tests are now failing-- because the tests aren't passing the aircraftModel, they're expecting you to retrieve it from the RadarTargetModel.aircraftModel.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -172,8 +172,8 @@ export default class RadarTargetModel {
         this._theme = theme;
 
         this._init(aircraftModel)
-            ._initializeScratchPad()
-            .enable();
+            .enable()
+            .setDefaultScratchpad(aircraftModel);
```
erikquinn, line 176: enable should be done last, only when the thing is fully ready-- so let's reverse these.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -299,6 +279,8 @@ export default class RadarTargetModel {
     */
     enable() {
         this._eventBus.on(EVENT.SET_THEME, this._setTheme);
+
+        return this;
```
erikquinn, line 283: 👍 just add `@chainable` to the docblock to whenever we return `this`.

src/assets/scripts/client/scope/ScopeModel.js
```diff
@@ -302,6 +302,10 @@ export default class ScopeModel {
             return [false, 'ERR: SCRATCHPAD MAX 3 CHAR'];
         }
 
+        if (scratchPadText === '.') {
+            return radarTargetModel.setDefaultScratchpad(radarTargetModel.aircraftModel);
+        }
```
erikquinn, line 307: Don't need to pass anything here, since the RadarTargetModel can access its own aircraftModel.

src/assets/scripts/client/scope/RadarTargetModel.js
j-651: If enable() comes at the end, it shouldn't need to be chainable, right?

src/assets/scripts/client/scope/RadarTargetModel.js
erikquinn: More accurately (as discussed), we should just do the .setDefaultScratchpad() from _within_ `this._init()`.

That would leave _this_ line as:
```
this._init(aircraftModel)
    .enable();
```

src/assets/scripts/client/scope/RadarTargetModel.js
erikquinn: all the stuff where we initialize is fine to be chainable (so .init(), .reset(), .enable(), .disable(), ._initSomethingInParticular(), etc). Normal methods like .doSomething() should not be chainable.

src/assets/scripts/client/scope/RadarTargetModel.js
```diff
@@ -544,6 +534,26 @@ export default class RadarTargetModel {
...
 
+    /**
+     * Set default value of the scratchpad
+     *
+     * @for RadarTargetModel
+     * @method setDefaultScratchpad
+     * @private
+     * @chainable
+     */
+    setDefaultScratchpad() {
```
erikquinn, line 545: This is actually a _public_ method, meaning it can be accessed from OUTSIDE the RadarTargetModel class (which we do-- the ScopeModel calls RadarTargetModel.setDefaultScratchpad). A private method would be one that is only allowed to be accessed from within the RadarTargetModel class. If no other classes _need_ to use a given method, it should be made private (because bugs).

Also, all private methods should begin with an underscore. Since this doesn't have one, that's how we know whether it's public/private at a glance when we go to use it. So if you ever find yourself in a situation where you're doing like `this.someObject._doSomething()`, you _know_ the designer of `someObject` has a reason why you probably shouldn't do that.

tldr: remove `@private`

j-651: OK, good to know. That was leftover from when I copy-pasted the `_initializeScratchpad()` function and I didn't make that distinction with underscores and private functions. Pushing correction now.

j-651: As I look back at this, I'm also removing `@chainable` since per our conversation that no longer applies and adding `@return {array} [success of operation, system's response]`. since that was missing

PR #1619, opened 2020-07-01
feature/1593 - Add search bar in airport list

### LINKED ISSUE #1593: Suggestion: Search Bar for Airport Picker
Browser: N/A

Airport: All

#### Steps to reproduce
- Not an issue

As the list of airports grows, is there any interest in adding a searchbar to the top of the airport list that lets you filter through the available ones?

This would search by ICAO code or airport name.

JPNite3:
Resolves #1593 

The purpose of this pull request is to add a search bar to the top of the airport list for the user to search for airports.

### DIFF
--- a/src/assets/scripts/client/ui/UiController.js
+++ b/src/assets/scripts/client/ui/UiController.js
@@ -737,7 +737,6 @@ class UiController {
      * @for UiController
      * @method onInitiateAirportSearch
      */
-
     onInitiateAirportSearch() {
         EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-search', 'start');
 
--- a/src/assets/scripts/client/ui/UiController.js
+++ b/src/assets/scripts/client/ui/UiController.js
@@ -700,9 +700,10 @@ class UiController {
      * @for UiController
      * @method _onClickOpenAirportDialog
      */
+
     _onClickOpenAirportDialog() {
-        $('#search').on('keyup', function() {
-            let value = $(this).val().toLowerCase();
+        $('.search').on('keyup', function() {
+            const value = $(this).val().toLowerCase();
             $('.dialog-body li').filter(
                 function() {
                     $(this).toggle($(this).text().toLowerCase().indexOf(value) > -1);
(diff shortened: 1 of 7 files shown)


src/assets/scripts/client/ui/UiController.js
```diff
@@ -662,6 +677,24 @@ class UiController {
         this.$switchAirport.addClass(SELECTORS.CLASSNAMES.ACTIVE);
     }
 
+
+    /**
+     * @for UiController
+     * @method onInitiateAirportSearch
+     */
+
```
n8rzz, line 685: remove empty space

src/assets/scripts/client/ui/UiController.js
```diff
@@ -662,6 +677,24 @@ class UiController {
...
+    onInitiateAirportSearch() {
+        EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'airport-search', 'start');
+
+        const value = this.$airportDialog.find(SELECTORS.DOM_SELECTORS.AIRPORT_SEARCH).val().toLowerCase();
+
+        $('.dialog-body li').each(
```
n8rzz, line 691: This isn't the correct place for this, as you guessed, but you're close.  I'll follow up with some more info on where this should actually live.

n8rzz (dismissed):
I'll see if I can find where that filter should live, but overall looking pretty good.

src/assets/scripts/client/ui/UiController.js
tomomasuda: Thank you!

j-651:
@tomomasuda Since there are custom `:focus` styles applied, I'd recommend setting `outline: 0` for the new input.

Also, it might be helpful to show a simple "No airports found" message when the query does not match any of the available airports.

src/assets/scripts/client/ui/UiController.js
j-651: @n8rzz do you have a suggestion as to where this should be moved?

erikquinn (approved):

PR #1629, opened 2020-07-14
devops/1628 - Replace travis status checks with GitHub Actions

### LINKED ISSUE #1628: Move all status checks to github actions
- coveralls is only measuring the difference between subsequent pushes, not the difference between a PR and its base branch
- all checks should appear individually in the checks list at the bottom of a PR, not embedded within the Travis-CI check as individual jobs

erikquinn:
Resolves #1628.

For an example of how this would appear without the existing checks. Intentionally failed some to show the messages:

Note that the coverage failure persists upon subsequent pushes unless the lost coverage is rectified.

image


erikquinn:
Still needs checks set up on develop/master _on push_, then this will be ready.

PR #1634, opened 2020-08-03
enhancement/1633 - Add fixes to PDX

### LINKED ISSUE #1633: Add missing fixes to PDX
Airport: KPDX

Add a few missing fixes from instrument procedures.

cessnahat:
Resolves #1633

The purpose of this pull request is to add missing IAP fixes to PDX.


Fechulo:
I'm not sure how I feel about adding fixes that aren't used in any SID or STAR. Since they're not very useful in openScope, they're mainly just cluttering the scope. Generally speaking, just having the IAF and FAF is enough. If you'd like, it looks like the OCITY arrival has changed since PDX was last updated so you could fix that and maybe update the traffic if you're up for it.

N864T:
Good point Fechulo. I still think the IFs should be added, but maybe all those other ones aren't necessary. I do think some key fixes from the missed approach procedures should be added, though. And I'll check out the OCITY while I'm at it :)

Fechulo (approved):
Nice work!

PR #1655, opened 2020-10-14
feature/1654 - Add KGSO

### LINKED ISSUE #1654: Add KGSO - Piedmont Triad Int'l
GSO is a commercial airport in Greensboro, NC, that also serves Winston-Salem and High Point, NC. It is third-busiest airport in North Carolina and a small hub for FedEx. It has a number of standard and RNAV departure and arrival procedures.

MichaelDMcCracken:
Resolves #1654 

The purpose of this pull request is to add KGSO.


MichaelDMcCracken:
Ready to be reviewed.

Fechulo (changes requested):
Hey @MichaelDMcCracken, awesome work on this so far. Just a few fairly minor things before I approve this. If you have any questions, feel free to reach out on Slack :)

- [x] Make sure to include an AIRAC key at the top
- [x] Magnetic variation is -8.5 according to charts
- [x] The Wind and departure and arrival runway keys should go before the defaultMaps key
- [x] Since support for multiple airspace sections was brought in in the last update, it would be awesome if it could be added to this, as per the following diagram 
image
- [x] Runway 5L/23R should be defined before 2R/23L
- [x] Runway 14 doesn't have an ILS
- [x] QUAK7 exitPoints should be in alphabetical order
- [x] TRI9 exitPoints should be in alphabetical order
- [x] TRSH1 05L/R departures should turn to hdg 050 on passing 1440ft also
- [x] BLOCC2 has no strict altitude restriction at TENNI or BLOCC. Those "Expect" restrictions shouldn't be included in the STAR definitions
- [x] BROOK4 entryPoints should be in alphabetical order
- [x] BROOK4 has no strict altitude restriction at MURKY or TRAKS
- [x] HENBY3 entryPoints should be in alphabetical order
- [x] HENBY3 has no strict altitude restriction at HENBY
- [x] TRAKS3 entryPoints should be in alphabetical order
- [ ] TRAKS3 has no strict altitude restriction at TRAKS
- [x] Airways key should go near the top, just after the runway definitions
- [x] The current traffic rate of 31.8AcpH falls under the easy category. So you can either reduce the traffic (below 20AcPh for beginner level), or set the level to easy on the airportLoadList.js
- [ ] Make sure you add an airport guide also.

You can request another review from me once you've fixed these and I'll check them off on my next review. Great work!

MichaelDMcCracken:
> Those "Expect" restrictions shouldn't be included in the STAR definitions

I've made all the other changes so far. I'll make this one as well, but I guess I wish I understood the reason behind not putting them there. The issue I find is that in most cases the center "controller" will leave the aircraft too high, because most of those fixes with the expect altitudes are outside of the approach controller's airspace. So the only other way to get the aircraft to cross TRAKS at 11,000' as you'd get in real life is to have the spawnPattern set to 11,000', unless I'm missing something.

Fechulo:
@MichaelDMcCracken The logic behind it is that (in real life) controllers would give the actual descent clearances for those STARs, whereas with RNAV arrivals, controllers will issue a 'descend via' instruction which tells pilots to descend as per the restrictions along the STAR. It can be inconvenient in the sim to do it this way though. As you point out, we don't have adjacent controllers descending our arrivals to the appropriate height, so I'm willing to ignore it in this case and leave the restrictions there.

erikquinn:
Changing to a "feature", since this is a completely _new_ addition. Enhancements are improvements on something that already exists. Carry on! :)

MichaelDMcCracken:
Ignore my commit 1b43efd, accidentally included additional files that were unrelated.

MichaelDMcCracken:
@Fechulo 
Everything you've requested had been checked off. Ready for final review.

Fechulo (approved):
Nice work! 🎉

PR #1659, opened 2020-10-18
enhancement/1658 - LIPZ edit airspace

### LINKED ISSUE #1658: LIPZ - fix Airspace
Airport: LIPZ

#### Issue Description
Airspace structure doesn't go down to ground level, making departures spawn outside airspace.

Fratee:
Resolves #1658 

The purpose of this pull request is to fix wrong airspace in LIPZ


Fratee:
Hi @Fechulo ! Small fix ready for review :)

Fechulo:
Hey @Fratee, thanks for fixing this. Would you be able to move the airspace labels around a little so the depcition is clearer?
image

Fratee:
> 
> 
> Hey @Fratee, thanks for fixing this. Would you be able to move the airspace labels around a little so the depcition is clearer?

Hi @Fechulo , would love to but i don't know how to. Do i have to add another property for the airspace? I didn't find much in docs (I probably missed something after the new update). Can you please give me more direction? Thanks
Also is it possible to give name to the airspace instead of "(#1)" ?

Fechulo:
@Fratee Sorry, meant to include this in my comment:
```
"labelPositions": [
    [50.1, 9.25]
],
```
You can put that in as a key for each polygon. Not sure about the changing the name though... @erikquinn?

Fratee:
@Fechulo Airspace labels are done. About naming is because the different CTR airspace has a designator and would be nice to show the actual label "Zone '1'"
immagine
Not big deal if can't be done

Fechulo:
I've tried a few options and none of them seem to be working, so it might not be possible to add a name. If we figure it out we can add them then.

Fechulo (approved):

erikquinn (changes requested):
image

@Fratee Please show ALL altitudes owned in each polygon, rather than depending on other overlapping ones. It must be obvious when turning on the labels what altitudes are owned, which is why we would show polygon 0 as being `000-125` instead of `000-015`.

Fixing!

erikquinn (approved):
Changes complete.

PR #1661, opened 2020-10-20
refactor/1651 - Fix airports with broken airspace

### LINKED ISSUE #1651: Departure Aircraft Missing
Browser: Chrome

Airport: CYHZ, CYOW, EGGW, EICK, LIPZ, LROP

Approximate play time before issue: When starting

Issue Description
When opening up any of the listed airports, all departure aircraft say they are leaving the airspace without being on their route. The score is also reduced by 25 for each aircraft.

Screenshot 2020-10-09 at 5 00 11 PM

Steps to reproduce
Open up openScope and select one of the previously mentioned airports.

Fechulo:
Resolves #1651 

Fixes the following airports:
- [x] CYHZ
- [x] CYOW
- [x] EGGW
- [x] LROP
- [x] LOWW (technically works, but the FLs are incorrect)

Test [here](https://openscope-dev-pr-1661.herokuapp.com/)


erikquinn (changes requested):
image

To minimize the number of polygons (and the need for enclave/exclave situations), these should always be done by selecting an area and showing ALL ALTITUDES owned in that area. For example, in the central circle of this polygon, we own 000-180, but it shows "000-017". We _could_ just stack the 017-029 and the 029-125 and the 125-180 labels above the 000-017 label to achieve the same explanation, but it is obviously less clear for no real reason. The polygon itself.

And again, the stacking of 029-125 with 125-180 should not be two different polygons-- just 029-180.

I realize some of this may be being done to easily correspond with the AIPs, but the calculations of determining whether an aircraft is within a given polygon is extremely time consuming, so all possible efforts should be made to use the simplest and fewest polygons without a loss of accuracy. Complex airspace with tens of polygons might have noticeably sluggish panning performance.

Fechulo:
@erikquinn Cool that makes sense 👍 I don't think I'll have the time to make these changes before the next release, but I don't want to leave these airports unusable in the meantime. Any chance we could merge this (with the pinky promise that I'll improve this as soon as I have the time)? Alternatively deactivate the broken airports until they're fully functional?

erikquinn:
No worries, that'll work. I'll see what I can do in the next two days, hopefully avoid turning them off. Mostly just wanted to pass it long before you take the time to do (or review) more airports with similar situations! 😃

erikquinn:
CYHZ also had a huge number of points for a tiny polygon; I've simplified that to maximize performance. Probably something to keep an eye on, as people shouldn't be using this many points unless they make a clear visual difference when zoomed all the way in.

image
...simplified to...
image

erikquinn:
Another side effect-- in the below image, it's extremely difficult to tell what airspace we own, also making it difficult to identify where there could be a mistake. In this case, we quite literally do not own between 4001 and 4999 feet, causing a/c to leave the airspace when descending, and quickly reentering:

image

erikquinn:
I was able to do CYHZ, LOWW, and LROP. Unfortunately, EGGW and CYOW were too complicated and had to be deactivated until they can be fixed up.

erikquinn (approved):

PR #1663, opened 2020-10-24
refactor/1662 - Fix KLAX airspace

### LINKED ISSUE #1662: LAX Airspace
Browser: All

Airport: KLAX

Approximate play time before issue: N/A, first airspace penalty occurs ~5 mins after starting

#### Issue Description
This happens with a variety of STARS as well as on the FAC. Aircraft enter the airspace and even with descent via the STAR, they leave the airspace. 

#### Steps to reproduce
- Select KLAX
Give all aircraft .dvs (unless coming in via KIMMO or another similar STAR) and expect (either 25L or 24R)
Play as necessarily, you will notice that descent via the STAR on many (one example in the ANJLL) STARS forces a penalty after a few minutes
I have also noticed that when giving vectors to the FAC, even if according to the airspace depiction are in my airspace, they still leave occasionally.

Fechulo:
Resolves #1662


Fechulo:
@matthewpow you can test it [here](https://openscope-dev-pr-1663.herokuapp.com/). Let me know if you still encounter any issues. If possible provide a screenshot of the aircraft (and the area around it) so I know where the problematic sector is

Fechulo:
There's still a few gaps between sectors, so some more work needs to be done here. Arrivals from the south can no longer be separated so these spawnPatterns should be tweaked

matthewpow (changes requested):
I do agree with gaps between sectors. I have not experienced a problem with south arrivals however with more playtime that may present itself. There were less issues with arrivals from the east that were the southern of the two (This is confusing... I know. I have attached a picture and the one circled is the one that leaves the airspace if "fph" is used.) I only had a chance to test for a few minutes however once I get more time, I'll vector people down to final and let you know what I see.
Screen Shot 2020-10-24 at 5 28 34 PM

PR #1678, opened 2020-11-18
bugfix/1565

### LINKED ISSUE #1565: Aircraft taxi from the runway to the runway
You can instruct aircraft already on the runway to taxi to the runway. Although a cool 'feature', this is not desired behaviour
euro

spectreofmyself:
Resolves #1565 

The purpose of this pull request is to prevent aircraft from taxiing to a runway they're already holding short of

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftModel.js
+++ b/src/assets/scripts/client/aircraft/AircraftModel.js
@@ -1305,8 +1305,12 @@ export default class AircraftModel {
             return [false, 'unable to taxi to runway, we have just landed'];
         }
 
-        if (this.flightPhase === FLIGHT_PHASE.WAITING && this.fms.departureRunwayModel === runwayModel) {
-            return [false, 'we\'re already holding short of the runway'];
+        if (this.fms.departureRunwayModel === runwayModel) {
+            if (this.flightPhase === FLIGHT_PHASE.WAITING) {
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1305,6 +1305,16 @@ export default class AircraftModel {
             return [false, 'unable to taxi to runway, we have just landed'];
         }
 
+        if (this.fms.departureRunwayModel === runwayModel) {
+            if (this.flightPhase === FLIGHT_PHASE.WAITING) {
+                return [false, 'we\'re already holding short of the runway'];
+            }
+
+            if (this.flightPhase === FLIGHT_PHASE.TAXI) {
+                return [false, 'we\'re already taxiing to that runway'];
```
erikquinn, line 1314: Let's include the runway number here as well, so if someone doesn't notice it immediately, they can notice the red response and have a bit more information to remember what they did wrong.

erikquinn (changes requested):
Nice work @spectreofmyself!

We want to add tests to cover the new branches in logic (which is why the checks on the PR are failing).

I'll take care of that and the below issue.

Thanks for finding and fixing this!

erikquinn (approved):
:shipit:

PR #1679, opened 2020-11-18
feature/1670 - Add command to get coordinates at clicked point

### LINKED ISSUE #1670: Debug Tool: Get coordinates at clicked point
To ease airport development in terms of airspace, add a keyboard shortcut to get the coordinates at a clicked point.

erikquinn:
Resolves #1670.

### DIFF
--- a/src/assets/scripts/client/InputController.js
+++ b/src/assets/scripts/client/InputController.js
@@ -225,15 +225,17 @@ export default class InputController {
      * @private
      */
     _addMeasurePoint(event, shouldReplaceLastPoint = false) {
-        const mouseCanvasPos = CanvasStageModel.calculateCanvasPositionFromPagePosition(
+        const mouseCanvasPosition = CanvasStageModel.calculateCanvasPositionFromPagePosition(
             event.pageX, event.pageY
         );
-        let relativePosition = CanvasStageModel.calculateRelativePositionFromCanvasPosition(...mouseCanvasPos);
+        let relativePosition = CanvasStageModel.calculateRelativePositionFromCanvasPosition(...mouseCanvasPosition);
 
         // Snapping should only be done when the shift key is depressed
         if (event.originalEvent.shiftKey) {
-            const [aircraftModel, distanceFromAircraft] = this._findClosestAircraftAndDistanceToCanvasPosition(...mouseCanvasPos);
-            const [fixModel, distanceFromFix] = this._findClosestFixAndDistanceToCanvasPosition(...mouseCanvasPos);
+            const [aircraftModel, distanceFromAircraft] = this._findClosestAircraftAndDistanceToCanvasPosition(
+                ...mouseCanvasPosition
+            );
+            const [fixModel, distanceFromFix] = this._findClosestFixAndDistanceToCanvasPosition(...mouseCanvasPosition);
             let distance;
             let nearestModel;
 
@@ -1001,7 +1003,7 @@ export default class InputController {
     }
 
     /**
-     * Retrieve and return the x/y offset from the airport center (km) where the user clicked in the provided event
+     * Given a mouse click event, retrieve and return the [x, y] offset from the airport center, in km
      *
      * @for InputController
      * @method _calculateRelativePositionFromEvent
@@ -1052,10 +1054,31 @@ export default class InputController {
     }
 
     /**
+     * Log the provided lat/lon coordinates to the console, display in command log, and copy to clipboard
+     *
+     * @for InputController
+     * @method _logAndCopyCoordinates
+     * @param latLonCoordinates {array<number>} [lat, lon]
+     * @returns undefined
+     * @private
+     */
+    _logAndCopyCoordinates(latLonCoordinates) {
+        const coordinateText = latLonCoordinates.map((coord) => coord.toFixed(9)).join(', ');
+
+        window.navigator.clipboard.writeText(coordinateText).then(() => {
+            console.log(coordinateText);
+            UiController.ui_log(`Clicked coordinates: ${coordinateText} (logged to console and copied to clipboard!)`, true);
+        });
+    }
+
+    /**
      * Triggered when a user clicks on the `right` mouse button and
      * records the position of the `right click` event.
      *
+     * @for InputController
+     * @method _onRightMousePress
      * @param event {jquery Event}
+     * @returns undefined
      * @private
      */
     _onRightMousePress(event) {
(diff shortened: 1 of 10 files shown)


src/assets/scripts/client/InputController.js
```diff
@@ -221,11 +222,12 @@ export default class InputController {
...
-            event.pageX, event.pageY
-        );
-        const { x, y } = currentMousePosition;
-        let modelToUse = this._translatePointToKilometers(x, y);
+        // const currentMousePosition = CanvasStageModel.translateMousePositionToCanvasPosition(
+        //     event.pageX, event.pageY
+        // );
+        // const { x, y } = currentMousePosition;
+        // let modelToUse = this._translatePointToKilometers(x, y);
+        let modelToUse = this._getMouseRelativePositionFromEvent(event);
```
n8rzz, line 230: is there a more specific name that could be used here?  is this a `position` of some sort?  Also, what are we _using_ this for?  

perhaps something like `positionToStartMeasurement` or something similar

src/assets/scripts/client/InputController.js
```diff
@@ -1002,6 +1003,38 @@ export default class InputController {
...
+
+    /**
+     * Retrieve and return the x/y relative position where the user clicked in the provided event
+     *
+     * @for InputController
+     * @method _getMouseRelativePositionFromEvent
+     * @param event {jQuery Event}
+     * @return {array<number>}
+     */
+    _getMouseRelativePositionFromEvent(event) {
```
n8rzz, line 1031: are we getting, calculating or translating here?

src/assets/scripts/client/InputController.js
```diff
@@ -1002,6 +1003,38 @@ export default class InputController {
...
+        const { x, y } = this._getMousePositionRelativeToCenterOfViewFromEvent(event); // ignores panning/zooming
+        const relativePosition = this._translatePointToKilometers(x, y);
```
n8rzz, line 1033: this `const` could be removed in favor of (adding an empty line 😉 ) and directly returning:
```js
return this._translatePointToKilometers(x, y);
```

erikquinn:
@n8rzz thanks 😄 Will address when I get back to this one. Just had to stop work briefly and wanted to resume on a different machine, so I pushed her in a very incomplete state. Will incorporate these changes upon resuming!

src/assets/scripts/client/InputController.js
erikquinn: more specifically, it's a relative position. just updated the method name on this line in a find-and-replace really, but a good callout and it now is clearer what it does 👍

src/assets/scripts/client/InputController.js
erikquinn: get, then calculate. So we'll call it `._calculateRelativePositionFromEvent()`

src/assets/scripts/client/InputController.js
erikquinn: I'll do you one better... the whole method wasn't needed 😉

src/assets/scripts/client/InputController.js
```diff
@@ -221,16 +222,15 @@ export default class InputController {
      * @private
      */
     _addMeasurePoint(event, shouldReplaceLastPoint = false) {
-        const currentMousePosition = CanvasStageModel.translateMousePositionToCanvasPosition(
+        const mouseCanvasPos = CanvasStageModel.calculateCanvasPositionFromPagePosition(
```
n8rzz, line 225: please use complete words in names `pos` -> `position`

src/assets/scripts/client/InputController.js
```diff
@@ -975,21 +975,36 @@ export default class InputController {
         return this._aircraftCommander.runCommands(aircraft, parsedCommand.args);
     }
 
+    /**
+     * Retrieve and return the x/y offset from the airport center (km) where the user clicked in the provided event
```
n8rzz, line 979: I know what you're saying here, but the wording is a bit tricky.  Maybe something like:

```txt
Given a user mouse click, retrieve and return x/y offset from airport center
```

src/assets/scripts/client/InputController.js
```diff
@@ -1025,6 +1040,23 @@ export default class InputController {
...
 
+        // copy mouse click position to clipboard on shift+alt+rightclick
+        if (event.originalEvent.shiftKey && event.originalEvent.altKey) {
+            const relativePosition = this._calculateRelativePositionFromEvent(event);
+            const referencePosition = AirportController.current.positionModel;
+            const latLonCoordinates = DynamicPositionModel.calculateGpsCoordinatesFromRelativePosition(
+                relativePosition, referencePosition
+            ).map((coord) => coord.toFixed(9)).join(', ');
+
+            navigator.clipboard.writeText(latLonCoordinates).then(() => {
```
n8rzz, line 1051: * please abstract this to it's own method
* instead of accessing `navigator` directly, access via the `window`: `window.navigator`. That makes stubbing easier later on.

n8rzz (changes requested):

src/assets/scripts/client/InputController.js
erikquinn: I was cheating for the 130 character limit for lint 😞 

I can break it into two lines I suppose.

src/assets/scripts/client/InputController.js
erikquinn: `Given a mouse click event, retrieve and return the [x, y] offset from the airport center, in km`

src/assets/scripts/client/InputController.js
erikquinn:
```javascript
    /**
     * Triggered when a user clicks on the `right` mouse button and
     * records the position of the `right click` event.
     *
     * @param event {jquery Event}
     * @private
     */
    _onRightMousePress(event) {
        if (MeasureTool.isMeasuring) {
            this._removePreviousMeasurePoint();

            return;
        }

        // copy mouse click position to clipboard on shift+alt+rightclick
        if (event.originalEvent.shiftKey && event.originalEvent.altKey) {
            const relativePosition = this._calculateRelativePositionFromEvent(event);
            const referencePosition = AirportController.current.positionModel;
            const latLonCoordinates = DynamicPositionModel.calculateGpsCoordinatesFromRelativePosition(
                relativePosition, referencePosition
            );

            return this._logAndCopyCoordinates(latLonCoordinates);
        }

        this._markMousePressed(event, MOUSE_BUTTON_NAMES.RIGHT);
    }

    /**
     * Log the provided lat/lon coordinates to the console, display in command log, and copy to clipboard
     *
     * @for InputController
     * @method _logAndCopyCoordinates
     * @param latLonCoordinates {array<number>} [lat, lon]
     * @returns undefined
     * @private
     */
    _logAndCopyCoordinates(latLonCoordinates) {
        const coordinateText = latLonCoordinates.map((coord) => coord.toFixed(9)).join(', ');

        window.navigator.clipboard.writeText(coordinateText).then(() => {
            console.log(coordinateText);
            UiController.ui_log(`Clicked coordinates: ${coordinateText} (logged to console and copied to clipboard!)`, true);
        });
    }
```

PR #1680, opened 2020-11-18
bugfix/1190 - re-add .notSelectable css class

### LINKED ISSUE #1190: notSelectable class doesn't do anything
Somewhere along the way, the definition of the `notSelectable` class got removed. It's still trying to be used to prevent us from highlighting text that isn't meant to be highlighted (such as the clock, tutorial text, etc).

According to [stack overflow](https://stackoverflow.com/a/4407335), it should be able to work on all modern browsers via the below:
```
.notSelectable {
  -webkit-touch-callout: none; /* iOS Safari */
    -webkit-user-select: none; /* Safari */
     -khtml-user-select: none; /* Konqueror HTML */
       -moz-user-select: none; /* Firefox */
        -ms-user-select: none; /* Internet Explorer/Edge */
            user-select: none; /* Non-prefixed version, currently
                                  supported by Chrome and Opera */
}
```

spectreofmyself:
Resolves #1190

The purpose of this pull request is to re-add functionality to the `.notSelectable` class

@erikquinn thanks for your help on my last PR - I couldn't see any test cases for this, so I presume there's no need to add those for CSS, haha. If I'm missing anything please just let me know

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### Bugfixes
 - <a href="https://github.com/openscope/openscope/issues/1565" target="_blank">#1565</a> - Prevent aircraft from taxiing to a runway they're already holding short of
-- <a href="https://github.com/openscope/openscope/issues/1190" target="_blank">#1190</a> - Fix .notSelectable CSS class functionality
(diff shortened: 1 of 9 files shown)


n8rzz (approved):

erikquinn (approved):
Good one! Always thought we had this but I guess maybe not? Works great now though.

Thanks! :)

PR #1698, opened 2021-01-09
refactor/1697 introduce parse function which returns ParsedCommand

### LINKED ISSUE #1697: Introduce parse function which returns ParsedCommand
This has two advantages:

1. By renaming, the call site in `InputController.js` gets more readable (see the PR)
2. This keeps the diff low as

in  a following PR, I want to split  `ParsedCommand` into two parts (one which holds the data used in the `InputController` and one Parser class which does the parsing/validating). By this, we will have a clear processing pipeline of reading from the CLI -> parsing -> Error or Command returned.

--- 
forget it
I just noticed I can do it in one PR by introducing a `parse` function which returns a `parsedCommand`

JonasCir:
Closes #1697 
By this, the call site in `InputController.js` gets more readable. 
Consider this PR as a trick to reduce the diff of the following PR. Once this is merged, I can start to separate the parsing process from the command that is returned and handled.


JonasCir:
I just had an idea...

n8rzz (commented):
I’m not sure I understand the name change here. `CommandParser` describes and action. It does something that the name describes. `ParsedCommand`, however, suggests a thing has already been done. This name describes an immutable thing. 

Perhaps that’s the direction?  I want to call it out now, tho, cause naming is incredibly important.

JonasCir:
@n8rzz Your totally right, I updated the PR now such that there is an actual `parse` method in the parser which returns the `ParsedCommand`. This should be more intuitive and provides us with a clean interface.

JonasCir:
This is better :) The parser logic is now separated from the command that is returned.

test/commands/ParsedComman.spec.js
```diff
@@ -0,0 +1,60 @@
+import _isEqual from 'lodash/isEqual';
```
n8rzz, line 1: missing a `d` in the filename 😉

JonasCir: good eye :)

src/assets/scripts/client/InputController.js
```diff
@@ -778,29 +778,29 @@ export default class InputController {
...
+            cmd = parser.parse();
         } catch (error) {
             UiController.ui_log('Command not understood', true);
-
             throw error;
         }
-
-        if (aircraftCommandParser.command !== PARSED_COMMAND_NAME.TRANSMIT) {
-            return this.processSystemCommand(aircraftCommandParser);
+        const parsedCommand = cmd;
```
n8rzz, line 795: Why are we reassigning here?  Could we continue to use `cmd`?

test/commands/ParsedCommand.spec.js
```diff
@@ -0,0 +1,60 @@
+import _isEqual from 'lodash/isEqual';
+import ava from 'ava';
+import CommandParser from '../../src/assets/scripts/client/commands/parsers/CommandParser';
+
+
```
n8rzz, line 5: Please use a single line for spacing. Here and below too

n8rzz (changes requested):
Just a couple small things and you’re good to go.

src/assets/scripts/client/InputController.js
JonasCir: the reason is that I want the parsed command to be `const`, but I cannot do it in the try block as it is then only scoped the block. To make it `const`, I need to trick and use a reassign here.

test/commands/ParsedCommand.spec.js
JonasCir: Done :)

n8rzz (approved):

PR #1701, opened 2021-01-10
documentation/1699 - update+refine aircraft command documentation

### LINKED ISSUE #1699: Aircraft command documentation updates + refine wording
Browser: n/a

Airport: n/a

#### Issue Description
I've reviewed the aircraft commands documentation and in-app tutorial text for any required updates and other refinements. Documenting suggested changes here, input/discussion welcomed.

##### `aircraft-commands.md`
- update table of contents to remove "Move Data Block" which is a scope command
- `takeoff` - "climb to the altitude specified" implies that altitude is specified as an argument to this command, clarify by rewording as "... altitude they were *previously cleared to*"
- `taxi` - "hold short of *the* specified runway"
- `ils` - '*' is an alias, not a hotkey, i.e. it is accepted as valid input by the parser
- `fix` - add missing empty line to start new para for "deprecated" text. compare `abort`.
- `hold`
  - add documentation for distance-based holds, added in #1459
  - `l`,`r`,`left`,`right` are all acceptable for direction argument
  - holding over present position if no fix provided is intended behavior but doesn't actually work (#873 / #1318), document this
  - remove mention of commands that implicitly clear the hold #1333, in favor of explicit `exithold` command
- `exithold` - clarify behavior when no specific fix is given. see #1276, #1333
  - at some point I hope to revisit what the intended behavior ought to be and implement accordingly, but for now correctly document the actual behavior
- `altitude` - also mention that acft will not descend below MSA
- `heading`
  - 'shortkey' -> 'hotkey'
  - `l`,`r`,`left`,`right` are all acceptable for direction argument
  - document "turn by specified angle" with one- and two- digit angle specifying heading change
- `speed`
  - '+' and '-' are aliases, not hotkeys
  - add clarification that this is airspeed which will differ from ground speed 

##### tutorial text in `TutorialView.js`
ought to be json-ized, but that's a different issue.

- basic control instructions: radar vectors - make clearer distinction between how to specify target heading vs desired heading change
  - "Usually, you will specify a three-digit heading using..."
  - "instruct an aircraft to turn left *or* right by a given number of degrees if you specify a one- or two-digit number."

#### Steps to reproduce
n/a

cake-pie:
Resolves #1699

The purpose of this pull request is to bring aircraft command documentation up-to-date with various changes, clarify details about command parameters, better document discrepancies between intended vs actual implemented behavior, and other small refinements to wording.

### DIFF
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -54,7 +54,6 @@
 [System Commands](#system-commands)
 
 - [Airport](#airport)
-- [Move Data Block](#move-data-block)
 - [Pause](#pause)
 - [Timewarp](#timewarp)
 - [Tutorial](#tutorial)
@@ -96,9 +95,9 @@ _Aliases -_ `takeoff`, `to`, `cto`
 _Hotkey -_ `numpad /`
 
 _Information -_ This command clears the specified plane for takeoff. They
-will climb to the altitude specified, or in accordance with a SID if told
-previously to "climb via the SID". If neither has happened, they will ask
-for an altitude assignment before they agree to take off.
+will climb to the altitude they were previously cleared to, or in accordance
+with a SID if told previously to "climb via the SID". If neither has happened,
+they will ask for an altitude assignment before they agree to take off.
 
 _Syntax -_ `AAL123 cto`
 
@@ -107,7 +106,7 @@ _Syntax -_ `AAL123 cto`
 _Aliases -_ `taxi` / `wait` / `w`
 
 _Information -_ This command tells the specified plane to taxi to and hold short
-of specified runway.
+of the specified runway.
 
 _Syntax -_ `AAL123 taxi [runway]`
 
@@ -142,9 +141,7 @@ _Syntax -_ `AAL123 dvs` or `AAL123 dvs [alt]`
 
 ### ILS
 
-_Aliases -_ `ils` / `i`
-
-_Shortkey -_ `numpad *`
+_Aliases -_ `ils` / `i` / `*`
 
 _Information -_ This command clears for an ILS approach to a runway. The
 aircraft will continue on its assigned heading until intercepting the localizer,
@@ -164,6 +161,7 @@ These commands allow you to manipulate the route in the aircraft's FMS.
 ~~_Aliases -_ `f` / `fix` / `track`~~
 
 ~~_Syntax -_ `AAL123 f [fixname]`~~
+
 *_This command has been deprecated_*
 
 ### Hold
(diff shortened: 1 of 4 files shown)


n8rzz (approved):

documentation/aircraft-commands.md
```diff
@@ -142,9 +141,7 @@ _Syntax -_ `AAL123 dvs` or `AAL123 dvs [alt]`
 
 ### ILS
 
-_Aliases -_ `ils` / `i`
-
-_Shortkey -_ `numpad *`
+_Aliases -_ `ils` / `i` / `*`
```
erikquinn, line 144: WIth this and other cases where shortkeys/hotkeys are being moved to an alias, we're losing context on some purposeful functionality. The shortkey/hotkey (same meaning iirc, just not all using the same word) implies that you can press _that key_ to insert the command _with the appropriate spaces surrounding it_.

So in this case, while from the parser's point of view, `*` is simply an alias for `ils`, from the user's point of view they have the following options to utilize this command:
- type `ils` followed by a space and the runway number
- type `i` followed by a space and the runway number
- press the `*` key on the keyboard, followed by the runway number (with no space separating them)

The hotkeys also have the advantage of being able to be chained without pressing the space bar. For example, the following key strokes `[right arrow] 2 5 0 [down arrow] 3 0 * 2 8 R` would yield `t r 250 d 30 * 28R` in the command bar without ever touching the space key.

So somehow, we do need to differentiate by which keys perform this behavior and which are simply characters that need to be entered in the command bar for it to be understood by the parser.

documentation/aircraft-commands.md
```diff
@@ -164,28 +161,36 @@ These commands allow you to manipulate the route in the aircraft's FMS.
...
+Optionally, you may also specify:
+
+- the direction of turns during the hold with `left`, `l`, `right`, or `r`
+- the leg length, either as `[time]min` or `[distance]nm`
+- the `radial` (a 3-digit course, eg. 001 to 360) which defines the outbound leg.
+
+Any combination of these arguments provided in any order is acceptable, as long
+as the command `hold` comes first.
+
+_Syntax -_ `AAL123 hold [fixname] [left|l|right|r] [leg_time]min|[leg_dist]nm [radial]` ~~or `AAL123 hold`~~
```
erikquinn, line 193: Please add this sentence back-- I know it is not specific to this command, but is important since the `exitHold` command is not the only way to do so, and may not even be working.

>To clear the aircraft out of the hold, you can use the `exitHold` command below, clear it direct to a fix, or assign it a new heading.

documentation/aircraft-commands.md
```diff
@@ -319,27 +329,32 @@ _Syntax -_ `AAL123 fph`
...
 
-_Syntax -_ `AAL123 fh [hdg]` or `AAL123 (rightarrow) [hdg]` or `AAL123 t r [hdg]`
+The target heading is specified by a three-digit heading. You can force the
+aircraft to reach the heading by turning left or right by inserting `left`,
+`l`, `right`, or `r` before the new heading, as demonstrated below.
 
-### Speed
+Alternatively, you can instruct an aircraft to turn from its current heading by
+a given angle by specifying the direction of turn with the heading change as
+a one- or two-digit number of degrees
```
erikquinn, line 345: Add period and drop hyphens:
`a one or two digit number of degrees.`

documentation/aircraft-commands.md
```diff
@@ -319,27 +329,32 @@ _Syntax -_ `AAL123 fph`
...
 
-_Aliases -_ `speed` / `slow` / `sp`
+_Syntax -_ `AAL123 fh [hdg]` or `AAL123 (rightarrow) [hdg]` or `AAL123 t r 270` (target heading) or `AAL123 t r 45` (heading change)
+
+### Speed
 
-_ShortKey -_ `numpad +` / `numpad -`
+_Aliases -_ `speed` / `slow` / `sp` / `+` / `-`
```
erikquinn, line 351: +/- are hotkeys

documentation/aircraft-commands.md
```diff
@@ -319,27 +329,32 @@ _Syntax -_ `AAL123 fph`
...
 
 _Information -_ This command sets the target speed; aircraft will stay within
 their safe speeds if you tell them to fly faster or slower than they are able
 to. It takes some time to increase and reduce speed. Remember that speed is
-always expressed in knots.
+always expressed in knots, and is indicated airspeed, whereas our radar scope
+can only display groundspeed.
```
erikquinn, line 357: Good change. In fact, knots is super obvious (and beside the point), so let's simplify to:
`Remember that speed assignments are given in indicated airspeed, whereas our scope only displays groundspeeds.`

documentation/scope-commands.md
```diff
@@ -4,29 +4,53 @@ Below is a detailed list outlining all of the available scope commands, includin
 
 Other than moving the data block or accepting handoffs, most commands require that the radar target be under your scope's control.
 
-Please note that the bracketed words (and brackets themselves) used in the syntax examples should be replaced with the appropriate data. Also note that `CID` was used in the examples, but aircraft may be referenced with any of ~~(a) the CID~~, (b) the full callsign, or ~~(c) the squawk code~~.
+Please note that the bracketed words (and brackets themselves) used in the syntax examples should be replaced with the appropriate data. Note that in lieu of having to manually type the `callsign` into the command input, it is also possible to click on the aircraft on the radar scope; doing so will automatically append its callsign to the command input and execute the completed command.
+
+While `callsign` is used in the examples, it is envisioned that in future it will be possible for  aircraft to also be referenced by their CID or squawk code.
```
erikquinn, line 9: - Double space between `for` and `aircraft`.
- it currently works with a click, the full callsign, or the squawk code. Only the CID is not yet implemented for this.

erikquinn (changes requested):
Very nice! Few small changes and I say :shipit:

documentation/aircraft-commands.md
cake-pie: Ah, right. It's worth highlighting that behavior of *numpad*  `+`, `-` and `*`.

On the other hand I'd wanted to make it explicit that inputting them using non-numpad keys will also be acceptable to the parser -- in contrast with, say, the arrow keys, which when pressed are substituted by the corresponding actual commands.

So... both?

documentation/aircraft-commands.md
cake-pie: done

documentation/aircraft-commands.md
cake-pie: done

documentation/aircraft-commands.md
cake-pie: done

documentation/scope-commands.md
cake-pie: fixed

erikquinn (approved):

PR #1708, opened 2021-01-13
bugfix/1707 - fix various minor keyboard input bugs / UX issues

### LINKED ISSUE #1707: Various minor keyboard input bugs / UX issues
A handful of minor things I noticed in `InputController.js` related to keyboard inputs and controls; bundling because changes are concentrated in this one file; also I think a couple of them would conflict if put in separate PRs so this seems more painless.

---
#### Issue Description: input history navigation bug

PgUp/PgDown can be used to cycle (prev/next) through aircraft according to history of input commands.

UpArrow/DownArrow -- in classic control method setting -- seems intended to have the same functionality, but both are mapped to "previous" and thus cycling does not function correctly

#### Steps to reproduce
- Pick any airport, issue some commands to different aircraft.
- From empty command input, use PgUp/PgDown to cycle through history of aircraft callsigns, note being able to navigate forward and back through history.
- From empty command input, try using UpArrow/DownArrow to do the same, notice both only go further back in history.

--- 
#### Issue Description: aircraft command hotkeys not confined to aircraft input mode

causes specific UX issue in the case of `numpad /`.
It is set up as a hotkey that instead appends ` takeoff `, and this behavior still occurs if the input is in scope mode, which prevents that key from being used for the move data block command. Instead of being able to quickly punch in numbers and `/` on the numpad alone, users must use the other `/ ?` key on the keyboard.

#### Steps to reproduce
- In scope mode, try entering move data block command using numpad alone.

--- 
#### Issue Description: add `/` as alias for takeoff

a)
`+`, `-`, `*` are all acceptable aliases for their respective commands (speed, ils). `/` is the odd one out.

Use case: quickly amend a typed command using the non-numpad version of the key.  

e.g. Say you're using hotkeys and you fatfingered `*` instead of `-` while typing, and the input looks like `AAL123 * 200 a 32`. You can relatively quickly navigate back (ctrl+left arrow) and fix using the `- _` key (two to the left of backspace). Notice that using `numpad -` does not work here because it will still *append* ` - ` to the end of the partially typed command and you'd get `AAL123 * 200 a 32 - ` instead.

Whereas for example say you've fatfingered `*` instead of `/` and are faced with `UAL456 * cvs + 350`. It is not possible to amend using `/ ?`; shortest alternative is to use `to` instead.

b)
Arrow keys as hotkeys (with arrow keys control mode) substitute with the shortest one-character aliases for their respective commands, `c`, `d`, `t l` and `t r`. `/` is also unusual in this respect for substituting with the full `takeoff` command.

cake-pie:
Resolves #1707

The purpose of this pull request is to fix these minor issues:
- down arrow for input command history navigation
- restrict aircraft command hotkeys behavior to aircraft mode, no longer affect scope mode
- make `/` an alias for `takeoff`, for user convenience and consistency
  - command documentation updated accordingly

Follow-up on #1699
While updating the aircraft command documentation I also just noticed that there was inconsistent use of both slash and comma to delimit different aliases or hotkeys for the same command. Changed to commas across the board for consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Bugfixes
 - <a href="https://github.com/openscope/openscope/issues/1280" target="_blank">#1280</a> - Remove redundant timeout-based TAXI->WAITING mechanism 
 - <a href="https://github.com/openscope/openscope/issues/1705" target="_blank">#1705</a> - Re-fix range ring drawing position
+- <a href="https://github.com/openscope/openscope/issues/1707" target="_blank">#1707</a> - Fix minor keyboard input inconsistencies and correct command documentation
 
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1703" target="_blank">#1703</a> - Add more PTL ranges, some including 30sec PTLs
--- a/documentation/aircraft-commands.md
+++ b/documentation/aircraft-commands.md
@@ -103,7 +103,7 @@ _Syntax -_ `AAL123 cto`
 
 ### Taxi
 
-_Aliases -_ `taxi` / `wait` / `w`
+_Aliases -_ `taxi`, `wait`, `w`
 
 _Information -_ This command tells the specified plane to taxi to and hold short
 of the specified runway.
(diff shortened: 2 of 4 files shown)


3tractorrides (approved):

cake-pie:
@n8rzz wrong account?

n8rzz:
yes, you caught me ☹️

n8rzz (approved):

PR #1710, opened 2021-01-13
refactor/1700 - jsonize tutorial

### LINKED ISSUE #1700: JSON-ize tutorial
Tutorial steps and text should be refactored out of `TutorialView.js` into a JSON file so that changes to its contents do not require modifying javascript sources.

### LINKED ISSUE #1709: Regression: tutorial using "values from current sim state"
#### Issue Description
While working on #1700 I noticed a regression in #634 that #1159 attempted to address, but the issue was mis-characterized and the fix not quite complete.

The tutorial code was designed to be able to pull values from the *current* sim state and include them in the tutorial text. The way it was written, initialization `tutorial_init_pre()` in `TutorialView.js` was only intended to be invoked once on app load. The key piece of machinery is actually the `parse` property of `TutorialStep` which stores a function that fetches values from sim state and substitute them at the time when [`getText()`](https://github.com/openscope/openscope/blob/f00fdf8a5b83bd6ce11e0b8aa886570eab2ee6a3/src/assets/scripts/client/ui/TutorialStep.js#L27-L41) is actually executed.

Specifically, it was intended for the tutorial to pick the oldest departure aircraft (bottom-most in strip view) as the example departure aircraft and guide the user in getting it in the air and on its way.  Regardless of changing airports or progress in the sim, if and when the user views the tutorial and the relevant step is displayed, `getText()` causes the function `parse` to be re-evaluated and that should fetch values from the sim *at that very point in time*, guaranteeing a valid example is used.

A flaw in the implementation led to #634 which inadvertently misdiagnosed "repetitive code" and made [this change](https://github.com/openscope/openscope/pull/634/files#diff-1e72740d59ab6d28c14e3de2e1707589fdc16ad88797cca915e0a33e06f1a2b9R196) introducing a new variable `departureAircraft` to "reduce redundant code". This had the effect of making the example departure aircraft permanently point at the very first departure spawned for the initial airport when the sim first loaded.

#1159 dealt with the most obvious symptom: after changing airports, the example departure refers to a callsign etc that no longer exists. By re-doing initialization (`tutorial_init_pre()` in `TutorialView.js`) when airport is changed, the example departure becomes permanently stuck as the first spawned departure for the current airport. Once that aircraft has been cleared to take off, and departed and despawned, we're back in the situation where tutorial refers to a nonexistent aircraft.

A case can be made that this is not a huge problem: the user went through the tutorial and now knows how to control departures, so the example is no longer needed. On the other hand if this is how you want it, then the whole complex `parse` and `getText()` machinery is rendered pointless and unnecessary -- the text replacement could be done within `tutorial_init_pre()` and then just store the resulting text for all subsequent use.

cake-pie:
Resolves #1700, resolves #1709

The purpose of this pull request is to refactor the tutorial, moving its contents to JSON so that any future updates / changes can (mostly) be done without modifying *.js sources.

Incidentally also fixes a regression of not correctly fetching *current* values from the sim for use in tutorial examples.

I've made an effort to avoid `eval()` and its ilk, but there is one `new Function()`constructor [here](https://github.com/openscope/openscope/blob/37ae9d246751ec984003e9e53e2c4b718dfe64f3/src/assets/scripts/client/ui/TutorialView.js#L269) to support arbitrary transformations to the values pulled from the sim -- needed to retain [this uppercasing](https://github.com/openscope/openscope/blob/37ae9d246751ec984003e9e53e2c4b718dfe64f3/assets/tutorial/tutorial.json#L43), used for nothing else. If anyone has ideas on how to handle this, I'm all ears.

Todo:

- [x] remove old code for populating tutorial -- retained to allow comparison for now
- [x] remove #1159 code that re-initializes tutorial at every airport change, should not be needed with improved fix
   - [x] verify example callsign changed on airport change
   - [x] verify example callsign changed after original example despawned
- [x] changelog


src/assets/scripts/client/ui/TutorialView.js
```diff
@@ -240,9 +243,64 @@ export default class TutorialView {
...
+            },
+            departureAircraft: function() {
+                return prop.aircraft.list.filter((aircraftModel) => aircraftModel.isDeparture())[0];
+            }
+        };
+
+        zlsa.atc.loadAsset({ url: 'assets/tutorial/tutorial.json', immediate: true })
+            .done((response) => {
+                response.forEach((step) => {
+                    if (Array.isArray(step.replace)) {
```
n8rzz, line 259: could we abstract this logic into a dedicated helper or service?  there is quite a bit here that isn't dedicated to the TutorialView, but parsing json.

src/assets/scripts/client/ui/TutorialView.js
```diff
@@ -509,6 +567,28 @@ export default class TutorialView {
...
+     * @for TutorialView
+     * @method _getPropFetcher
+     * @return [{function}] A function that will fetch a given descendent property of an object without using eval()
+     */
+    _getPropFetcher(desc) {
+        return ((obj) => {
+            let result = obj;
+            const arr = desc.split('.');
+            while (arr.length) {
+                if (result === undefined || result === null) return result;
```
n8rzz, line 587: * please use block `if` statements
* you could use loose equality here to so we only need one check:
```js
if (result == null) { 
  
}
```

n8rzz (changes requested):

src/assets/scripts/client/ui/TutorialView.js
cake-pie: moved handling of each step into a separate function.

src/assets/scripts/client/ui/TutorialView.js
cake-pie: done

cake-pie:
Actually, hang on. Want to remove the `new Function()` if it is not truly necessary.

Why is aircraft type ICAO lowercased when instantiating from the json:

https://github.com/openscope/openscope/blob/67f23ecfa27e73eb698ee7fe7c008ac03fdb8d3e/src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js#L154

and then converted them back to uppercase for presentation (for strip view in this excerpt):

https://github.com/openscope/openscope/blob/67f23ecfa27e73eb698ee7fe7c008ac03fdb8d3e/src/assets/scripts/client/aircraft/AircraftTypeDefinitionModel.js#L212

It's naturally an uppercase-only string; couldn't we just consistently handle it in uppercase throughout the code?  
(Same can probably be said for a number of other properties.)

cake-pie:
Removed support for arbitrary code, instead enforcing `AircraftTypeDefinitionModel.icao` to consistently be handled as uppercase values.

Strangely, now failing this test, which I don't think should have been affected? :shrug:

```
  aircraft › AircraftTypeDefinitionModel › throws when passed invalid parameters

  /home/runner/work/openscope/openscope/test/aircraft/AircraftTypeDefinitionModel.spec.js:10

    9:     t.throws(() => new AircraftTypeDefinitionModel({}));       
   10:     t.throws(() => new AircraftTypeDefinitionModel(42));       
   11:     t.throws(() => new AircraftTypeDefinitionModel('threeve'));

  Function returned:

  AircraftTypeDefinitionModel {
    _id: 'BaseModel-701',
    capability: undefined,
    category: undefined,
    ceiling: undefined,
    engines: undefined,
    icao: undefined,
    icaoWithWeightClass: 'undefined/L',
    name: undefined,
    rate: undefined,
    runway: undefined,
    speed: undefined,
    weightClass: undefined,
  }
```

cake-pie:
Test failure unrelated. #1744 has been filed for that.

n8rzz (approved):

cake-pie:
Strongly recommend merging #1745 at the same time to prevent spurious test failures for other PRs.

PR #1713, opened 2021-01-15
refactor/1712 centralize z-index declarations for ease of layer management

### LINKED ISSUE #1712: css/less : centralize z-index declarations for ease of layer management
#### Issue Description
While tinkering on a command autocomplete feature for #21 I needed to find out what `z-index` would be required to prevent canvas elements such as range rings from overlapping on top of my UI elements.

The lazy thing to do would be to set arbitrarily big `z-index` but it's better to do this in a coordinated way.

Indeed there are currently [12 different](https://github.com/openscope/openscope/search?q=z-index) `*.less` files declaring `z-index` with a wide range of values from 1, 2 to 100, 125, 200, 500 and even 1000, 1500.

This is hard to navigate and make sense of.  
For ease of layer management, instead have a section in `_vars.less` with something like:

```
// layer management
@static-canvas-z: 1;
@dynamic-canvas-z: 2;
@log-z: 5;
@airport-info-z: 5;
@strip-view-z: 10;
@controls-z: 20;
@dialogbox-z: 50; // shared
@pause-z: 99;
@loading-z: 100;
```

Much easier to understand what's going on.  
Respective elements get their z-index value via these variables.

cake-pie:
Resolves #1712

The purpose of this pull request is to centralize z-index declarations for ease of layer management


n8rzz (approved):
☺️

erikquinn (approved):

PR #1730, opened 2021-01-23
Enhancement/1729 - Add Helicopters to KDCA

### LINKED ISSUE #1729: Add Helicopters to DCA
Helicopters are a massive part of DCA tower's job- Coast Guard helicopters are based at DCA, MEDEVAC and law enforcement helicopters are constantly buzzing around DC, and the most iconic of the DC helicopters, Marine One, regularly flies through DCA's airspace. This issue will primarily address helicopters in the DC area, but I fully intend to add them to other airports- New York, LAX, and many others are all candidates. It's not the primary feature of the sim, but in my opinion it's really important to make this sim as realistic as possible- adding helicopters will only add realism.

Hawaiicoder:
Resolves #1729 

The purpose of this pull request is to add helicopters to KDCA- they're a massive part of DCA tower's job and will improve realism.

### DIFF
--- a/assets/aircraft/AS50.json
+++ b/assets/aircraft/AS50.json
@@ -1,0 +1,38 @@
+{
+    "name": "Airbus Helicopters AS350",   
+    "icao": "AS50",             
+    "engines": {
+        "number": 1,            
+        "type": "T"             
+    },
+    "weightClass": "S",         
+    "category": {
+        "srs": 1,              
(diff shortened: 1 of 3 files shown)


erikquinn:
This is a part of the local controller's responsibilities, which is something we're trying to phase out anyway in favor of focusing on improving the realism of the approach simulation.

Closing.

Issue #1718, opened 2021-01-17
Difference between aircraftCommandMap and aircraftCommandDefinitions


JonasCir:
#### Issue Description
While writing tests for `aircraftCommandMap.js` and `aircraftCommandDefinitions.js`, I noticed that both files are not in sync.

1. `transmit` is a system command, but has no parser or validators associated.
2. `debug` has a parser and a validator, but not entry in the command map
3. ` has its own parser and validator (see [here](https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandDefinitions.js#L138)) but I think this should be an alias for something
4. `sayRoute` is a command, but has no parser or validator associated.
5. There are a couple of commands where I'm not sure that the command handlers are implemented: `airport`, `debug`, `auto`, `clear`, `transmit`, `tutorial`.

### How to resolve this
Bring the two files in sync. Remove or add entries.

JonasCir:
Ok to get the discussion started:

1. Remove `transmit`
2. Remove `debug`
3. move ` to its associated command as alias (not certain which one it is)
4. add `zeroArgumentsValidator` and `noop` parser to `sayRoute`
5. PARSED_COMMAND_NAME is switched to run the code of `airport`,  `auto`, `clear`, `transmit`, `tutorial` etc. This is inconsistent and I think we should use `functionName` as we do normally, but I will address this later in time

What do you think @erikquinn ? :)

erikquinn:
1. I think there may be something special with that one, but can't remember off the top of my head. @n8rzz?
2. Provided it's not used, that's fine
3. Just remove it... used to be the old way to move data blocks (before scope commands existed)
4. yeap!
5. Also defer to @n8rzz, I don't remember how that all worked in detail, and haven't really done any work on this area.

n8rzz:
I'm interested to see what you come up with here.  `transmit` and `debug` are _likely_ holdovers from the original implementation and were things kept in place to reduce the amount of change when we rebuilt the command parser.

You have fresh eyes and are able to look at the current state of things and ask why.  Please keep doing that.  If it looks like it doesn't belong, it might not.

erikquinn:
I agree. As a general sentiment, if you can make things make more sense, go for it. :)

JonasCir:
Alright, thank you for the feedback, will do :)

PR #1738, opened 2021-01-29
feature/21 - Command search and autocomplete with argument validation

### LINKED ISSUE #21: Command Reference

cake-pie:
Resolves #21

The purpose of this pull request is to add a feature for command search and autocomplete with argument validation.

Status is still draft, but should deploy as a workable demo

Things to note / try playing with:
- use `tab` to activate -- it is the standard key for autocomplete in many command line contexts.
- key for toggling between aircraft mode and scope mode has been changed to ```` ` ````, which is reminiscent of [toggling console in PC games](https://en.wikipedia.org/wiki/Console_(computer_games)) (#1718 confirms that the key is freed up)
- depending on where cursor/selection is within command input upon activation, some contents of command input may be ingested into autocomplete as initial input, and will be replaced upon autocomplete commit.
- will intelligently figure out if it should be trying to search/autocomplete system commands or aircraft commands
- supports searching e.g. "skip" will suggest `pd`, check `altkeys` in json file for more
- command suggestions can also be clicked


cake-pie:
:rocket: Deployment is up and working as expected, keen for folks to take it for a spin!

cake-pie:
@erikquinn @n8rzz any feedback on this?

n8rzz (commented):
Thanks for the ping.  I hadn't looked cause this is still listed as a draft PR.

The code here looks pretty good. One general thing I'll call out: Try to avoid specificity in the css selectors.  Something like `#autocomplete-suggests td.autocomplete-command-suggestion` could expressed as simply `.autocomplete-command-suggestion`.

I'll have some style tweaks for you a bit later. I'm going to need a little bit of time to write those up, as I"m just starting my work day at the moment. Probably tonight, but ping me if it gets to be a day or two.

cake-pie:
@n8rzz I was more concerned about whether the *behavior* is in line with what was requested / expected -- wanted some indication that things are on the right track before investing into clean up and polish.

But hey, if you're set to dive right into code review, I'll take it.

> Try to avoid specificity in the css selectors.

Yeah, now that you mention it, some of that may be too defensive.
Kind of a knee jerk reaction from having to troubleshoot problems related to [other `.active` class](https://github.com/openscope/openscope/pull/1738/commits/ed6077bfeefa7671be1ae1284efd06abe9f0be52) not being specific enough.

> need a little bit of time to write those up,

no problem, tia!

erikquinn:
@cake-pie This is so wild! Very impressive stuff!

I see no issues with the functionality that I'd want to change. If I could be so bold to ask though, would it be feasible to get it to bring up the argument portion when a command is entered _without_ using the autocomplete? For example, "how do I format holds again?"... `AAL123 hold `, then press tab, and it would show you this portion:

image

I think _that_ may even be the most valuable use case. Even without it though, this is worth a lot, so thanks a lot for your work on it 😄

n8rzz:
Sorry for the delay on this.  Here are those style feedback items I was talking about earlier this week:

Screen Shot 2021-02-20 at 5 43 20 PM

1. update `.autocomplete-legend` padding to `10px 0`
2a. update `.autocomplete-popup` padding to `0 5px`
2b. I'm curious, why did you go with `table` here and not `flex` `space-between`?
3a. update `.autocomplete-input` to have a `width` of `100%`
3.b I'd love to also see a `label` element here is `.u-isVisuallyHidden` attached to it.  But since accessibility hasn't really been a thing, I'll be fine without it.

cake-pie:
> would it be feasible to get it to bring up the argument portion when a command is entered _without_ using the autocomplete?

There are a couple of problematic aspects with this request -- design decisions at play here, explained further below. (Worth documenting for future reference.)

I'll look into alleviating your use case by other methods, no promises though.

---

### What gets ingested from command input into autocomplete?

> For example, "how do I format holds again?"... `AAL123 hold `, then press tab

This already doesn't work the way you thought it would, because you have a space after `hold`.

Consider that a user might be chaining commands. It's entirely possible the user has typed `AAL123 hold `&nbsp;with the intent of "default hold at current position" and wants to chain an additional instruction after that, but forgot the command and needs to look it up.

Therefore `AAL123 hold `&nbsp;+ `tab` will result in autocomplete activating with empty initial input for a new command, with the intent of appending the resulting autocompleted command after `AAL123 hold `.

Whereas `AAL123 hold` + `tab` (no space!) will cause the autocomplete to ingest `hold` and then replace it with the full autocompleted result; this is the correct starting point for what you actually want.

Keep in mind that there are also other use cases beyond just 'append at the end of the pre-existing command input string'.

Perhaps a user has typed `AAL123 hold expect 22R` and then changes their mind and wants to add arguments to `hold` but needs help with the exact argument format?

In general, when the autocomplete is activated using `tab`, the command input element might either have *any* portion of its input string selected (defined by left and right boundary), or have a cursor (= a zero-length selection) at *any* position within its input string.

Need to be flexible enough to handle any of those situations in a sensible way, while supporting a variety of use cases.
This involves a bit of consideration re: given a cursor/selection positioning, what use case does it map to?

For simplicity, I've chosen to simply treat the command input string simply as tokens separated by whitespace. No parsing or inspection of the tokens will occur, that's a rabbit hole I don't want to go down. Instead, I will simply consider only the positioning of cursor/selection relative to tokens.

The designed behavior for input ingestion can then be described as follows:
- if any input is ingested, it will be replaced by the eventual output
- a cursor not touching any token
  - autocomplete input initially empty
  - `command arg arg` inserted at cursor position
- a cursor on the right edge of a token `foo bar| baz` 
  - `bar` ingested as initial input of autocomplete
  - becomes `foo command arg arg baz`
- a cursor in the middle of a token `foo ba|r baz`
  - `bar` ingested as initial input of autocomplete
  - becomes `foo command arg arg baz`
- a cursor on the left edge of a token `foo |bar bar`
  - will *not* ingest the token `bar`, autocomplete input initially empty
  - becomes `foo command arg arg bar baz`
  - this facilitates insertion of an additional command
- a selection with left/right boundary in the middle of a token will grow to encompass the whole token
  - `foo |ba|r baz`, `foo b|a|r baz`, `foo b|ar| baz` all result in
    - `bar` ingested as initial input of autocomplete
    - becomes `foo command arg arg baz`
- selections do not grow to encompass additional tokens that they merely touch but do not intersect
  - `foo| ba|r baz`, `foo| bar |baz`, `foo b|ar |baz` all result in
    - `bar` ingested as initial input of autocomplete
    - becomes `foo command arg arg baz`

---

### Skipping command suggestion+selection step and proceed straight to argument validation step

Keep in mind that the command suggestion+selection step is not merely a pure autocomplete (i.e. user input is strictly a prefix of a valid command) but also provides for lookup by alternative search terms, in case the user has forgotten the exact phraseology.

From your perspective, `hold` is an exact match for a command, so you want to skip straight to argument validation.
However, it's also possible that `AAL123 hold` + `tab` might be a user who has an arrival aircraft on short final, wondering "how do I make a departure aircraft taxi to the runway but _**hold short**_ of it?"

Given that some words may be very common in atc phraseology (e.g. `clear`) there will inevitably be the occasional collision where a command or alias also happens to be match for alternative search terms for a different command. Therefore I am hesitant to put in a hard skip, it might surprise/confuse users who were looking for something else.

---

I'm thinking that instead of exactly what you've requested, it might be possible to instead make it such that when in the command suggestion step, when the input is an exact match for a command, to pre-highlight that suggestion entry without having to use the up/down arrow keys or click it with the mouse.

i.e. currently, `AAL123 hold` + `tab` shows two suggestions, with neither highlighted. Instead see if we can make it show the two suggestions, with the option for `hold` already highlighted. It would then take only one more keypress (`tab` again) to go to the argument format guide+validation.

Does that seem like a reasonable compromise?

cake-pie:
> style feedback

Oh, you meant css style, I thought you were referring to code conventions.

> 1. update `.autocomplete-legend` padding to `10px 0`
> 2a. update `.autocomplete-popup` padding to `0 5px`

I see what you are going for, but will do it in a different way than suggested.
(Reason: it is a conscious design decision to have the text in left column of suggestions align exactly with the input below.)

> 2b. I'm curious, why did you go with `table` here and not `flex` `space-between`?

huh? not sure I understand; flex is for layout in one dimension.
I have multiple rows, columns should line up, and columns should size to fit widest content

> 3a. update `.autocomplete-input` to have a `width` of `100%`

It is intentionally not 100%.
I am aware that longer command input underneath it peeks out on the right and will deal with that by other means. (part of intended visual polish)

cake-pie:
> Try to avoid specificity in the css selectors. Something like `#autocomplete-suggests td.autocomplete-command-suggestion` could expressed as simply `.autocomplete-command-suggestion`.

I'd forgotten since it's been a couple weeks since I originally worked on this -- appears that it does have to be this verbose in order to properly override `padding` defined for `#autocomplete-suggests td`

erikquinn:
> I'm thinking that instead of exactly what you've requested, it might be possible to instead make it such that when in the command suggestion step, when the input is an exact match for a command, to pre-highlight that suggestion entry without having to use the up/down arrow keys or click it with the mouse.
>
> i.e. currently, AAL123 hold + tab shows two suggestions, with neither highlighted. Instead see if we can make it show the two suggestions, with the option for hold already highlighted. It would then take only one more keypress (tab again) to go to the argument format guide+validation.
>
> Does that seem like a reasonable compromise?

Absolutely, `AAL123 hold`, tab, tab and you see the parameters is a good solution to the use case. And I agree, only with a perfect match to the command's full name or any of its aliases. So `AAL123 c` tab would also preselect `climb`.

erikquinn (commented):
Defer to @n8rzz.

Only other thing that would be great is to honor an `Enter` at any time during the process. having to tab back _out_ of the autocomplete is not a behavior people would be likely to assume is necessary once they've gotten what they need out of it. Hitting enter should exit the autocomplete and send the command as-is.

Just advise if you wanted to do this now (if not, it can be a later/never thing), and I'll just wait to hear back from Nate on if he is satisfied with the adjustments you made.

cake-pie:
Not to forget other loose ends:

### Todo:
- recommend additional scrutiny of `candidate` and `validate` regexes in [commandAutocompleteConfig.json](https://github.com/openscope/openscope/blob/feature/21/assets/autocomplete/commandAutocompleteConfig.json)
  - role of regexes is described [here](https://github.com/openscope/openscope/blob/feature/21/assets/autocomplete/commandAutocompleteConfig.schema.json#L95-L102). string is `trim()`ed before tested against regex.
  - inviting extra eyeballs from @JonasCir (doing parsing refactor) and @OlaRune (hold args changes)
- [ ] update `commands.md` documentation
- [ ] changelog entry

Re: changelog, we might want to highlight that the key to toggle between transmit/scope command entry has been changed.
Wishful thinking perhaps, but it *might* help to *slightly* reduce the number of surprised users. @erikquinn?

```
#21 - Use backquote (`) to toggle between entering aircraft commands and scope commands
#21 - Add command search and autocomplete with argument validation
```

If a separate atomic issue+PR is required, cherry pick 11a5631572e6576fbcac552ca7c803d7fad9556b

---

> `Enter` at any time during the process / once they've gotten what they need out of it

A user who is using the autocomplete to lookup the keyword + required parameters for a particular command presumably intends to issue that command. They can do so by continuing to input a valid set of arguments for the command while in the autocomplete, and then with one keypress commit the full command+args into the main command input, with the assurance that it is valid (in "shape", not necessarily in value).

If I understand correctly you are asking for one keypress to skip out of the autocomplete UI without having completed the argument validation process, i.e. while there is no green indicator on any combination of params. I guess this would then dump the command together with whatever empty/partial arguments into the command input, and user will still have to finish up typing the full command+args -- potentially making mistakes while doing so -- before transmitting it.

Sounds like simply reversing the order of steps required with no savings in effort, while losing guarantee of validity.
Not sure it's worth implementing additional machinery to support this.

Maybe I'm just not understanding the use case?


> exit the autocomplete and send the command as-is.

Even with valid command+args in the autocomplete, it cannot be assumed that a user necessarily wants to immediately 'transmit' once they exit the autocomplete UI; it is perfectly valid to continue chaining additional commands.

OlaRune:
>   * inviting extra eyeballs from @JonasCir (doing parsing refactor) and @OlaRune (hold args changes)

I haven't looked at the above, but I will.

> #21 - Use backquote (`) to toggle between entering aircraft commands and scope commands

Hmm, regarding this I think it would be good to think a little about different keyboard layouts. I guess your intention is for this to be the key above Tab, but that's not true for many other keyboard layouts. On a Swedish keyboard that character is a bit of an effort to type since it will be shift-"key left of backspace". 
E3B2D344-B8E1-45D0-847C-987247F2C935

I suggest either finding a way to bind it to the specific key instead of the character (maybe that's how it works, then this can be disregarded and explained in the docs instead), or bind it to multiple characters (for example both ` and §, and perhaps another for german keyboards etc.).

cake-pie:
> different keyboard layouts [...] intention is for this to be the key above Tab

We use [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) on [browsers that support it](https://caniuse.com/mdn-api_keyboardevent_code):

https://github.com/openscope/openscope/blob/3ea9e5aa29ad3ca94ad885a6e1a62d9a9023838e/src/assets/scripts/client/InputController.js#L497-L502

This should represent the physical position of the key above `Tab` and below `Esc` regardless of keyboard layout

JonasCir:
> inviting extra eyeballs from @JonasCir (doing parsing refactor)

Thank you for that @cake-pie :) I think I can take a look today or tomorrow. Thank you for you awesome work :rocket:

OlaRune:
> > different keyboard layouts [...] intention is for this to be the key above Tab
> 
> We use [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) on [browsers that support it](https://caniuse.com/mdn-api_keyboardevent_code):
> 
> https://github.com/openscope/openscope/blob/3ea9e5aa29ad3ca94ad885a6e1a62d9a9023838e/src/assets/scripts/client/InputController.js#L497-L502
> 
> This should represent the physical position of the key above `Tab` and below `Esc` regardless of keyboard layout

Perfect! No worries then, just clarify in the docs which key it is so people with other keyboards understand.

erikquinn:
> A user who is using the autocomplete to lookup the keyword + required parameters for a particular command presumably intends to issue that command. They can do so by continuing to input a valid set of arguments for the command while in the autocomplete, and then with one keypress commit the full command+args into the main command input, with the assurance that it is valid (in "shape", not necessarily in value).
>
> If I understand correctly you are asking for one keypress to skip out of the autocomplete UI without having completed the argument validation process, i.e. while there is no green indicator on any combination of params. I guess this would then dump the command together with whatever empty/partial arguments into the command input, and user will still have to finish up typing the full command+args -- potentially making mistakes while doing so -- before transmitting it.

I'm saying the user should be able to press enter while still in the autocomplete to execute the command. Currently, you have to use tab to _exit_ the autocomplete before pressing enter has any effect. Instead, once you have a green indication, you should be able to press enter and have it commit that content to the command bar and send the command.

And also, just to confirm, `commandAutocompleteConfig.json` is automatically generated, right?

erikquinn:
In other words...

> Even with valid command+args in the autocomplete, it cannot be assumed that a user necessarily wants to immediately 'transmit' once they exit the autocomplete UI; it is perfectly valid to continue chaining additional commands.

...yes, it can and should be assumed that they want to immediately transmit the command. There is no need to contextualize the use of the enter key based on what is happening in the command bar. If enter is pressed, command is sent. Use tab to play with autocomplete. Super straightforward. :)

cake-pie:
> Instead, once you have a green indication, you should be able to press enter and have it commit that content to the command bar and send the command.

Ah, I completely misunderstood. You want `enter` while in the autocomplete to serve as shortcut for commit+transmit.

> And also, just to confirm, `commandAutocompleteConfig.json` is automatically generated, right?

No, it's handcrafted, and warrants extra checking -- particularly the more complex regexes such as for `star`, `route` and `reroute`. And then there's `hold` which gets *extra grody* because of the multiple optional args that can appear in any order.

JonasCir:
So I took a brief look (I will look at the details tomorrow):

> key for toggling between aircraft mode and scope mode has been changed to `, which is reminiscent of toggling console in PC games (#1718 confirms that the key is freed up)

This is unfortunately not true, see [here](https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandMap.js#L122) :crying_cat_face: 

My second point is that I plan to unify the [commandDefinitions](https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandDefinitions.js) and [commandMap](https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandMap.js) into **one** file such that there is a single source of truth for the commands in this project. `assets/autocomplete/commandAutocompleteConfig.json` duplicates a lot of information that is already present (and should got into this single file).

I'm super impressed by the regex you came up with! I also thought about doing the parsing with regex, if we manage to put everything into one place, I think I might be able to reuse them and significantly cut down the parsing code in size!

erikquinn:
@JonasCir the linked reference to the moveDataBlock command is only an old reference which displays a message telling you the command has moved to the scope commands, accessible via tab. The command can simply be removed, and there is no conflict. Good callout though as we should remove the old command altogether now.

JonasCir:
@erikquinn ok cool will remove it tomorrow then! :)
https://github.com/openscope/openscope/pull/1792
 @cake-pie my draft PR is not properly cleaned up, but just to make clear what I meant about earlier about merging all the different files...

cake-pie:
@JonasCir:

> single source of truth for the commands in this project.

A worthy goal.

> duplicates a lot of information that is already present (and should got into this single file).

Pre-existing command parsing code was unable to meet the needs of this feature, which is why I had to cook up the autocomplete config json.

Here are some things to keep in mind for any future refactor to integrate the data together.

Autocomplete requirements (vs capabilities of existing command parser):
- ability to have different annotation for different (groups of) aliases for the same command
  - implemented as `variants` which may have different `altkeys` and different `explain` wording or bolded shortcuts
  - see `altitude`, `heading`, `speed` for prime examples, but also `takeoff`, `taxi`, `cancelhold`...
  - parser doesn't care, it just [redirects all aliases to the primary command](https://github.com/openscope/openscope/blob/2883859c35fa97cee21fe5c9fb24d8fcf5fed4cb/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandMap.js#L238-L240)
- ability to have different annotation for different valid combinations of arguments
  - implemented as `paramsets`
  - most commonly explains default behavior of command with single optional argument
  - `heading` (turn) is an example of more complex behaviors requiring different `explain` wording
  - parser doesn't care about this
- ability to validate parameters for each `paramsets` _**separately**_, each `paramsets` has its own `validate` regex
compared to command parser:
  - only some of the commands have their own argument validator in `argumentValidators.js`
  - using a single validator for all possible combinations of arguments, e.g. [heading/turn](https://github.com/openscope/openscope/blob/2883859c35fa97cee21fe5c9fb24d8fcf5fed4cb/src/assets/scripts/client/commands/parsers/argumentValidators.js#L233)
  - many commands just use `singleArgumentValidator`, `zeroOrOneArgumentValidator` etc, only checking number of args
- ability to screen partially typed argument strings for potential (in)validity
  - validity indicator turns red when `candidate` regex no longer matches
  - command parser only cares about complete commands

Some of these requirements lean more towards a "documentation" than "implementation" side of things, so I hesitate to "contaminate" the parser code with it. Also I wasn't really keen to dive into the innards of the parser code, so ultimately I just did what was needed to get the feature off the ground just for a start.

>  if we manage to put everything into one place, [...] significantly cut down the parsing code in size!

Given that the autocomplete requirements are a superset of what basic command parsing needs, it's likely possible to refactor the parser to be driven by the same spec, yes.

> I also thought about doing the parsing with regex, [...] I think I might be able to reuse them

Having built this, I'd suggest that maybe regex isn't the best choice after all.

I'd started off checking viability of the approach by writing regexes for args like runways, and used `cross` as a pilot test for optional arguments in any order. All seemed well at first, but by the time I got to `hold`, `route` and the like it'd become a real struggle vs the limitations of regexes.

Having said that, if the regexes do work correctly then there's nothing to preclude reuse.
You might notice that `validate` regexes even have capturing groups in place already.

cake-pie:
This should make some of the complex regexes a bit less gnarly and easier to comprehend

cake-pie:
Still awaiting review before finalizing with necessary documentation updates and changelog entry.

erikquinn:
If you're awaiting review, _request_ a review. We filter by "review requested", so if you don't re-request it, your LRs will get missed.

Re-requested for you now.

erikquinn (changes requested):
One quick thing I saw just now was the runway regex must be expecting at least 2 digits, but this is not always the case-- many places (including the US) exclude the leading 0. This would affect taxi and ils commands.

image

n8rzz (approved):

cake-pie:
> _request_ a review

Thought I already did earlier, maybe something didn't work the way I expected, my bad.

> exclude the leading 0

Easy fix, will do, probably over the weekend.

erikquinn:
@cake-pie if you get a chance, would love to get this one in for the upcoming release, so ideally by Friday the 28th if able :)

erikquinn:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #1742, opened 2021-01-30
enhancement/1741 - Allow one decimal in holding pattern leg length

### LINKED ISSUE #1741: Leg length for hold definitions in airport files only allows integer minutes
# Background
**Example:** When defining a holding pattern at a waypoint with an initial heading of 090, left turns, 1 minute legs and a speed restriction of 230 knots the following string is used: `"090|left|1min|S230-"`.

When trying to implement the Stockholm/Arlanda airport (#1695), 1.5 minute legs are used for some holds. When using this it gives a runtime error `Uncaught Error: Unexpected parameter for Fix 'ELTOK': 1.5min`.

After looking at the code I found that the following regular expression in `globalConstants.js` is responsible for validating if the argument is a leg length: `HOLD_DISTANCE: /^([1-9]|1[0-9])(nm|min)$/i,`

It obviously only allows integer leg lengths between 1 and 19 nautical miles or minutes, not fractions.

## Suggestion
Edit the regular expression to allow leg lengths with one decimal between 0.1 and 19.9 min or nm. It should still be valid to use integers between 1 and 19 as before.

## Subtasks
- [x] Edit the regular expression to allow decimals as stated above
- [x] Do  code inspection to make sure decimal leg lengths shouldn't break anything
- [x] Do some manual testing to see that it works as intended
- [x] If issue #1737 is still open, mention this new information there
- [ ] <s>If issue #1737 is closed, update the documentation that the issue added to airport-format.md</s>

OlaRune:
Resolves #1741 

The purpose of this pull request is to allow one decimal in the leg length definition of holding patterns.

### DIFF
--- a/src/assets/scripts/client/constants/globalConstants.js
+++ b/src/assets/scripts/client/constants/globalConstants.js
@@ -31,7 +31,7 @@ export const REGEX = {
     ALTITUDE_RESTRICTION: /^A([0-9]{1,3})([+-]?)$/,
     COMPASS_DIRECTION: /^[NESW]/,
     DOUBLE_DOT: /\.\./g,
-    HOLD_DISTANCE: /^(0(?=\.)|[1-9]|1[0-9])(((?<!0)\.0|\.[1-9]))?(nm|min)$/i,
+    HOLD_DISTANCE: /^((?:0(?=\.)|[1-9]|1[0-9])(?:(?<!0)\.0|\.[1-9])?)(nm|min)$/i,
     LAT_LONG: /^([NESW])(\d+(\.\d+)?)([d °](\d+(\.\d+)?))?([m '](\d+(\.\d+)?))?$/,
     SPEED_RESTRICTION: /^S([1-9][0-9]{2})([+-]?)$/,
     SW: /[SW]/,
--- a/src/assets/scripts/client/constants/globalConstants.js
+++ b/src/assets/scripts/client/constants/globalConstants.js
@@ -31,7 +31,7 @@ export const REGEX = {
     ALTITUDE_RESTRICTION: /^A([0-9]{1,3})([+-]?)$/,
     COMPASS_DIRECTION: /^[NESW]/,
     DOUBLE_DOT: /\.\./g,
-    HOLD_DISTANCE: /^([1-9]|1[0-9])(nm|min)$/i,
+    HOLD_DISTANCE: /^(0(?=\.)|[1-9]|1[0-9])(((?<!0)\.0|\.[1-9]))?(nm|min)$/i,
     LAT_LONG: /^([NESW])(\d+(\.\d+)?)([d °](\d+(\.\d+)?))?([m '](\d+(\.\d+)?))?$/,
     SPEED_RESTRICTION: /^S([1-9][0-9]{2})([+-]?)$/,
     SW: /[SW]/,
--- a/src/assets/scripts/client/constants/globalConstants.js
+++ b/src/assets/scripts/client/constants/globalConstants.js
@@ -31,7 +31,7 @@ export const REGEX = {
     ALTITUDE_RESTRICTION: /^A([0-9]{1,3})([+-]?)$/,
     COMPASS_DIRECTION: /^[NESW]/,
     DOUBLE_DOT: /\.\./g,
-    HOLD_DISTANCE: /^((?:0(?=\.)|[1-9]|1[0-9])(?:(?<!^0)\.0|\.[1-9])?)(nm|min)$/i,
+    HOLD_DISTANCE: /^((?:0(?=\.)|[1-9]|[1-4][0-9])(?:(?<!^0)\.0|\.[1-9])?)(nm|min)$/i,
     LAT_LONG: /^([NESW])(\d+(\.\d+)?)([d °](\d+(\.\d+)?))?([m '](\d+(\.\d+)?))?$/,
     SPEED_RESTRICTION: /^S([1-9][0-9]{2})([+-]?)$/,
     SW: /[SW]/,
--- a/src/assets/scripts/client/constants/globalConstants.js
+++ b/src/assets/scripts/client/constants/globalConstants.js
@@ -31,7 +31,7 @@ export const REGEX = {
     ALTITUDE_RESTRICTION: /^A([0-9]{1,3})([+-]?)$/,
     COMPASS_DIRECTION: /^[NESW]/,
     DOUBLE_DOT: /\.\./g,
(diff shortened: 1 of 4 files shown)


OlaRune:
I have inspected the code manually and could find no problems with this change. Integers and floating point numbers are all treated the same in Javascript, so there shouldn't be a problem in any subsequent calculation either.

Furthermore I have manually tested editing leg length of holding patterns and measuring them using ctrl-click, and it seems to work.

src/assets/scripts/client/constants/globalConstants.js
```diff
@@ -31,7 +31,7 @@ export const REGEX = {
     ALTITUDE_RESTRICTION: /^A([0-9]{1,3})([+-]?)$/,
     COMPASS_DIRECTION: /^[NESW]/,
     DOUBLE_DOT: /\.\./g,
-    HOLD_DISTANCE: /^([1-9]|1[0-9])(nm|min)$/i,
+    HOLD_DISTANCE: /^(0(?=\.)|[1-9]|1[0-9])(((?<!0)\.0|\.[1-9]))?(nm|min)$/i,
```
OlaRune, line 34: This is some pseudo-regex (?) for understanding what this is doing.

1. Match beginning of string
2. Match a 0 if it is followed by a dot, otherwise match 1 to 9 or 10 to 19
3. Match .0 if it is not preceded by a 0, otherwise match .1 to .9
4. Match nm or min
5. Match end of string
6. Option to match nm and min without case-sensitivity

erikquinn:
@OlaRune Please show a significant use case for sub-mile precision on leg lengths.

src/assets/scripts/client/constants/globalConstants.js
cake-pie: This accomplishes your goal of allowing one decimal place for lengths in `min`, but inadvertently allows the same for `nm` as well. Do we really want that? Are there use cases to support that?

cake-pie:
Besides permitting one decimal place for lengths in airport file hold definitions, this also has the effect of permitting it for user entered commands in the sim. Same parser is used to interpret both, so may not be trivial to allow the former without the latter.
- is it desirable/acceptable for user commands to have decimal?
- if so, user facing `aircraft-commands.md` will need updated accordingly

src/assets/scripts/client/constants/globalConstants.js
cake-pie: I know the regex is currently only used to test and not match, but for future-proofing suggest that it should have two capturing groups. The first should capture the entire numeric portion, and the second containing the units. Use non-capturing groups for everything else.

n8rzz: I'll defer to @erikquinn on if we actually want that or not. He can say whether or not thats actually a thing.

I'm going to guess fractional `nm` probably isn't a thing and, thus, I love the idea of multiple capture groups.  I'm also a fan of make future us happy because we did things in a flexible way today 😉 

I like your plans here!

n8rzz (approved):
As written, I'm happy with your solution.  If you want to add multiple capture groups for the regex, feel free to tack that on here as well.

OlaRune:
Before this is merged I would like to change the following:
- [ ] Change the REGEX to only have two capturing groups, one for the number and one for the unit
- [ ]  Add information about decimal holding distances to aircraft-commands.md, also mention that the values have to be in ranges 1-19 or 0.1-19.9

I would also have someone verify that these assumptions are OK:
- [ ] It is acceptable that this also allows one decimal leg lengths when defining in nautical miles, even though the initial problem only concerned minutes
- [ ] It is acceptable that this change also allows one decimal leg in aircraft commands issued by the user
- [ ] Only allowing leg lenghts from 0.1 to 19.9 nm or min was a conscious design desicion and we want to keep it that way. (It would be trivial, even give a simpler REGEX, to allow values between 0.1 and 99.0)

Opinions? Feel free to just tick the boxes in the lower paragraph if you agree. @erikquinn @n8rzz @cake-pie

cake-pie:
> * acceptable to allow one decimal leg lengths in nautical miles

I raised this concern because the assumption might not hold, [need use cases for sub-mile precision](https://github.com/openscope/openscope/pull/1742#issuecomment-770308774) to support inclusion.
It's possible to add a lookahead to the regex and only allow decimal part if followed by `min` units but not `nm`:
`^((?:0(?=\.)|[1-9]|1[0-9])(?:(?:(?<!^0)\.0|\.[1-9])(?=min))?)(nm|min)$`

I'll also defer to @erikquinn re: whether to overlook the lack of use case in favor of slightly simpler regex

> * acceptable to allow one decimal leg in aircraft commands issued by the user

Am not aware of ATC issuing fractional minute/mile leg lengths. But we can rationalize with the following logic:
If there are charts that specify "X", then it is valid and conceivable for ATC to verbally issue "X".

> * only allowing lengths 0.1 to 19.9 nm or min was a conscious design decision and we want to keep it that way.

Added in #1444 (https://github.com/openscope/openscope/commit/45cbd53f6545bb19a3116dd4af80ed0d9befe257). Was not discussed, but seems very reasonable, no use case for larger values.
Not broken, why fix?

OlaRune:
Thanks @cake-pie for your input. Regarding the last thing about allowed ranges, it was just a thought I had, no need to change it.

Regarding the first to points I guess we'll wait for input from @erikquinn. I will leave it as it is for now.

n8rzz (approved):

erikquinn:
> Only allowing leg lenghts from 0.1 to 19.9 nm or min was a conscious design desicion and we want to keep it that way. (It would be trivial, even give a simpler REGEX, to allow values between 0.1 and 99.9)

@OlaRune 20nm legs are quite common, in the en route environment at least... Not sure if it was limited before, but in my wildest dreams, the most I could ever imagine would be 40 miles... so maybe let's go with `0.1-49.9`, or however else is easiest to cap it at some value that allows up to (at least) 40.

erikquinn (changes requested):
Please expand acceptable range up to 40nm.

OlaRune:
> Please expand acceptable range up to 40nm.

I don't know how I can mark this requested change as resolved (not that used to GitHub), but I'll re-request your review since that's what I can find when clicking around on your request.

I have changed the range from 1 to 49 or 0.1 to 49.9. This change made the tests fail, which was good, since I should have updated the tests to verify the behaviour for decimal numbers anyway. Tests are fixed now wrt. allowed range and decimal numbers.

erikquinn:
@OlaRune yep, just request a new review :)

erikquinn (approved):

PR #1696, opened 2021-01-09
feature/1695 - Copy EHAM airport file to have something to start from

OlaRune:
Resolves #1695 

The purpose of this pull request is to add ESSA Stockholm/Arlandra airport.

PR #1748, opened 2021-02-02
bugfix/1690 - Clear log upon airport changed

### LINKED ISSUE #1690: Log line from previous airport stuck
Browser: Chrome 87.0.4280.88

Airport: LOWW Vienna

Approximate play time before issue: Immediate

#### Issue Description
After switching airport from ENGM to LOWW the line "Oslo Approach, IBK222 with you descending through 13000 for 5000" was stuck in the text log above the command input prompt. When there was a message from an aircraft entering the LOWW airspace, both Oslo and Wien was displayed simultaneously, and when there was no message related to LOWW, only the Oslo line was displayed. (see both attached screenshots)

oslo_wien
only_oslo

#### Steps to reproduce
This error is easy to reproduce:
1. Start a game at an arbitrary airport.
2. Open the airport selection menu.
3. Wait for an aircraft to approach.
4. Just as the aircraft reports that it is with you, switch to another airport.
5. The line from the first airport is now stuck when playing at the new airport.

cake-pie:
Resolves #1690

The purpose of this pull request is to clear the log upon the airport being changed (or reloaded the same one).

This prevents log lines for the previous airport from getting stuck and never removed.

This also immediately clears log lines that would not have gotten stuck, but are still in the midst of their fadeout
 -- also desirable, there is no point leaving them on screen for up to another 10s, they are no longer relevant.

### Other solutions considered

Given [the cause of the issue](https://github.com/openscope/openscope/issues/1690#issuecomment-770246447) one may be tempted to somehow prevent the timers governing the fadeout of stuck log lines from being purged by `GameController.destroyTimers()` call in `AppController.onAirportChange()`. 

However, considering `game.timeouts` is a utility that could be used for a variety of timers with different callbacks, it's probably not a good idea (brittle, error-prone) to try and filter through it and selectively discard some entries while preserving others.

Anyway, it is better to immediately clear the log of all lines as they are irrelevant after airport switch/reload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - <a href="https://github.com/openscope/openscope/issues/1707" target="_blank">#1707</a> - Fix minor keyboard input inconsistencies and correct command documentation
 - <a href="https://github.com/openscope/openscope/issues/1716" target="_blank">#1716</a> - Prevent timewarping by negative values
 - <a href="https://github.com/openscope/openscope/issues/1744" target="_blank">#1744</a> - Fix validation helper method for preventing non-empty objects
+- <a href="https://github.com/openscope/openscope/issues/1690" target="_blank">#1690</a> - Clear log on airport change
 
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1703" target="_blank">#1703</a> - Add more PTL ranges, some including 30sec PTLs
--- a/src/assets/scripts/client/ui/UiController.js
+++ b/src/assets/scripts/client/ui/UiController.js
@@ -358,6 +358,8 @@ class UiController {
      * @chainable
      */
     setupHandlers() {
+        this.onAirportChangeHandler = this.onAirportChange.bind(this);
+
         return this;
     }
 
@@ -370,6 +372,8 @@ class UiController {
      * @method enable
      */
     enable() {
+        this._eventBus.on(EVENT.AIRPORT_CHANGE, this.onAirportChangeHandler);
+
         // TODO: move these to properly bound handler methods
         this.$fastForwards.on('click', (event) => GameController.game_timewarp_toggle(event));
         this.$githubLinkElement.on('click', (event) => this.onClickGithubLink(event));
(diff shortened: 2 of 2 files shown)


n8rzz (approved):

n8rzz:
I've updated the coverage threshold to not complain about loosing a small amount of coverage.  could you re-push to trigger those checks again?

cake-pie:
> re-push to trigger those checks again

I did actually `git commit --amend` and `git push --force` in this case to trigger the checks, but just discovered and verifed that it is possible to re-run checks on the same commit. Steps as follows.

1. expand the existing checks summary on the PR page, and click through to the detailed results
.
step1

2. hit that re-run jobs button and select the option to re-run all jobs
(perhaps in future we may be able to run only selected jobs?)
Note that the button is absent if you simply navigate to the ["checks" tab](https://github.com/openscope/openscope/pull/1748/checks) of the PR.  You'll have to expand the workflow run on the left and pick a job so that its log shows on the right,. Only then will the button appear.
.
step2

3. you can also [do this via the repo actions tab](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow) but will have to hunt down the specific workflow run you want from the full list

For future reference @n8rzz

OlaRune:
> > re-push to trigger those checks again
> 
> I did actually `git commit --amend` and `git push --force` in this case to trigger the checks, but just discovered and verifed that it is possible to re-run checks on the same commit. Steps as follows.
> 
> 1. expand the existing checks summary on the PR page, and click through to the detailed results
>    .
>    step1
> 2. hit that re-run jobs button and select the option to re-run all jobs
>    (perhaps in future we may be able to run only selected jobs?)
>    Note that the button is absent if you simply navigate to the ["checks" tab](https://github.com/openscope/openscope/pull/1748/checks) of the PR.  You'll have to expand the workflow run on the left and pick a job so that its log shows on the right,. Only then will the button appear.
>    .
>    step2
> 3. you can also [do this via the repo actions tab](https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow) but will have to hunt down the specific workflow run you want from the full list
> 
> For future reference @n8rzz

I thought this would help me restart the stuck checks on PR #1696 but I can't open the check details to find the re-run button if they've never finished... 😢

PR #1002, opened 2018-06-11
bugfix/955 - Overhaul `sid` command

eoinbutler:
Resolves openscope/openscope#955

The purpose of this pull request is to restore the functionality of the `sid` and `star` commands

PR #1761, opened 2021-02-09
devops/1749 - More robust unit tests for input validation, improved validation code

### LINKED ISSUE #1749: More robust tests for data validation (throws TypeError)
This is a follow-up to #1744. Since #1700 uncovered tests that spuriously passed:

- make tests for `throws` more robust: ensure that the throw originates from the expected place
  - can be accomplished by requiring a match to the specific error message
  - tests in `classname.spec.js` should be confined to throws in `classname.js`
    (from a preliminary look, exceptions currently exist)

- with the help of new tests, double check all validation (`throws new TypeError`)

cake-pie:
Resolves #1749 

The purpose of this pull request is to make unit tests for input validation more robust and improve validation code accordingly

### Scope
This covers all methods that have validation involving one or more of `_isEmpty`, `_isArray`, `_isObject` either directly or via the helper methods in `validatorUtilities.js`. This generally means constructors and `init` methods, usually dealing with data ingested from json files.

Was originally intended to cover anything that `throw new TypeError` but that turned out to be too much to swallow at once. 

### Validation code changes
- where constructor or `init` takes multiple arguments, check for missing args first before validating individual args
  - this facilitates writing tests that expect certain error messages; otherwise when there are missing args, which error message you get is dependent on the order in which args get validated.
- use `isEmptyOrNotArray` or `isEmptyOrNotObject` helpers as appropriate
- use `instanceof` checks as appropriate
- use a consistent error message wording throughout
- drop pointless `else`

### Tests changes
- tests `t.throws()` now check against expected error message; this ensure throws are indeed from the validation being tested and not due to other reasons
- separate test for missing args when multiple arguments are involved
- add tests for new validation code
- add missing tests

### Misc. changes
- reordered tests in `Fms.spec.js` to be in same order that the respective tested methods appear in `Fms.js`
  - originally intended to make more changes; soon after this I realized I needed to rein in the scope of the PR so those changes were dropped, but no harm keeping the reordering.
- in `Fms.js` change [this `throws`](https://github.com/openscope/openscope/blob/bc3caf812627b6c073eed7b96d6c4c088abb7b3d/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js#L936) in `replaceDepartureProcedure` to a `return [false, <readback>]`
  - makes consistent with the rest of the `replaceDepartureProcedure` method
  - originates from user command, should be displayed to user as readback in log rather than thrown to console
  - in #1002, was originally a `return` before refactoring in commit 491573e (inadvertently?)changed it to a `throw`
- discard defunct `isEmptyObject` from validation helpers
- typo fixes


src/assets/scripts/client/airport/MapCollection.js
```diff
@@ -23,22 +26,22 @@ export default class MapCollection extends BaseCollection {
...
-        } else if (mapJson.length === 0) {
-            throw new TypeError(
-                'Invalid mapJson parameter passed to MapCollection. Expected an array with at least one element'
-            );
+        if (_isNil(mapJson) || _isNil(defaultMaps) || _isNil(airportPositionModel) || !_isNumber(magneticNorth)) {
+            throw new TypeError('Invalid parameter(s) passed to MapCollection constructor. ' +
+                'Expected mapJson, defaultMaps, airportPositionModel and magneticNorth to be defined, but received ' +
+                `${typeof mapJson}, ${typeof defaultMaps}, ${typeof airportPositionModel} and ${typeof magneticNorth}`);
+        }
+        if (isEmptyOrNotArray(mapJson)) {
```
n8rzz, line 34: missing empty line above `if` statements

src/assets/scripts/client/airport/MapModel.js
```diff
@@ -25,18 +30,26 @@ export default class MapModel extends BaseModel {
...
-        } else if (!_isString(map.name)) {
-            throw new TypeError(
-                `Invalid parameter, map.name must be a string, but found ${typeof map.name}`
-            );
+        if (_isNil(map) || _isNil(airportPosition) || !_isNumber(magneticNorth)) {
+            throw new TypeError('Invalid parameter(s) passed to MapModel constructor. ' +
+                'Expected map, airportPosition and magneticNorth to be defined, ' +
+                `but received ${typeof map}, ${typeof airportPosition} and ${typeof magneticNorth}`);
+        }
+        if (isEmptyOrNotObject(map)) {
```
n8rzz, line 38: missing empty line above `if` statements

n8rzz (approved):
My only complaint on this is the line spacing of the new `if` statements.  Please be sure to add an empty line above each `if` statement

src/assets/scripts/client/airport/MapModel.js
cake-pie: done

src/assets/scripts/client/airport/MapCollection.js
cake-pie: done

Issue #47, opened 2017-01-16
Separate "Airport" / "Facility" / "Sector"


erikquinn:
Currently, everything that has to do with playing in a certain airspace is stored as properties of the "airport". But what about places where a single ATC facility is providing services to *multiple* airports? (aka just about everywhere)

Work should be done to move away from keeping things like navigational data (fixes/airways/procedures), video maps, etc under the airport, and make new classes called `Facility` and `Sector` which represents the given ATC facility and control position being simulated. For example, if I am working Miami Approach (consolidated to a single sector), I would work arrivals/departures to/from KMIA, KFLL, KOPF, KTMB, and others. My `Sector` within the Miami TRACON `Facility` has the stuff relating to my role in the simulation, and each of the airports are an `Airport`, which have their own runways and procedures defined within.

Here's what I'm thinking:
```
Facility
    └> SectorCollection
        └> Sector
```
In the following examples, we will use KSFO.

The `Facility` would be like the ATC *facility* we are emulating. This is NorCal Consolidated TRACON ("NCT"). It will have a collection of `Sector`s within `SectorCollection` that can be used, and will contain all information that is general to the area, but not specific to any particular sector.

The `Sector` will be each real-world (or simplified) sector are emulating. For example, SFO arrivals are handled by Area B of NCT, including the "Woodside", "Foster", "Boulder", "Laguna", "Niles", and "Cedar" sectors. Each has their own airspace boundaries, frequencies, etc.

In a given simulation, users can choose to work only a single sector within a given facility (perhaps Woodside, the 28L final sector), all the sectors of the facility combined simultaneously, or a combination of certain adjacent sectors (perhaps Woodside and Foster, the final sectors for the 28s).

---

And things will be divided up similar to this:

### Facility:
 - Fixes
 - Airways
 - `SectorCollection`

### Sector:
 - Airspace
 - Frequency, Callsign, Position Symbol
 - Scope Layout
 - Airports to load

### Airport:
 - Instrument Procedures (SID/STAR/TRXN)
 - Runways
 - Taxiways
 - Gates & Parking Spots

---

Originally reported in zlsa/atc#745.

OlaRune:
I'm just gravedigging a little to see what has been discussed before...

A problem that needs to be discussed is spawnPatterns. Every playable sector (and combination of sectors) would need to have their own spawnPatterns to simulate controllers of adjacent sectors transferring aircraft to the player in a beleivable way.

[...]

PR #1773, opened 2021-02-14
bugfix/1762 - defensive coding in Fms initialization

### LINKED ISSUE #1762: Runtime error when defining spawnPattern with both origin and destination (and origin is implemented)
Browser: Mozilla Firefox 85.0.1
Airport: ESSA (while implementing, though it doesn't really matter)
Approximate play time before issue: At startup

#### Issue Description
_Note: I know that this use of spawnPatterns isn't recommended (and apparently not supported...) because of other issues, such as no handling of separation at spawn when aircraft from different spawnPatterns spawn along the same airway. Though my guess is that the plan is for it to be supported in the future, which is why I'm reporting this as an issue._

When adding spawnPatterns for arrivals (though I would guess the corresponding problem happens for departures) where both `origin` and `destination` are entered, and the `origin` is an airport that is implemented in the sim, a runtime error is thrown. Everything works fine when the `origin` airport is not implemented.

Works fine (Heathrow is not in sim):
`{
            "origin": "EGLL",
            "destination": "ESSA",
            "category": "arrival",
            "route": "INVOL.N866.BEDLA..ELTOK.ELTOK7M.ESSA01R",
            "altitude": "18000",
            "speed": "330",
            "method": "random",
            "rate": 0.7,
            "airlines": [
                ["sas/short", 1],
                ["baw/short", 1]
            ]
        }`

Not working (Manchester is in sim):
`{
            "origin": "EGCC",
            "destination": "ESSA",
            "category": "arrival",
            "route": "INVOL.N866.BEDLA..ELTOK.ELTOK7M.ESSA01R",
            "altitude": "18000",
            "speed": "330",
            "method": "random",
            "rate": 0.7,
            "airlines": [
                ["sas/short", 1],
                ["baw/short", 1]
            ]
        }`

The error is `Uncaught TypeError: Expected instance of RunwayModel, but received null` and below that `setDepartureRunway Fms.js:1020`. I assume that it is expecting me to include the complete route from departure runway -> SID -> route -> STAR -> arrival runway, but this shouldn't be necessary.

#### Steps to reproduce
- Add a spawnPattern where the origin is included in the sim (see example above)
- Build and start
- Problem occurs right after loading

cake-pie:
Resolves #1762

The purpose of this pull request is to add validation/protection to initialization methods in `Fms`:
- ~initialization of arrival/departure airport+runway gated behind an explicit check of the `category` i.e. arriving or departing flight rather than based on the presence/absence of `destination`/`origin`~
- validate + enforce expectation that `destination`/`origin` airport is the current airport
  - in future (to support #47) needs expanded to airports within tracon

### DIFF
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -271,9 +271,7 @@ export default class Fms {
 
         this._verifyRouteContainsMultipleWaypoints();
         this._initializeFlightPhaseForCategory(category);
-
-        this._initializeAirportsAndRunways(category, origin, destination);
-
+        this._initializeAirportsAndRunways(origin, destination);
         this._initializeFlightPlanAltitude(altitude, category, model);
         this._initializePositionInRouteToBeginAtFixName(nextFix, category);
 
@@ -305,18 +303,20 @@ export default class Fms {
      *
      * @for Fms
      * @method _initializeAirportsAndRunways
-     * @param category {string} Flight category (arrival, departure, or overflight)
      * @param origin {string} ICAO identifier specified by spawn pattern
      * @param destination {string} ICAO identifier specified by spawn pattern
      * @private
      */
-    _initializeAirportsAndRunways(category, origin, destination) {
-        if (category === FLIGHT_CATEGORY.DEPARTURE) {
+    _initializeAirportsAndRunways(origin, destination) {
+        const originLowerCase = origin.toLowerCase();
+        const destinationLowerCase = destination.toLowerCase();
+
+        if (originLowerCase === AirportController.current.icao) {
             this._initializeDepartureAirport(origin);
             this._initializeDepartureRunway();
         }
 
-        if (category === FLIGHT_CATEGORY.ARRIVAL) {
+        if (destinationLowerCase === AirportController.current.icao) {
             this._initializeArrivalAirport(destination);
             this._initializeArrivalRunway();
         }
@@ -331,12 +331,6 @@ export default class Fms {
      * @private
      */
     _initializeArrivalAirport(destinationIcao) {
-        const destinationIcaoLowerCase = destinationIcao.toLowerCase();
-
-        if (destinationIcaoLowerCase !== AirportController.current.icao) {
-            return;
-        }
-
         this.arrivalAirportModel = AirportController.airport_get(destinationIcao);
     }
 
@@ -374,12 +368,6 @@ export default class Fms {
      * @private
      */
     _initializeDepartureAirport(originIcao) {
-        const originIcaoLowerCase = originIcao.toLowerCase();
-
-        if (originIcaoLowerCase !== AirportController.current.icao) {
-            return;
-        }
-
         this.departureAirportModel = AirportController.airport_get(originIcao);
     }
 
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -320,6 +320,12 @@ export default class Fms {
             return;
         }
 
+        destinationIcao = destinationIcao.toLowerCase();
+
+        if (destinationIcao !== AirportController.current.icao) {
+            return;
+        }
+
         this.arrivalAirportModel = AirportController.airport_get(destinationIcao);
     }
 
@@ -358,6 +364,12 @@ export default class Fms {
      */
     _initializeDepartureAirport(originIcao) {
         if (originIcao === '') {
+            return;
+        }
+
+        originIcao = originIcao.toLowerCase();
+
+        if (originIcao !== AirportController.current.icao) {
             return;
         }
 
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -316,10 +316,6 @@ export default class Fms {
      * @private
      */
     _initializeArrivalAirport(destinationIcao) {
-        if (destinationIcao === '') {
-            return;
-        }
-
         destinationIcao = destinationIcao.toLowerCase();
 
         if (destinationIcao !== AirportController.current.icao) {
@@ -363,10 +359,6 @@ export default class Fms {
      * @private
      */
     _initializeDepartureAirport(originIcao) {
-        if (originIcao === '') {
-            return;
-        }
-
         originIcao = originIcao.toLowerCase();
 
         if (originIcao !== AirportController.current.icao) {
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
@@ -272,15 +272,7 @@ export default class Fms {
         this._verifyRouteContainsMultipleWaypoints();
         this._initializeFlightPhaseForCategory(category);
 
-        if (category === FLIGHT_CATEGORY.DEPARTURE) {
-            this._initializeDepartureAirport(origin);
-            this._initializeDepartureRunway();
-        }
-
-        if (category === FLIGHT_CATEGORY.ARRIVAL) {
-            this._initializeArrivalAirport(destination);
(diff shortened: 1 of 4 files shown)


src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -271,10 +271,17 @@ export default class Fms {
...
+
+        if (category === FLIGHT_CATEGORY.DEPARTURE) {
+            this._initializeDepartureAirport(origin);
+            this._initializeDepartureRunway();
+        }
+
+        if (category === FLIGHT_CATEGORY.ARRIVAL) {
+            this._initializeArrivalAirport(destination);
+            this._initializeArrivalRunway();
+        }
```
erikquinn, line 283: What if they depart and arrive at fields which are both under our control? This will be possible.

erikquinn (changes requested):

src/assets/scripts/client/trafficGenerator/SpawnScheduler.js
```diff
@@ -39,7 +39,7 @@ class SpawnScheduler {
             throw new TypeError('Invalid parameter. SpawnScheduler requires aircraftController to be defined.');
         }
 
-        this.aircraftController = aircraftController;
+        this._aircraftController = aircraftController;
```
n8rzz, line 42: good call!

n8rzz (changes requested):

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: I should add part of the original intent of the `origin` / `destination` fields was that they were intended to replace the `category` (since they are rather duplicitave). I can't remember the reason why we didn't just do it at that time though, so maybe there's something it would have broken.

But the `category` alone will not support all the required combinations, hence moving to just defining if and where they should (a) spawn on the ground at, and/or (b) which airport they are landing at.

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -271,10 +271,17 @@ export default class Fms {
 
         this._verifyRouteContainsMultipleWaypoints();
         this._initializeFlightPhaseForCategory(category);
-        this._initializeDepartureAirport(origin);
-        this._initializeDepartureRunway();
-        this._initializeArrivalAirport(destination);
-        this._initializeArrivalRunway();
+
+        if (category === FLIGHT_CATEGORY.DEPARTURE) {
```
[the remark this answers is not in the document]
cake-pie: Can do, though will still need the compare for `FLIGHT_CATEGORY.ARRIVAL` because overflight is also a category (though afaik currently unused)

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
cake-pie: If as per other discussion in #1762 there is a desire to have both the origin and destination sections in the flight strip be populated (for immersive reasons) then you once again *cannot* rely purely on presence/absence of `origin` and `destination` properties in the `aircraftInitProps` passed to `Fms` constructor, because both will be populated with values to be shown on the flight strip -- [here](https://github.com/openscope/openscope/blob/c391f605530dac4c5e9f8554d4ecf56c5b17f28d/src/assets/scripts/client/aircraft/AircraftModel.js#L506) is the call from `AircraftModel` constructor which passes along the same set of initialization properties that it received from `AircraftController._createAircraftWithInitializationProps()`. `AircraftModel` will itself also [parse](https://github.com/openscope/openscope/blob/c391f605530dac4c5e9f8554d4ecf56c5b17f28d/src/assets/scripts/client/aircraft/AircraftModel.js#L521) the same set of init props to obtain the [values](https://github.com/openscope/openscope/blob/c391f605530dac4c5e9f8554d4ecf56c5b17f28d/src/assets/scripts/client/aircraft/AircraftModel.js#L685-L686) that ultimately show up in the flight strip.

Even overflights could/would have `origin` and `destination` values.

> `category` alone will not support all the required combinations

Then perhaps we will need to consider expanding `category` to support four possibilities: "arriving here", "departing here", "neither (overflight)" and "both". (where "here" currently means one airport but in future could be a TRACON with multiple fields)

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -309,7 +316,9 @@ export default class Fms {
      * @private
      */
     _initializeArrivalAirport(destinationIcao) {
-        if (destinationIcao === '') {
+        destinationIcao = destinationIcao.toLowerCase();
```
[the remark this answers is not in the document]
cake-pie: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -350,7 +359,9 @@ export default class Fms {
      * @private
      */
     _initializeDepartureAirport(originIcao) {
-        if (originIcao === '') {
+        originIcao = originIcao.toLowerCase();
```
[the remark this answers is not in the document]
cake-pie: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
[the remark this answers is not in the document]
cake-pie: done

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: I can't imagine a scenario where the `category` is a necessary parameter; everything can be more clearly inferred from the origin and destination. The mere existence of `category` could only serve to confusion when inevitably someone specifies a "departure" category with no origin, etc. Since we have to specify _where_ they're departing/arriving already, there's no reason that they _are_ a departure/arrival.

The conversation in #1762 is agreed to this effect, with the change being not to use a `origin`/`destination` property on the spawnPattern, but instead to specify origin/destination in the airline list for a given spawnPattern.

> Even overflights could/would have origin and destination values.

This is the desired behavior. All aircraft in the simulation would have origin/destination values (unless they weren't specified, for example with an overflight). We _do not need_ to know an aircraft's "category" for anything the sim does. There are places that we have logic branches based on its value, but these can and should be handled in a different way anyway.

For example, in the current implementation, there is only one airport that we control at a time. An aircraft is an arrival if `aircraftModel.fms.arrivalAirportModel.icao === AirportController.current.icao`. If an aircraft is being spawned and `aircraftModel.fms.departureAirportModel.icao === AirportController.current.icao`, then we know to spawn it on the ground, etc.

The issue is that what we truly cannot do is add support for multiple airports as long as decisions are made based on a "category". Every aircraft is a departure, overflight, _and_ arrival-- we shouldn't be labeling the aircraft as one thing for its whole life, as it is only a matter of the perspective of the where the controller is along their route. That's why we want to move to specifying where they start and end, and make all our decisions based on that alone.

cake-pie: Removed `category` check and instead rely purely on whether `origin`/`destination` refer to current airport to decide whether or not to initialize the `departureAirportModel`+`departureRunwayModel` / `arrivalAirportModel`+`arrivalRunwayModel` respectively.

I'm not thoroughly convinced this is fully safe -- what if both origin and destination are set as current airport, would that surface other problems elsewhere, etc.

But it does at least still guarantee that [the reported null value](https://github.com/openscope/openscope/issues/1762#issuecomment-778681639) is eliminated so I'll leave it at that.

n8rzz (approved):

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
erikquinn: > I'm not thoroughly convinced this is fully safe -- what if both origin and destination are set as current airport, would that surface other problems elsewhere, etc.

Right, a valid concern, and it might. My point is that we have to face those issues anyway in the effort to move to supporting various future functionality, so avoiding them is just further delaying the transition. 👍

erikquinn (approved):

PR #1784, opened 2021-02-20
refactor/1783 - no new AircraftTypeDefinitionModel per AircraftModel

### LINKED ISSUE #1783: Instances of AircraftModel do not need their own copy of AircraftTypeDefinitionModel
In `AircraftModel` constructor:

https://github.com/openscope/openscope/blob/7a70be11d653aa594f7293bd5effbda1646c530c/src/assets/scripts/client/aircraft/AircraftModel.js#L483

`options.model` is already an instance of `AircraftTypeDefinitionModel`; this has the effect of creating a copy.

Consider that `options.model` is drawn from the `AircraftTypeDefinitionCollection` that is populated at app start from consolidated `aircraft.json`, containing data about performance of each aircraft type. Such data is never modified.

Therefore, individual `AircraftModel` should not need to create their own copy of `AircraftTypeDefinitionModel`.
The above should simply be:

```js
this.model = options.model;
```

This change breaks some tests but that can be rectified by using fixtures.

cake-pie:
Resolves #1783

The purpose of this pull request is to not make a new copy of `AircraftTypeDefinitionModel` for every individual aircraft.

Why aren't the fixtures in `aircraftFixtures.js`?, you may ask.
I did try that at first; it doesn't work. Might be a circular dependency issue.


n8rzz (approved):
good find!

PR #1788, opened 2021-02-22
devops/1787 - JSON schema consistency

### LINKED ISSUE #1787: JSON schema document standards
Part one of #1777.

The two JSON schema we currently have were admittedly a bit of a slapdash job.

- lay out some initial guidelines for JSON schema to ensure better consistency moving forward
  - may evolve depending on needs, this is just to get started somewhere
- begin applying them to the schema that we have

---

JSON Schema documents `*.schema.json` must:

- identify the version of JSON Schema meta-schema used, e.g.:
  `"$schema": "https://json-schema.org/draft/2019-09/schema"`
- validate against the meta-schema specified (= be a valid schema)
  until we have our own tools set up, check using [this online validator](https://www.jsonschemavalidator.net/) for now
- have `$id` specifying its absolute URI in the root schema
  should match glob `http://openscope.io/assets/**/*.schema.json`
- have a `title`; `description` is optional
- order: `$schema`, `$id`, `title`, optional `description`, `type`, then etc.; and finally `$defs` (if any)

---

Sub-schemas (i.e. each field/property) must

- provide an explanatory `description`; `title` not required
  - exception: inlined re-usable sub-schema via `"$ref": "..."`
  - exception: arrays using list validation -- redundant, it's a list of whatever is described under its `items`
    - clarification: arrays using tuple validation not exempt
- order: optional `title`, `description`, `type`, then etc.

---

Re-usable sub-schemas must

- be defined in `$defs` property of root schema
  - do not use `definitions` (obsolete but still appears in some older tutorials/docs)
  - for ease of lookup, must be immediate child of `$defs`, not nested elsewhere
- be defined under a camelCase key, e.g. `tutorialStep`
  - key is a straightforward noun describing what it is
  - do not append "definition"/"def" e.g. ~`tutorialStepDefinition`~ / ~`tutorialStepDef`~
- have `$anchor` value, identical to the key
  - do not use `$id`!
- order: `$anchor`, optional `title`, `description`, `type`, then etc.

---

Comments
json does not allow javascript-style comments, these are workarounds to use

- `$comment`
  - this key can be used to add comments in schema documents

- `_comment`
  - standardized key to be used for comments in json files;
  - include this as an allowable optional property of objects where someone might reasonably want to add a dev comment

cake-pie:
Resolves #1787

The purpose of this pull request is to establish and follow basic guidelines for JSON schema documents.

### DIFF
--- a/assets/autocomplete/commandAutocompleteConfig.schema.json
+++ b/assets/autocomplete/commandAutocompleteConfig.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema",
-  "$id": "http://www.openscope.co/assets/autocomplete/commandAutocompleteConfig.schema.json",
+  "$id": "http://www.openscope.io/assets/autocomplete/commandAutocompleteConfig.schema.json",
   "title": "Openscope ATC simulator command autocomplete configuration file schema",
   "type": "object",
   "properties": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1977" target="_blank">#1977</a> - Update scoring documentation
 - <a href="https://github.com/openscope/openscope/issues/1991" target="_blank">#1991</a> - Update KPHL magnetic variation
+- <a href="https://github.com/openscope/openscope/issues/1787" target="_blank">#1787</a> - Standardize JSON schemas
 
 
 # 6.28.0 (July 3, 2022)
--- a/assets/airports/airportLoadList.schema.json
+++ b/assets/airports/airportLoadList.schema.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema",
-  "$id": "http://openscope.co/assets/airports/airportLoadList.schema.json",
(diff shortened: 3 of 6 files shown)


cake-pie:
Now that autocomplete was also merged into trunk, this has been rebased, rechecked, and updated to use openscope.io instead of openscope.co URIs. Ready to go.

assets/autocomplete/commandAutocompleteConfig.schema.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2019-09/schema",
-  "$id": "http://www.openscope.co/assets/autocomplete/commandAutocompleteConfig.schema.json",
+  "$id": "http://www.openscope.io/assets/autocomplete/commandAutocompleteConfig.schema.json",
```
n8rzz, line 3: @erikquinn Looks like our `.co` domain is still pointing to Heroku. When you have some time, could you update the dns to point to Render?

n8rzz (approved):
I don't think we need to update `.co` -> `.io` in the docs. Erik will get that back up and running soon.

cake-pie:
I just went with the domain used on the repository front page; up to you guys which you prefer to have as the primary one.

assets/autocomplete/commandAutocompleteConfig.schema.json
erikquinn: done

PR #1796, opened 2021-03-24
bugfix/1789 - Empty initial commit

### LINKED ISSUE #1789: Route amendment removing a useless airway not behaving correctly
At KSEA, change a departure from `HAROB6 FEPOT Q3 FAMUK` to go direct FAMUK after FEPOT, by entering `route fepot..famuk`. Since there actually are no fixes between FEPOT and FAMUK on Q3, it does not handle the change correctly, yielding `HAROB6 FEPOT FAMUK Q3 FAMUK` instead of `HAROB6 FEPOT FAMUK`.

This is happening somewhere within `RouteModel._createAmendedConvergentLeg()`.

OlaRune:
Resolves #1789 

The purpose of this pull request is to fix a wrongfully represented route on KSEA.


OlaRune:
Okay, so I fixed this rather quickly, just to do something tonight.

However, after I fixed it I checked the deployed version on openscope.io and I couldn't reproduce the original issue. Maybe the root cause has been fixed elsewhere? I have included one screenshot of how it looked before and one of how it looks after the fix. See below!

Maybe this issue/PR should be closed without merging? Need your input @erikquinn.

**Before any changes:**
2021-03-24-210000_1920x1080_scrot

**After fix described in issue:**
2021-03-24-210032_1920x1080_scrot

OlaRune:
> Hey @OlaRune! While this does resolve the issue I found, it doesn't address the underlying problem-- the issue was about how the sim _handled_ the route change instruction, not the fact that the airport file had the route used that way. Though goofy, it's technically still valid syntax to include an airway as a method used to travel between two fixes.
> 
> The solution to the issue will need to be a change to how the `FEPOT..FAMUK` route string is absorbed into the `KSEA16L.HAROB6.FEPOT.Q3.FAMUK` route string.
> 
> Closing for now; if you'd like to take a swing at the problem described, feel free to reopen this for it :)

I see! I didn't see the part where it said "change a departure", so that's why I thought the problem didn't exist anymore. Yeah, this can stay closed. Sorry! :)

PR #1803, opened 2021-04-07
enhancement/1732 - Add pronunciation to fixes in airport files

### LINKED ISSUE #1732: Add pronunciation to fixes in airport files
# Background
In real life there are fixes whose identifiers resemble real words, in these cases they are often pronounced as these words by pilots and ATC. There can also be fixes named after places on the ground, whose identifier is an abbreviation of that. In those cases they are often pronounced as the place name.

An example is [this Youtube video](https://youtu.be/EWAsQ3qldo8) where the aircraft in their mayday call reports that their intention is to continue direct to SPRKY, which they pronounce "Sparky".

# Suggestion
Add an **optional** `spoken` parameter that can be used when defining fixes, which tells the sim how the identifier is pronounced, similar to how `name` works for SIDs and STARs.

This can also be used for forcing some readouts to use the phonetic alphabet if the voice engine tries to be too clever for its own good, and things just get weird, as commented by @cake-pie below.

# Subtasks
- [x] Add an optional parameter `spoken` to fix definitions
- [x] Make the text-to-speech use the `spoken` parameter if it exists
- [x] Update documentation/airport-format.md
- [x] Update tests accordingly

OlaRune:
Resolves #1732

The purpose of this pull request is to add an optional `spoken` parameter to fixes that defines what is said by the speech engine.


src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -162,8 +180,11 @@ class FixCollection extends BaseCollection {
      * @private
      */
     _buildFixModelsFromList(fixList, referencePosition) {
-        _forEach(fixList, (fixCoordinates, fixName) => {
-            const fixModel = new FixModel(fixName, fixCoordinates, referencePosition);
+        _forEach(fixList, (fixData) => {
+            if (!fixData.spoken) {
+                fixData.spoken = null; // the initializer for FixModel handles this correctly
```
OlaRune, line 185: Maybe this is unnecessary. Is it already `null` or `undefined`? (comment for myself)

src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -100,6 +100,24 @@ class FixCollection extends BaseCollection {
...
+     * @for FixCollection
+     * @method getFixSpokenName
+     * @param fixName {string}
+     * @return {string}
+     */
+    getFixSpokenName(fixName) {
+       const fixModel = this.findFixByName(fixName);
+
+       if (!fixModel) {
+           return fixName.toLowerCase(); // lower case to force speech engine to pronounce instead of spell
```
OlaRune, line 115: Maybe this should be upper-case? How are strange fix names pronounced in lower-case? (comment for myself)

src/assets/scripts/client/navigationLibrary/FixModel.js
```diff
@@ -74,23 +84,41 @@ export default class FixModel extends BaseModel {
...
         // TODO: should this be a throwing instead of returning early?
         if (!fixName || !fixCoordinate || !referencePosition) {
             return;
         }
 
         this.name = fixName.toUpperCase();
+
+        if (!fixSpoken) {
+            this._spoken = fixName.toLowerCase();
+        } else {
```
OlaRune, line 118: Assign the default value before the `if` to avoid the `else`. (comment for myself)

src/assets/scripts/client/navigationLibrary/FixCollection.js
OlaRune: I've experimented a bit, and it will spell fixes that are completely impossible to pronounce, for example `XFDCG`. However, other names, such as `ASDFG`, will have it try to pronounce them, which sounds super weird. I think I'll have to make it spell out all fixes that don't have a defined `spoken`.

@Fechulo: See this comment as the answer to our discussion on Slack. Too bad it didn't work out nicely! :sob:

OlaRune: This is now changed.

src/assets/scripts/client/navigationLibrary/FixModel.js
OlaRune: This is fixed.

src/assets/scripts/client/navigationLibrary/FixCollection.js
OlaRune: Most likely `undefined` if it is not in the JSON. I have changed it and tested and it seems to work.

src/assets/scripts/client/navigationLibrary/FixModel.js
```diff
@@ -74,23 +84,40 @@ export default class FixModel extends BaseModel {
...
      */
-    init(fixName, fixCoordinate, referencePosition) {
+    init(fixName, fixSpoken, fixCoordinate, referencePosition) {
         // TODO: should this be a throwing instead of returning early?
         if (!fixName || !fixCoordinate || !referencePosition) {
             return;
         }
 
         this.name = fixName.toUpperCase();
+        this._spoken = this.name; // upper-case forces spelling, experiments shows that this works best for unknown pronunciation
```
OlaRune, line 115: I'm still not completely sure about this, I'll let you reviewers say what you think of this.

**Uppercase**
Forces names to always be spelled out if we don't have a defined `spoken` parameter. This will always lead to reasonable results, although we miss some opportunities where the voice engine could have pronounced the names in a better way.

**Lowercase**
This forces the voice engine to try pronouncing the names as words. Works nicely for names resembling words, such as `AUBRN`. Also works well for some hard-to-pronounce names, such as  `XDRCX`, which it spells out instead of pronouncing. Produces really weird results for other names that shouldn't be pronounced but it still tries, such as `ASDFG`, which makes it impossible to understand what it is saying.

What do you think? The lowercase version was used before, but I think it's better to force spelling now, since we can override it by defining a `spoken` instead.

OlaRune:
**For the reviewers:**
I have added `spoken` parameters to 3 fixes at KSEA, since those are used in `airport-format.md`. These can be used if you want to test the new functionality from the PR deployment.

For example, order a `direct` to any of these to hear it.
- EUG
- KRUZR
- PAE

src/assets/scripts/client/navigationLibrary/FixModel.js
erikquinn: Definitely still lowercase. Most waypoints have some sort of discernable pronunciation, and that is what would be used on frequency, not spelling out a 5-letter fix. And as you mention, when the fix doesn't have a reasonable pronunciation, it spells it out. But we definitely don't want to require a spoken parameter to prevent the sim from pronouncing `APPLE` as `alpha papa papa lima echo`-- the `spoken` parameter is the exception when we want it to be pronounced a specific way which is different than what the robot lady chose :)

assets/airports/cyhz.json
```diff
@@ -588,96 +588,96 @@
...
-        "YQM"  : ["N46d11.34", "W064d34.25"],
-        "YZX"  : ["N44d55.37", "W065d06.11"],
-        "ZHZ"  : ["N44d48.14", "W063d35.38"],
-        "ZNS"  : ["N44d58.00", "W063d25.62"]
-    },
+    "fixes": [
+        {"name": "AGMIR", "coordinates": ["N45d13.70", "W061d57.33"]},
+        {"name": "ALLEX", "coordinates": ["N44d25.00", "W067d00.00"]},
+        {"name": "AMUNA", "coordinates": ["N44d57.74", "W063d07.54"]},
+        {"name": "ARDEE", "coordinates": ["N45d07.40", "W064d23.17"]},
```
erikquinn, line 595: This pattern worries me. If using an array of objects rather than the fix name as the key of an object property, we introduce the possibility of duplicated fixes. As it stands currently, this is protected by a lint error (attempting to define two of the same key in an object). I would suggest instead something like:
```
"fixes": {
        "AGMIR": ["N45d13.70", "W061d57.33"],
        "ALLEX": ["N44d25.00", "W067d00.00"],
        "AMUNA": {"coordinates": ["N44d57.74", "W063d07.54"], "spoken": "amoonuh"},
        "ARDEE", {"coordinates": ["N45d07.40", "W064d23.17"], "spoken": "party},
}
```

and then from the FixModel constructor, we can simply check if `Array.isArray(fixData)`-- if so, treat as positional array with spoken undefined, and else `const { coordinates, spoken } = fixData;` to pull out the two values and use them. If the fix data provided is not an object with those two properties, an error will be thrown naturally, ensuring the existence of the parameters without any other manual defensiveness. 😄

src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -162,8 +180,8 @@ class FixCollection extends BaseCollection {
      * @private
      */
     _buildFixModelsFromList(fixList, referencePosition) {
-        _forEach(fixList, (fixCoordinates, fixName) => {
-            const fixModel = new FixModel(fixName, fixCoordinates, referencePosition);
+        _forEach(fixList, (fixData) => {
+            const fixModel = new FixModel(fixData.name, fixData.spoken, fixData.coordinates, referencePosition); // FixModel accepts undefined spoken
```
erikquinn, line 184: We should just pass `fixData` here, and have the FixModel constructor destruct the object in its _init() method. I can see you were just matching the pattern that was already there, but where practical we prefer to squish these together into an object and pass that object to the constructor. If anything is missing, we deal with it in ._init() and throw/return as appropriate.

src/assets/scripts/client/navigationLibrary/FixCollection.js
```diff
@@ -100,6 +100,24 @@ class FixCollection extends BaseCollection {
...
 
+    /**
+     * Returns the pronunciation of the fix name if it exists, otherwise returns the input name
+     *
+     * @for FixCollection
+     * @method getFixSpokenName
+     * @param fixName {string}
+     * @return {string}
+     */
+    getFixSpokenName(fixName) {
```
erikquinn, line 111: I find it hard to think of a case where we would have the fix name, but not the FixModel itself. Should we not just achieve this with `FixModel.getSpokenName()`?

Is this maybe just for the readbacks, like with the `pd` command? If so and the fix model isn't readily accessible (or inconvenient), this is fine. Just checking since I wasn't expecting to see it here.

assets/airports/cyhz.json
OlaRune: Wilco! That sounds great.

src/assets/scripts/client/navigationLibrary/FixCollection.js
OlaRune: Sounds good!

src/assets/scripts/client/navigationLibrary/FixCollection.js
OlaRune: There are two reasons why I chose to do it this way.

1. Easier use of the function for evaluating the spoken name directly in readback strings, without having to look up if the actual fix exists or not. If it doesn't exist the input is read back.

2. The readback string is constructed in a few different places, with access to different object types. I'm on mobile right now, but I recall that somewhere it had a FixModel, in most places a WaypointModel, and I think there was a HoldModel or something at one place. These other types of models don't have an easy way to access the underlying FixModel, they have just copied the fix name. So we can't access the FixModel without searching for it in the FixCollection anyway. So I decided this solution would need less code and lookups from different places.

I prefer to leave this as it is, if you don't have very strong feelings about doing it another way.

src/assets/scripts/client/navigationLibrary/FixModel.js
OlaRune: Alright, I'll change it back to lowercase so we'll let the airport developer make sure they are pronounced correctly if something is weird.

OlaRune:
@erikquinn I have updated the pull request according to your comments. Since it is now backwards compatible with the old fix definition syntax I have reverted the changes to the airport files. Also updated tests accordingly and added a few new tests.

Feel free to review again when you have time. Less files to look through now. :smile:

src/assets/scripts/client/navigationLibrary/FixCollection.js
erikquinn: Okay, that all makes sense. Fine as is.

erikquinn (approved):
Pending lint/tests passing, this is approved with a slight modification.

Not sure why I didn't suggest this instead in the first place, but there's really no need for an object-based definition at all. lat/lon are always required, and the spoken parameter is always optional... so we can just stick with the array shape, and where desired, add the spoken string as a third element in the array. This also reduces a lot of the required defensiveness involved with handling different ways the input data could be malformed by new airport devs 😉 

Now simply works with:
```javascript
"FIXXA": [41.124, -81.822],
"FIXXB": [41.922, -82.049, "fixbee"],
"FIXXC": [40.732, -81.527, "yabba dabba doo"],
"FIXXD": [40.284, -81.181],
...
```

PR #1804, opened 2021-04-07
Add P212 and update KAP

Hawaiicoder:
P212 added to KAP.


Fechulo (changes requested):
New lines missing at the end of both files

Hawaiicoder:
> New lines missing at the end of both files

Nuts! I always forget that, will address it now.

Fechulo (approved):

PR #1807, opened 2021-04-09
Add a bunch of light aircraft, GA fleet updates

### LINKED ISSUE #1806: Add King Air Series
The King Air series is legendary in general aviation! In OpenScope, we only have the King Air 350, so there's no reason not to include the rest of the family! I'll also try to add other important aircraft- I haven't seen the TBM or Baron, so those will probably be added. I'll also add them to US general aviation.

Hawaiicoder:
Resolves #1806 

The purpose of this pull request is to fix a lack of light aircraft in the sim which will be added to GA fleets.

### DIFF
--- a/assets/aircraft/BE9T.json
+++ b/assets/aircraft/BE9T.json
@@ -1,0 +1,37 @@
+{
+    "name": "Beech King Air",
+    "icao": "BE9T",
+    "engines": {
+        "number": 2,
+        "type": "T"
+    },
+    "weightClass": "S",
+    "category": {
+        "srs": 2,
+        "lahso": 7,
+        "recat": "F"
+    },
+    "ceiling": 31000,
+    "rate": {
+        "climb":      2380,
+        "descent":    1500,
+        "accelerate": 4,
+        "decelerate": 3
+    },
+    "runway": {
+        "takeoff": 0.845,
+        "landing": 1.613
+    },
+    "speed":{
+        "min":     77,
+        "landing": 100,
+        "cruise":  252,
+        "cruiseM": null,
+        "max":     265,
+        "maxM":    null
+    },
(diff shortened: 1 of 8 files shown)


assets/aircraft/BE9T.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "Beech King Air",
```
Marcel510, line 2:
```suggestion
    "name": "Beech King Air 90",
```
or
```suggestion
    "name": "Beech 90 King Air",
```

I'd suggest adding the "90" to the name, since Beech King Air is a whole aircraft family.

assets/aircraft/be9l.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "Beech King Air",
```
Marcel510, line 2: Same as for "be9t"

assets/aircraft/dh2t.json
```diff
@@ -0,0 +1,37 @@
+{
+    "name": "de Havilland Canda DHC-2 Mk3 Turbo Beaver",
```
Marcel510, line 2:
```suggestion
    "name": "de Havilland Canada DHC-2 Mk3 Turbo Beaver",
```

Marcel510 (changes requested):
@Hawaiicoder 
Awesome addition, really looking forward to having a greater variety of GA aircraft in the sim 👍 
Which source did you use for the performance data? Usually the Eurocontrol Aircraft Performance Database is not too bad but it doesnt include many GA aircraft :(
Other GA airlines should be updated at some point, too. But I think that's outside the scope for this PR and should be done separately.

Fechulo (approved):

PR #1824, opened 2021-05-29
feature/1405 - Implement automatic departures

### LINKED ISSUE #1405: Automatic GND and TWR control
Try and introduce automatic taxi and takeoff instructions if the user enables the option.

erikquinn:
Resolves #1405.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftConflict.js
+++ b/src/assets/scripts/client/aircraft/AircraftConflict.js
@@ -126,8 +126,8 @@ export default class AircraftConflict {
 
         // TODO: replace magic numbers with enum
         // Ignore aircraft in the first minute of their flight
-        if (gameTime - this.aircraft[0].takeoffTime < 60 ||
-            gameTime - this.aircraft[1].takeoffTime < 60) {
+        if (gameTime - this.aircraft[0].takeoffTime < 90 ||
+            gameTime - this.aircraft[1].takeoffTime < 90) {
             return;
         }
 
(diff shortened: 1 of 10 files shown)


src/assets/scripts/client/aircraft/AircraftConflict.js
```diff
@@ -126,8 +126,8 @@ export default class AircraftConflict {
 
         // TODO: replace magic numbers with enum
         // Ignore aircraft in the first minute of their flight
```
n8rzz, line 128: is this comment still valid?

n8rzz (approved):

Fechulo (changes requested):
LOVE IT! :shipit: 

Couple of things I noticed though:

- A few departures spawned in with landing aircraft on the runway, resulting in:
image

- Increasing the departure rate results in a few radar contact losses...
image
- The conflict detection doesn't seem to be doing great (KLAX, normal rate)
image

- Questionable decisions with crossing runways:
image
image

I'll keep playing with it and see if I find anything else, but I'm very excited to have this in the sim 😄

erikquinn:
@Fechulo yep! the first I was actually able to fix (thought it'd be too hard, but a quick solution was actually not too bad).

The second one is kinda the best we can do atm I think-- I increased the invincibility period from 60 seconds from starting takeoff roll to 90 seconds-- that's the slight delay you're seeing. Can be reduced if you think that's appropriate. It's just hard to get the tower to space them on their own.

Regarding crossing runways, that will need to be an enhancement to this feature. Just can't get something like that done rn. Will label the option appropriately so it's clear that this is an experimental feature. :)

Issue #1173, opened 2018-11-07
Lower-Case EVERYTHING


erikquinn:
Throughout the codebase, we store values of various data (routes, callsigns, fixes and whatnot) in varying cases. For consistency, we should stick to storing and manipulating only in lower case. This includes with routes.

At times when we receive data from an airport file or from user input which may be upper or lower case, we should force it to lower case before storing it.

At times when we need to display these values to the user in upper case, we should do a `.toUpperCase()` on the view only.

This should moderately decrease the overall number of case changes, and dramatically decrease the confusion on whether a case change is redundant or necessary.

felixscheffer:
Just wondering, is there a specific reason why we are favoring lower-case over upper-case?

PR #1831, opened 2021-06-22
bugfix/1258 - Fix STAR command parsing

### LINKED ISSUE #1258: STAR command cannot resolve route
Browser:
Chrome

Airport:
LKPR

Immediate

#### Issue Description
It is not possible to change assigned STAR for an aicraft using star command, for example "star vlm.vlm3r.lkpr30". It replies with "route of vlm.vlm3r.lkpr30 is not valid" error message and throws an exception to Console:

RouteModel.js:765 TypeError: Expected airway or procedure name, but we can't ' +
            'determine what kind of leg vlm3r is
    at LegModel.value (LegModel.js:289)
    at LegModel.value (LegModel.js:240)
    at new LegModel (LegModel.js:105)
    at RouteModel.value (RouteModel.js:763)
    at Fms.value (Fms.js:842)
    at Pilot.value (Pilot.js:271)
    at AircraftCommander.value (AircraftCommander.js:392)
    at AircraftCommander.value (AircraftCommander.js:157)
    at AircraftCommander.value (AircraftCommander.js:64)
    at InputController.value (InputController.js:748)

Reason: Mixed case in command and airport definition. 
Detailed analysis: In LegModel::init, _determineLegType() is called with route parameter in lowercase (vlm.vlm3r.lkpr30), passing the parameter to NavigationLibrary.hasProcedure(), which makes case sensitive search in airport star definitions - but the airoport data is defined in upper case. Therefore, the procedure name is not recognized and the exception is thrown.
Proposed solution: Convert route string to uppercase in AicraftCommander::runSTAR()

#### Steps to reproduce
- start LKPR airport
- wait until first incoming flight
- issue "STAR vlm.vlm3r.lkpr30" command

JPNite3:
Resolves #1258 

The purpose of this pull request is to fix the STAR command so that it parses the ENTRY.STAR.EXIT format correctly

### DIFF
--- a/src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
+++ b/src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
@@ -793,7 +793,7 @@ export default class RouteModel extends BaseModel {
         let starLegModel;
 
         try {
-            starLegModel = new LegModel(routeString);
+            starLegModel = new LegModel(routeString.toUpperCase());
         } catch (error) {
             console.error(error);
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -7351,7 +7351,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "aproba": {
               "version": "1.2.0",
@@ -7394,7 +7395,8 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
             },
             "concat-map": {
               "version": "0.0.1",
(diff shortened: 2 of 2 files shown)


src/assets/scripts/client/aircraft/FlightManagementSystem/RouteModel.js
```diff
@@ -793,7 +793,7 @@ export default class RouteModel extends BaseModel {
         let starLegModel;
 
         try {
-            starLegModel = new LegModel(routeString);
+            starLegModel = new LegModel(routeString.toUpperCase());
```
n8rzz, line 796: I'd be careful about making a change like this.  What problem is this solving exactly?  Are there tests you could add to both reproduce the problem and also prove it has been solved?

String comparisons should ideally be done as lowercase, so if anything this should be `toLowerCase()` instead.

JPNite3: Yeah, so this technically is not really solving it. Originally the STAR command does not work because when creating an instance of the LegModel (to apply the new STAR entered by the user), _determineLegType seems to not be able to find the STAR name when it's in lower case. This _technically_ fixes the issue but I understand I need to look more into trying and lowercasing stuff as #1173 says.

PR #1849, opened 2021-11-02
User selectable wind settings

### LINKED ISSUE #1847: Dynamic Wind
Enhancement to allow the user to change the wind direction and speed.

CanisLupus518:
Resolves #1847

The purpose of this pull request is to modify the Traffic Rate dialog to allow the user to select the current wind direction and speed used in the sim.


src/assets/scripts/client/info/AirportInfoController.js
```diff
@@ -258,6 +261,17 @@ export default class AirportInfoController {
...
 
+    /**
+     * Updates the wind information
+     *
+     * @for AirportInfoController
+     * @method onWindChange
+     */
+    onWindChange(currentWind) {
+        this.wind = this._buildWindAndGustReadout({ speed: currentWind.speed, angle: currentWind.angle });
+        this._render();
```
n8rzz, line 272: missing empty line between assignment ad method call

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -226,6 +251,34 @@ export default class TrafficRateController {
...
+     */
+    _buildWindSlider(key, label, value, onChangeMethod, min, max, step) {
+        const template = `
+            <div class="form-element">
+                <div class="form-label">${label}</div>
+                <input class="form-slider" type="range" name="${key}" value="${value}" min="${min}" max="${max}" step="${step}" />
+                <span class="form-value">${value}</span>
+            </div>`;
+        const $element = $(template);
+        const onChangeHandler = onChangeMethod.bind(this);
```
n8rzz, line 275: please bind this to the class as a handler within the `_setupHandlers` method:
```js
_setupHandlers() {
  this._onChangeWindDirectionHandler = this.onChangeWindDirection.bind(this);

  return this;
}
```

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -252,6 +305,38 @@ export default class TrafficRateController {
...
+     * @for TrafficRateController
+     * @method _onChangeWindDirection
+     * @param event
+     */
+    _onChangeWindDirection(event) {
+        const $target = $(event.target);
+        const $output = $target.next(`.${CLASSNAMES.FORM_VALUE}`);
+        const value = $target.val();
+        this._wind.angle = Number(value);
+        $output.text(value);
```
n8rzz, line 320: missing empty line between assignment and method call

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -252,6 +305,38 @@ export default class TrafficRateController {
...
+    /**
+     * Called when the wind direction is changed
+     *
+     * @for TrafficRateController
+     * @method _onChangeWindDirection
+     * @param event
+     */
+    _onChangeWindDirection(event) {
+        const $target = $(event.target);
+        const $output = $target.next(`.${CLASSNAMES.FORM_VALUE}`);
```
n8rzz, line 317: is there a more descriptive name we could use here?  `$output` doesn't mean much here and would be confusing for the person in this code.  perhaps `nextWindDirection` or `nextWindAngle` could work?

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -174,6 +185,20 @@ export default class TrafficRateController {
         this._rates = {};
         this._elements = {};
 
+
+        const airport = AirportController.airport_get();
+        this._wind = { speed: airport.wind.speed, angle: Math.round(radiansToDegrees(airport.wind.angle)) };
+
+
+        const $windDirSlider = this._buildWindSlider('wind direction', 'wind direction', this._wind.angle,
+            this._onChangeWindDirection, 10, 360, 10);
```
n8rzz, line 194: see comment below, but `_onChangeWindDirection` doesn't need to be passed here since it's already on the class.

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -226,6 +251,34 @@ export default class TrafficRateController {
...
+     * @method _buildWindSlider
+     * @param key {string}
+     * @param value {number} passed to the change handler
+     * @param onChangeMethod {function}
+     * @param min {int} min value of slider
+     * @param max {int} max value of slider
+     * @param step {int} step value of slider
+     * @return {jquery|HTML Element}
+     */
+    _buildWindSlider(key, label, value, onChangeMethod, min, max, step) {
```
n8rzz, line 267: remove `onChangeMethod` param

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -252,6 +305,38 @@ export default class TrafficRateController {
...
+    /**
+     * Called when the wind speed is changed
+     *
+     * @for TrafficRateController
+     * @method _onChangeWindSpeed
+     * @param event
+     */
+    _onChangeWindSpeed(event) {
+        const $target = $(event.target);
+        const $output = $target.next(`.${CLASSNAMES.FORM_VALUE}`);
```
n8rzz, line 333: is there a more descriptive name we could use here?

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -226,6 +251,34 @@ export default class TrafficRateController {
...
+     * @param value {number} passed to the change handler
+     * @param onChangeMethod {function}
+     * @param min {int} min value of slider
+     * @param max {int} max value of slider
+     * @param step {int} step value of slider
+     * @return {jquery|HTML Element}
+     */
+    _buildWindSlider(key, label, value, onChangeMethod, min, max, step) {
+        const template = `
+            <div class="form-element">
```
n8rzz, line 269: take a look at the `SELECTORS` constant we have setup.  for something like this, where will be accessing this later via js, we should add an additional class prefixed with `js`.  in this case maybe something like `js-windSlider` or something similar.

The reason for that is `.form-element` could be used on other forms that might be in the DOM.  using a specific `js` prefixed classname makes it obvious that A) it isn't something that is styled, B) it will be used by js at some point and C) it's unique

n8rzz (changes requested):

src/assets/scripts/client/ui/TrafficRateController.js
CanisLupus518: @n8rzz I followed the same pattern used here for the rate sliders, since the direction and speed handlers are different events. this prevents duplicate code by allowing me to pass the handler into the creation of the element as a param. I was under the impression based on patterns used elsewhere that _setupHandlers() was meant for eventBus events, and not for gui element events. If I make the proposed changes, I would end up having to add another buildXSlider method for the speed. Thoughts?

src/assets/scripts/client/ui/TrafficRateController.js
CanisLupus518: @n8rzz I am happy to make these changes, but I just want to point out that I was following the pattern used nearby in the same file to create the traffic rate sliders. Should I correct those as well?

src/assets/scripts/client/ui/TrafficRateController.js
n8rzz: `setupHandlers` is there to bind any handlers to `this`.  those handlers might be used as EventBus callbacks, dom event handlers, etc.  Its a single place to do all the binding.

I see what you mean, though, we are doing it like this elsewhere.  gross.  I'll leave it up to you; if you want to make the change, awesome.  If not, we can leave it.

src/assets/scripts/client/ui/TrafficRateController.js
n8rzz: see above

n8rzz (approved):
:shipit:

erikquinn (approved):

PR #1855, opened 2021-11-10
feature/1854 Support configuration of spawn-time commands in SpawnPattern

### LINKED ISSUE #1854: Support runway/fix specific tower assigned headings
There are some airports, like KPHL and the PHL2 SID, that assign departure headings to aircraft based on the exit fix they have filed as well as the runway they are using. Currently this has been baked into the SID but only works for the default runway configuration in the airport file and would not behave well if the runway config is changed with #34 

I propose a new tower section in the airport cfg where departure headings can be specified unique to runway and/or SID. This would also enable support for smaller airports that don’t have SIDS to be able to define tower assigned headings.

CanisLupus518:
Resolves #1854 

The purpose of this pull request is to expand the SpawnPattern to allow per runway commands to be sent to an aircraft at spawn time. This should support both departures and arrivals, and remove the need for pseudo-instructions\restrictions to be baked into SIDS and STARs to get aircraft to enter the airspace correctly.


CanisLupus518:
@erikquinn in your comment here https://github.com/openscope/openscope/issues/1854#issuecomment-963570858
you specify an intermediate property for runway. Is that necessary? do you anticipate also having commands not associated with a runway? (possibly thinking overflights?) If the intermediate property does need to be there, any objection to just using "Runway" for both arrival and departure patterns instead of making a property for each, since the patter itself is only one or the other

src/assets/scripts/client/trafficGenerator/spawnPatternModelJsonValidator.js
```diff
@@ -18,7 +18,8 @@ const ACCEPTED_KEYS = [
 const ACCEPTED_OPTIONAL_KEYS = [
     'offset',
     'period',
-    'variation'
+    'variation',
+    'commands'
```
n8rzz, line 22: please alphabetize this list

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -560,6 +589,11 @@ export default class AircraftController {
             aircraftModel.fms.departureRunwayModel.addAircraftToQueue(aircraftModel.id);
             aircraftModel.setFlightPhase(FLIGHT_PHASE.WAITING);
             aircraftModel.shouldTakeOffWhenRunwayIsClear = true;
+
+            if (runwayCommands && runwayCommands[aircraftModel.fms.departureRunwayModel.name]) {
+                const commandString = runwayCommands[aircraftModel.fms.departureRunwayModel.name];
+                this._runCommandOnNewAircraft(aircraftModel, commandString);
```
n8rzz, line 595: missing new line between assignment and method call

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -546,12 +568,19 @@ export default class AircraftController {
...
         const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
+        const runwayCommands = initializationProps.commands;
 
         // triggering event bus rather than calling locally because multiple classes
         // are listening for the event and aircraft model
         this._eventBus.trigger(EVENT.ADD_AIRCRAFT, aircraftModel);
 
+        if (isArrival && runwayCommands && runwayCommands[aircraftModel.fms.arrivalRunwayModel.name]) {
+            const commandString = runwayCommands[aircraftModel.fms.arrivalRunwayModel.name];
+            this._runCommandOnNewAircraft(aircraftModel, commandString);
```
n8rzz, line 581: missing new line between assignment and method call

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -546,12 +568,19 @@ export default class AircraftController {
...
         const isDeparture = initializationProps.category === 'departure';
+        const isArrival = initializationProps.category === 'arrival';
         const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
+        const runwayCommands = initializationProps.commands;
 
         // triggering event bus rather than calling locally because multiple classes
         // are listening for the event and aircraft model
         this._eventBus.trigger(EVENT.ADD_AIRCRAFT, aircraftModel);
 
+        if (isArrival && runwayCommands && runwayCommands[aircraftModel.fms.arrivalRunwayModel.name]) {
```
n8rzz, line 579: is this needed?  see comment at `._runCommandOnNewAircraft()`

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -560,6 +589,11 @@ export default class AircraftController {
             aircraftModel.fms.departureRunwayModel.addAircraftToQueue(aircraftModel.id);
             aircraftModel.setFlightPhase(FLIGHT_PHASE.WAITING);
             aircraftModel.shouldTakeOffWhenRunwayIsClear = true;
+
+            if (runwayCommands && runwayCommands[aircraftModel.fms.departureRunwayModel.name]) {
```
n8rzz, line 593: is this needed?  see comment at `._runCommandOnNewAircraft()`

erikquinn:
It was intentional, but may not be necessary. We tend to think really defensively, since we've gotten burned before on making assumptions of "nobody would ever need that", and then it turns out like all of Asia does, or the British needed it lol it's just my default state of being 😛 

I hadn't taken the time until now to ponder whether a use case exists to specify instructions based on criteria _other_ than runway. After thinking it through, I don't think any really exist other than _possibly_ aircraft type? Like maybe you'd want props/jets on different headings/altitudes. But this could (a) be added later as an enhancement to the feature, or (b) be handled without additional work by separating the spawn patterns.

So in the end, I agree that we should be fine with a simple `commands: { '8R': 'do this!', ... }` concept. :shipit:

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -618,11 +652,29 @@ export default class AircraftController {
...
     }
 
+    /**
+     * Execute a command on a new or preSpawned aircraft, ignoring controllability flag
+     *
+     * @param aircraft {AircraftModel}
+     * @param command {String}
+     * @private
+     */
+    _runCommandOnNewAircraft(aircraft, commandString) {
```
n8rzz, line 668: `_runCommandOnNewAircraft` -> `_runCommandOnPreSpawnAircraft`

There is some logic above that looks to be duplicated.  What if we called this method for _every_ prespawn aircraft?  If an aircraft doesn't have commands to run, then we return early.  If they do, we proceed and run commands

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -204,6 +217,15 @@ export default class AircraftController {
      */
     addItem = (item) => this.aircraft.list.push(item);
 
+    /**
+     * Get the instance of AircraftCommander that was created by this Controller
+     *
+     * @returns {AircraftCommander}
+     */
+    getAircraftCommander() {
```
n8rzz, line 225: I'd encourage moving this to a getter:

```suggestion
get aircraftCommander() {
    return ths._aircraftCommander;
}
```

if you do, make sure it's grouped with other `getter`s in this file (if any).

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -687,8 +689,8 @@ export default class AircraftCommander {
         }
 
         if (spotInQueue > 0) {
-            readback.log = `number ${spotInQueue} behind ${aircraftAhead.callsign}`;
-            readback.say = `number ${spotInQueue} behind ${aircraftAhead.getRadioCallsign()}`;
+            readback.log = `number ${spotInQueue} for departure`;
```
n8rzz, line 692: nice!

src/assets/scripts/client/InputController.js
```diff
@@ -41,15 +41,15 @@ export default class InputController {
...
+    constructor($element, aircraftController, scopeModel) {
         this.$element = $element;
         this.$body = null;
         this.$window = null;
         this.$commandInput = null;
         this.$canvases = null;
 
         this._eventBus = EventBus;
-        this._aircraftCommander = aircraftCommander;
+        this._aircraftCommander = aircraftController.getAircraftCommander();
```
n8rzz, line 52: I don't think we need this.  we could access the commander via `this._aircraftController.aircraftCommander.runCommands()`, provided we move that to a getter (see below)

n8rzz (changes requested):
This looks really good!  I've left mostly some nit-picky comments, but there are a few logic related ones in there too.  feel free to reach out if you have any questions!

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -128,14 +128,17 @@ export default class AircraftCommander {
...
+            if (!isPreSpawn) {
+                UiController.ui_log(`${aircraft.callsign}, ${r_log} ${response_end}`, redResponse);
+                speech_say(
+                    [
+                        { type: 'callsign', content: aircraft },
+                        { type: 'text', content: `${r_say} ${response_end}` }
+                    ],
+                    aircraft.pilotVoice
+                );
+            }
```
erikquinn, line 141: As you know, we skip this when we're prespawning, because we don't display the response. Rather than wrapping this in an if statement, since there's nothing else after this section, we should instead return early before it.

In fact, we can also skip some stuff above this, so let's do that! After the second for loop, at line 116, we can add `if (preSpawn) { return; }`, and then this section we can just continue inline without the if.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -687,8 +689,8 @@ export default class AircraftCommander {
         }
 
         if (spotInQueue > 0) {
-            readback.log = `number ${spotInQueue} behind ${aircraftAhead.callsign}`;
-            readback.say = `number ${spotInQueue} behind ${aircraftAhead.getRadioCallsign()}`;
+            readback.log = `number ${spotInQueue} for departure`;
+            readback.say = `number ${spotInQueue} for departure`;
```
erikquinn, line 693: To preserve this functionality (since it is quite useful), @n8rzz would you be opposed to the slight ickiness of the following resolution below?

image

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -204,6 +217,15 @@ export default class AircraftController {
...
+        return this._aircraftCommander;
+    }
```
erikquinn, line 227: This is a good opportunity for a _getter_:
```javascript
get aircraftCommander() {
    return this._aircraftCommander;
}
```

which is accessed just like a property, ie `this._aircraftController.aircraftCommander`. Any time you are simply accessing something that you don't want to be reassigned, using a getter like this is the way to go.

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -618,11 +652,29 @@ export default class AircraftController {
...
+     * @private
+     */
+    _runCommandOnNewAircraft(aircraft, commandString) {
+        const command = new CommandParser(`${aircraft.getCallsign()} ${commandString}`).parse();
+        const defaultControllable = aircraft.isControllable;
+        aircraft.isControllable = true;
+
+        this._aircraftCommander.runCommands(aircraft, command.args, true);
+
+        aircraft.isControllable = defaultControllable;
```
erikquinn, line 675: This just forces all aircraft to be controllable, which we don't want. If we did, we'd do it a different way. Should be handled within `AircraftCommander.runCommands()` as I described in a previous comment, and removed here.

erikquinn (changes requested):

src/assets/scripts/client/AppController.js
```diff
@@ -188,8 +185,8 @@ export default class AppController {
         UiController.init(this.$element);
 
         this.canvasController = new CanvasController(this.$canvasesElement, this.aircraftController, this.scopeModel);
-        this.aircraftCommander = new AircraftCommander(this.aircraftController, this.aircraftController.onRequestToChangeTransponderCode);
```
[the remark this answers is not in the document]
erikquinn: Actually, if we're treating it as a sub-part of the aircraftController, it seems fine to do this-- we can still use it via `AppController.aircraftController.aircraftCommander` (when the .getAircraftCommander() method is changed to a getter, anyway).

Resolving this one.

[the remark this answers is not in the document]
CanisLupus518: @erikquinn  I removed it because it is no longer used anywhere. It doesn't need to be passed into the `InputController` anymore because it can be retrieved from the `AircraftController` with a getter (after comments above), and it is now instantiated by the `AircraftController`. Nothing else in the `AppController` appeared to need a reference to it.

CanisLupus518:
@erikquinn should I also request a review from airport_reviewers since I modified KPHL to demo the feature?

erikquinn:
> @erikquinn should I also request a review from airport_reviewers since I modified KPHL to demo the feature?

Nah, not for now. They're welcome to weigh in, but I don't expect there to be anything of concern from this PR.

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -30,20 +30,21 @@ import { heading_to_string, radiansToDegrees, degreesToRadians } from '../utilit
  * @class AircraftCommander
  */
 export default class AircraftCommander {
-    constructor(aircraftController, onChangeTransponderCode) {
+    constructor(aircraftController) {
         this._eventBus = EventBus;
-        this._aircraftController = aircraftController;
-        this._onChangeTransponderCode = onChangeTransponderCode;
+        this._findAircraftById = aircraftController.findAircraftById.bind(aircraftController);
```
n8rzz, line 35: is this binding actually needed?  if it is, could we bind these within the `aircraftController` instead?

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -30,20 +30,21 @@ import { heading_to_string, radiansToDegrees, degreesToRadians } from '../utilit
  * @class AircraftCommander
  */
 export default class AircraftCommander {
-    constructor(aircraftController, onChangeTransponderCode) {
+    constructor(aircraftController) {
```
n8rzz, line 33: hmm, I'm torn.  should we pass an entire instance of the `AircraftController` or just references to the two methods we need a few lines below?  what do you think?

src/assets/scripts/client/aircraft/AircraftCommander.js
CanisLupus518: @n8rzz I'm not sure, I was just following @erikquinn's sugestion.

src/assets/scripts/client/aircraft/AircraftCommander.js
CanisLupus518: @n8rzz my personal preference would be to pass in the 2 methods. it would also make the fixture for the test a bit simpler

n8rzz: yeah, unless @erikquinn has any concerns, let's do that.  Its way cleaner that way.

CanisLupus518: @n8rzz this change is in and awaiting review

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: When I suggested it, it needed the binding. Might be doable from the aircraftController, I don't know, haven't tried it.

src/assets/scripts/client/aircraft/AircraftCommander.js
erikquinn: It was being passed in order to do the binding. It makes sense to me to bind here rather than the aircraftController, because the reason is obvious here, but not obvious in the aircraftController.

But if everything works doing it the other way, I'm not opposed.

CanisLupus518: @n8rzz if I do the binding in the controller, is it acceptable to do it as part of the call to the `Commander` constructor? Or does it need to go into `_setupHandlers()` and then the creation of the commander would need to be after that.

`this._aircraftCommander = new AircraftCommander(this.onRequestToChangeTransponderCode.bind(this), this.findAircraftById.bind(this));`

erikquinn: It's not a handler used by this class, so it should be fine just like that. Seems more sensible in the aircraftController than I thought it would be, so I'm fully satisfied with that strategy after verifying that it works :smiley:

n8rzz (approved):
Great work!!

src/assets/scripts/client/aircraft/AircraftCommander.js
CanisLupus518: @erikquinn I made this last change, let me know if there is anything else.

src/assets/scripts/client/trafficGenerator/spawnPatternModelJsonValidator.js
erikquinn: done

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: done

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: done

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: renamed and changed to return-early style

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: moved to the proper location (right beneath constructor).

src/assets/scripts/client/InputController.js
erikquinn: done

documentation/spawnPatternReadme.md
```diff
@@ -28,6 +32,10 @@ Lets look at some examples before we continue:
     "destination": "KLAS",
     "category": "arrival",
     "route": "BETHL.GRNPA1.KLAS",
+    "commands": {
+            "19L": "cross TOROO A100",
+            "1R": "cross TOROO A100"
+    },
```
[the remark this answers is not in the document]
erikquinn: done

src/assets/scripts/client/aircraft/AircraftCommander.js
```diff
@@ -41,8 +40,9 @@ export default class AircraftCommander {
      * @method runCommands
      * @param aircraft {AircraftModel}
      * @param commands {array<AircraftCommandParser>}
+     * @param isPreSpawn {boolean}
      */
-    runCommands(aircraft, commands) {
+    runCommands(aircraft, commands, isPreSpawn = false) {
         if (!aircraft.isControllable) {
             return true;
         }
```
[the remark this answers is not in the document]

src/assets/scripts/client/aircraft/AircraftCommander.js

src/assets/scripts/client/aircraft/AircraftCommander.js

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: done

src/assets/scripts/client/aircraft/AircraftController.js
erikquinn: done

erikquinn (approved):
Nice work! 🎉 

Do note that I made some changes to KPHL-- namely moving the JIIMS/PAATS HPAS back down to the appropriate altitudes of 80/90 respectively. I understand it's because of the control granted to PHLz by ZDC that means you should be able to turn/descend them in the holding pattern airspace... but this isn't how we can support that. We instead should add a feature which releases control to you in a specified way, rather than faking the altitudes to "make it work for now" (see #1856 for enabling early transfers of control). I also added some crossing restrictions _before_ the ones that enter the approach airspace for added realism, and recorded a bug observed with it (#1857).

As an alternative to #1856, I would also be willing to accommodate a "F this, give me the planes" button-- something like "All arrivals are my control", where it defaults to "off" and you can turn it "on", to allow manipulation of aircraft who have not reached the airspace. This is a lesser solution obviously, but is a shorter path we'd be more willing to accept than false manipulation of airspace.

Great job on this one-- very excited for the doors it opens for other airports! 😄

:shipit:

PR #1886, opened 2021-12-28
Enhancement/1702 Adjust HHOOD4.JKNOX arrival restrictions

### LINKED ISSUE #1702: KPDX HHOOD4 arrivals often too high
Browser: Google Chrome

Airport: KPDX (Portland, OR, USA)

Approximate play time before issue: After first "wave" of arrivals

#### Issue Description
Arrivals are consistently too high. In particular, many aircraft on the HHOOD4 don't even enter the airspace until they're over the airport. Very few meet the 6700' crossing restriction at SHAFR.

#### Steps to reproduce
- Play past the first "wave" of arrivals
- Also observed by executing "timewarp 50" and watching the altitudes of HHOOD4 arrivals

Would this have something to do with aircraft spawning too close to the airport? Maybe increasing the spawn distance would give aircraft more time to get on profile.

Flyboy31:
Added HHOOD and SHAFR waypoints along with their speed and altitude restrictions. They were missing prior, most likely causing the aircraft to be high above 6,700ft at SHAFR.

Resolves #1702

The purpose of this pull request is to fix KPDX's HHOOD4 arrivals often flying too high and not meeting speed or altitude restrictions.

### DIFF
--- a/assets/airports/kpdx.json
+++ b/assets/airports/kpdx.json
@@ -426,7 +426,7 @@
             "entryPoints": {
                 "BAYYR": [["BAYYR", "A130+|S270"]],
                 "DUFUR": ["DUFUR", ["RABBI", "A130+|S270"]],
-                "JKNOX": [["JKNOX", "A240+"], ["MYTOE", "A170+"], ["BAYYR", "A130+|S270"]],
+                "JKNOX": [["JKNOX", "A240+"], ["MYTOE", "A170+"], ["BAYYR", "A130+|S270"], ["HHOOD", "A90+|A150-|S250"], ["SHAFR", "A67|S210"]],
                 "JORAD": [["JORAD", "A240+"], ["LUUND", "A170+"], ["YENTL", "A130+|S270"]],
                 "JOTBA": [["JOTBA", "A240+"], ["SODDA", "A170+"], "DUFUR", ["RABBI", "A130+|S270"]],
                 "RABBI": [["RABBI", "A130+|S270"]],


Flyboy31:
Setting the HHOOD altitude crossing restriction at A90 (which is the minimum altitude for the HHOOD crossing) seems to have fixed the issue. It appears as if the aircraft are struggling to get down to 6,700ft on time. The reason seems to be they are crossing HHOOD at the highest possible altitude, 15,000ft, and not giving themselves enough time to descend. My fix should work for the moment but it should be looked at again.

Fechulo (changes requested):
That's not gonna work. The restrictions are all already there, you can see them in the body and and rwy sections on lines >435 

If aircraft aren't making the restrictions, it's probably due to aircraft performance

Flyboy31:
I've had to remove the HHOOD and SHAFR restrictions from the entrypoints as it seems the aircraft would pass HHOOD, move onto BLRUN, then read HHOOD again and turn around, pass HHOOD, then turn again and proceed direct SHAFR. At the moment this seems to have fixed the issue and the aircraft are crossing HHOOD at the lowest possible altitude restriction and crossing SHAFR at the altitude restriction of 6,700ft.

Fechulo:
> I've had to remove the HHOOD and SHAFR restrictions from the entrypoints as it seems the aircraft would pass HHOOD, move onto BLRUN, then read HHOOD again and turn around, pass HHOOD, then turn again and proceed direct SHAFR. 

That's what my original comment was referring to, yes. 

Regarding the current solution, I'm not sure I like it, since it's deviating from the real procedure. We can look into other alternatives if it's a real issue, maybe giving a lower speed to aircraft on the arrival when they spawn?

Flyboy31:
> > I've had to remove the HHOOD and SHAFR restrictions from the entrypoints as it seems the aircraft would pass HHOOD, move onto BLRUN, then read HHOOD again and turn around, pass HHOOD, then turn again and proceed direct SHAFR.
> 
> Yeah sorry, that's what my original comment was referring to. I shouldn't have written a review before you requested it, my bad
> 
> Regarding the current solution, I'm not sure I like it, since it's deviating from the real procedure. We can look into other alternatives if it's a real issue, maybe giving a lower speed to aircraft on the arrival when they spawn?

We can absolutely do that. I believe the speed on that arrival is 320 but I can reduce it to 280 and see if that fixes it.

Flyboy31:
After changing the speed, we're still encountering the same issue. I did test slowing the aircraft down to 210kts once it enters controllable airspace and it performed much better, 800ft lower. If there's a way to tell the aircraft to slow down after passing over HHOOD then that would fix the issue.

Fechulo (changes requested):
Reducing the spawn speed to 280kts seems like a reasonable solution for now

PR #1896, opened 2022-01-04
Enhancement/1345 Add reset buttons to traffic config dialog

### LINKED ISSUE #1345: Reset button for traffic scaling menu
Traffic scaling menu being added #1192 could use a [reset button ](https://github.com/openscope/openscope/pull/1192#issuecomment-473666917)that would
* put sliders back to 1/1/1
* revert spawn rates to defaults for the loaded airport

CanisLupus518:
Resolves #1345 

The purpose of this pull request is to add two buttons to the traffic dialog. 1 to reset the traffic settings to the "default" config in the airport file. Another to reset\restart the generated traffic based on the new settings in the dialog.

### DIFF
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -313,7 +313,7 @@ export default class AppController {
         this.aircraftController.aircraft_remove_all();
         this.scopeModel.radarTargetCollection.reset();
         SpawnScheduler.createPreSpawnDepartures();
-        SpawnPatternCollection.resetAirborneTraffic(this.aircraftController);
+        SpawnScheduler.resetAirborneTraffic();
     }
 
     // TODO: this should live in a view class somewhere. temporary inclusion here to prevent tests from failing
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -23,6 +23,7 @@ import { speech_init } from './speech';
 import { EVENT } from './constants/eventNames';
 import { SELECTORS } from './constants/selectors';
 import { TRACKABLE_EVENT } from './constants/trackableEvents';
+import _forEach from "lodash/forEach";
 
 /**
  * Root controller class
@@ -82,6 +83,7 @@ export default class AppController {
      */
     setupHandlers() {
         this.onAirportChangeHandler = this.onAirportChange.bind(this);
+        this.onTrafficResetHandler = this.onTrafficReset.bind(this);
 
         return this;
     }
@@ -93,6 +95,7 @@ export default class AppController {
      */
     enable() {
         this._eventBus.on(EVENT.AIRPORT_CHANGE, this.onAirportChangeHandler);
+        this._eventBus.on(EVENT.TRAFFIC_RESET, this.onTrafficResetHandler);
 
         return this;
     }
@@ -104,6 +107,7 @@ export default class AppController {
      */
     disable() {
         this._eventBus.off(EVENT.AIRPORT_CHANGE, this.onAirportChangeHandler);
+        this._eventBus.off(EVENT.TRAFFIC_RESET, this.onTrafficResetHandler);
 
         return this.destroy();
     }
@@ -297,6 +301,25 @@ export default class AppController {
         SpawnScheduler.startScheduler();
 
         this.updateViewControls();
+    }
+
+    /**
+     * event callback fired from within the `TrafficRateController` when traffic should be cleared and respawned.
+     *
+     * @for AppController
+     * @method onTrafficReset
+     */
+    onTrafficReset() {
+        EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'traffic-reset', AirportController.current.icao);
+        this.aircraftController.aircraft_remove_all();
+        this.scopeModel.radarTargetCollection.reset();
+        SpawnScheduler.createPreSpawnDepartures()
+
+        _forEach(SpawnPatternCollection.spawnPatternModels, (spawnPattern) => {
+            spawnPattern.preSpawnAircraftList = []
+            spawnPattern.createPreSpawnAircraft(this.aircraftController)
+            SpawnScheduler.resetTimer(spawnPattern)
+        });
     }
 
     // TODO: this should live in a view class somewhere. temporary inclusion here to prevent tests from failing
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -23,7 +23,6 @@ import { speech_init } from './speech';
 import { EVENT } from './constants/eventNames';
 import { SELECTORS } from './constants/selectors';
 import { TRACKABLE_EVENT } from './constants/trackableEvents';
-import _forEach from "lodash/forEach";
 
 /**
  * Root controller class
@@ -313,13 +312,8 @@ export default class AppController {
         EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'traffic-reset', AirportController.current.icao);
         this.aircraftController.aircraft_remove_all();
         this.scopeModel.radarTargetCollection.reset();
-        SpawnScheduler.createPreSpawnDepartures()
-
-        _forEach(SpawnPatternCollection.spawnPatternModels, (spawnPattern) => {
-            spawnPattern.preSpawnAircraftList = []
-            spawnPattern.createPreSpawnAircraft(this.aircraftController)
-            SpawnScheduler.resetTimer(spawnPattern)
-        });
+        SpawnScheduler.createPreSpawnDepartures();
+        SpawnPatternCollection.resetAirborneTraffic(this.aircraftController);
     }
 
     // TODO: this should live in a view class somewhere. temporary inclusion here to prevent tests from failing
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -312,7 +312,7 @@ export default class AppController {
         EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'traffic-reset', AirportController.current.icao);
         this.aircraftController.aircraft_remove_all();
         this.scopeModel.radarTargetCollection.reset();
-        AirportController.current.departureRunwayModel.reset();
+        AirportController.current.resetAllRunwayQueues();
         SpawnScheduler.createPreSpawnDepartures();
         SpawnScheduler.resetAirborneTraffic();
     }
--- a/src/assets/scripts/client/AppController.js
+++ b/src/assets/scripts/client/AppController.js
@@ -312,6 +312,7 @@ export default class AppController {
         EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'traffic-reset', AirportController.current.icao);
         this.aircraftController.aircraft_remove_all();
         this.scopeModel.radarTargetCollection.reset();
+        AirportController.current.departureRunwayModel.reset();
         SpawnScheduler.createPreSpawnDepartures();
         SpawnScheduler.resetAirborneTraffic();
     }
--- a/src/assets/scripts/client/ui/TrafficRateController.js
+++ b/src/assets/scripts/client/ui/TrafficRateController.js
@@ -102,7 +102,8 @@ export default class TrafficRateController {
 
         this._buildDialogBody();
         this.$element.append(this.$dialog);
-        this.$dialog.find('#reset-button').click(this._onFormReset)
+        this.$dialog.find('#reset-button').click(this._onFormReset);
+        this.$dialog.find('#restart-button').click(this._onTrafficReset);
 
         return this;
     }
@@ -121,6 +122,7 @@ export default class TrafficRateController {
         this._onChangeWindDirection = this.onChangeWindDirection.bind(this);
         this._onChangeWindSpeed = this.onChangeWindSpeed.bind(this);
         this._onFormReset = this.onFormReset.bind(this);
+        this._onTrafficReset = this.onTrafficReset.bind(this);
 
         return this;
     }
@@ -186,7 +188,7 @@ export default class TrafficRateController {
      * Resets the traffic dialog and spawn rates to default values
      *
      * @for TrafficRateController
-     * @method onRateReset
+     * @method onFormReset
      */
     onFormReset() {
         const airport = AirportController.airport_get();
@@ -197,6 +199,16 @@ export default class TrafficRateController {
             spawnPattern.resetRate();
         });
         this._buildDialogBody();
+    }
+
+    /**
+     * Resets the generated traffic using the current traffic values
+     *
+     * @for TrafficRateController
+     * @method onTrafficReset
+     */
+    onTrafficReset() {
+        this._eventBus.trigger(EVENT.TRAFFIC_RESET);
     }
 
     /**
--- a/src/assets/scripts/client/ui/TrafficRateController.js
+++ b/src/assets/scripts/client/ui/TrafficRateController.js
@@ -22,8 +22,8 @@ const UI_TRAFFIC_MODAL_TEMPLATE = `
         <p class="dialog-title">Traffic rate</p>
         <div class="dialog-body nice-scrollbar"></div>
         <div class="dialog-footer form-element">
-            <button id="reset-button" class="button">Default</button>
-            <button id="restart-button" class="button">Restart</button>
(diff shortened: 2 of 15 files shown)


src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -115,6 +120,7 @@ export default class TrafficRateController {
         this._onAirportChangeHandler = this.onAirportChange.bind(this);
         this._onChangeWindDirection = this.onChangeWindDirection.bind(this);
         this._onChangeWindSpeed = this.onChangeWindSpeed.bind(this);
+        this._onFormReset = this.onFormReset.bind(this);
```
n8rzz, line 123: hmm, seems I've been sleeping on PRs in this file.  We typically add the word `handler` to the bound method. 
```js
this._onFormResetHandler = this.onFormReset.bind(this);
```

But I see the two methods above this do not follow this convention. I'd encourage you to make that change, but it isn't required.

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -176,6 +182,23 @@ export default class TrafficRateController {
...
+     *
+     * @for TrafficRateController
+     * @method onRateReset
+     */
+    onFormReset() {
+        const airport = AirportController.airport_get();
+        this._wind = { speed: airport.defaultWind.speed, angle: Math.round(radiansToDegrees(airport.defaultWind.angle)) };
+
+        this._eventBus.trigger(EVENT.WIND_CHANGE, this._wind);
+        _forEach(SpawnPatternCollection.spawnPatternModels, (spawnPattern) => {
```
n8rzz, line 196: I'd encourage moving this work to the `SpawnPatternCollection` itself and exposing a method `.resetSpawnPatternRates()` or something.

Generally, when dealing with a collection of items, we try to have the collection do the work and simply expose methods for consumers to call. That keeps everything centralized and makes it easier to track down where things are changing.

Additionally, when you do move this to the collection, feel free to use the built-in foreach instead of lodash. This app went through a heavy refactor when the built-ins weren't yet built-in, so we used lodash. We don't need that anymore (for this anyway).

src/assets/style/module/dialog/_dialog.less
```diff
@@ -84,6 +84,13 @@
     flex-grow: 1;
 }
 
+.form-element button {
```
n8rzz, line 87: could you target the `.button` classname here instead of the element?

src/assets/scripts/client/ui/TrafficRateController.js
```diff
@@ -98,6 +102,7 @@ export default class TrafficRateController {
 
         this._buildDialogBody();
         this.$element.append(this.$dialog);
+        this.$dialog.find('#reset-button').click(this._onFormReset)
```
n8rzz, line 105: ugh, bad Nate!  more sleeping on PRs. 

when we need to access DOM elements via js, usually, we use a classname prefixed with `js-`. That we we can easily tell its a hook for js events and shouldn't be used for styling. Additionally, this selector should probably live with the rest of them here: https://github.com/openscope/openscope/blob/master/src/assets/scripts/client/constants/selectors.js

n8rzz (changes requested):
This is looking pretty good! I'm most concerned about some minor things with `id` vs `js-` classname and css element vs classname rules. otherwise the rest are mostly nitpicks.

Let me know if you have any questions, and keep up the great work!

src/assets/scripts/client/ui/TrafficRateController.js
CanisLupus518: @n8rzz  I am using the element id here as the selector, should I be using a js- classname instead?

n8rzz: yes please

n8rzz: and then reference that using the `SELECTORS` constant from the file I linked above

CanisLupus518:
@erikquinn @n8rzz just FYI, it looks like this feature makes an existing problem a bit more exposed. The app performance drops significantly as the number of spawned aircraft increases. Historically when I use played around with the traffic generation sliders this was a very gradual issue that slowed the app down as more aircraft spawn. However, with this new button that resets traffic, a user can dial the arrivals up to 3x or 4x at major airports and experience instant significant performance issues. At higher multiples, the app crashes (at least on my machine in Chrome). This problem is exacerbated for airports that have arrivals that spawn very far out from the airspace, or where the 1x aircraft rate is already somewhat high.

I wasn't sure if you thought this might be a blocker for this PR.

CanisLupus518:
@n8rzz The last issue remaining in the lint check is a circular dependency issue between SpawnScheduler and SpawnPatternCollection. This arrived because I was trying to follow the guidance of "let the collection do the work", but in this case, the work needs to make a call to the SpawnScheduler. From my point of view, this particular instance of the circular dep is a non-risk since it doesn't seem to effect any init or startup. My options are 1) keep it as is and see if you are ok accepting the lint error or 2) move this particular method back into the AppController and loop through the collection there. Thoughts?

CanisLupus518:
I suppose one possible solution would be to move SPawnScheduler.resetTimer(spawnPattern) into SpawnPatternModel itself

n8rzz:
> @n8rzz The last issue remaining in the lint check is a circular dependency issue between SpawnScheduler and SpawnPatternCollection. This arrived because I was trying to follow the guidance of "let the collection do the work", but in this case, the work needs to make a call to the SpawnScheduler. From my point of view, this particular instance of the circular dep is a non-risk since it doesn't seem to effect any init or startup. My options are 1) keep it as is and see if you are ok accepting the lint error or 2) move this particular method back into the AppController and loop through the collection there. Thoughts?

My preference would be to avoid circular dependencies at all costs, though it sounds like this is low-risk. Still, circular dependencies have a tendency to cause problems at some point in their life. We can keep it if you're confident it isn't an issue, but I'd love to see if there is another way. If moving it up to the `AppController` is low effort, I'd be inclined to do that.

n8rzz:
> @erikquinn @n8rzz just FYI, it looks like this feature makes an existing problem a bit more exposed. The app performance drops significantly as the number of spawned aircraft increases. Historically when I use played around with the traffic generation sliders this was a very gradual issue that slowed the app down as more aircraft spawn. However, with this new button that resets traffic, a user can dial the arrivals up to 3x or 4x at major airports and experience instant significant performance issues. At higher multiples, the app crashes (at least on my machine in Chrome). This problem is exacerbated for airports that have arrivals that spawn very far out from the airspace, or where the 1x aircraft rate is already somewhat high.
> 
> I wasn't sure if you thought this might be a blocker for this PR.

Off. Browser crashing is not a good thing at all, though it doesn't surprise me. There is a ton of math that happens every frame, and every additional aircraft is more math in the same amount of time. 

Perhaps we add some copy at the top of the dialog that calls this out as a possible side effect if traffic is increased too much?  Another option could be to cap values at a certain multiple, is something like that possible?

CanisLupus518:
> Perhaps we add some copy at the top of the dialog that calls this out as a possible side effect if traffic is increased too much? Another option could be to cap values at a certain multiple, is something like that possible?

There's a few issues I think that can be addressed. First is, with the spawn rate high enough, the traffic generation can have just ridiculous intervals (like a mile or less). Second is the total number of aircraft spawned during the "prespawn" process. So I think there are 2 things we can do. 1) cap the acph in the traffic settings to a number that will not be more dense than 5 MIT per flow. 2) Come up with some hard aircraft limit for prespawning and only spawn that many aircraft, even if the prespawned aircraft don't fill in all the way to the airspace.  Thoughts?

CanisLupus518:
@n8rzz OK, have a look. While digging into this, I discovered that the math related to calculating how many aircraft to prespawn and the miles in trail was behaving weird at higher spawn rates. This issue could be seen in that the prespawned traffic was visibly closer together than aircraft that are spawned afterward. I have committed an attempt to correct the math to a better estimated MIT for prespawned aircraft (using TAS instead of IAS) as well as limiting the MIT for arrival patterns. Currently I hardcoded the limit to 10 MIT, in an effort to prevent turning the sliders up too high and crashing the app, now they will have no effect once the traffic has reached 10 MIT.  This value of 10 is up to you and @erikquinn though, it was just a guess on a reasonable hard limit. It does seem to prevent crashes on the few airports I have tried it on.

n8rzz:
I love the populate forward idea, I think that absolutely makes sense. I think I like option 1 better, but only because that feels like the simpler of the two possible solutions.

@erikquinn do you have any strong feelings here?

n8rzz:
I'll peek at those changes a little later tonight. I'm pumped!

CanisLupus518:
@n8rzz @erikquinn I think, feature-wise, this is pretty much done. I am left with one issue that I'm not sure where to look to fix. For the prespawned departure(s), after resetting the traffic and recreating the prespawned departure on the ground, it will not take off on its own under System control. After issuing it a take off command manually everything else then works fine. I am missing where this check is for issuing the takeoff for departures.

n8rzz (approved):

n8rzz:
> @n8rzz @erikquinn I think, feature-wise, this is pretty much done. I am left with one issue that I'm not sure where to look to fix. For the prespawned departure(s), after resetting the traffic and recreating the prespawned departure on the ground, it will not take off on its own under System control. After issuing it a take off command manually everything else then works fine. I am missing where this check is for issuing the takeoff for departures.

I'm not entirely sure, but my first guess would be here: https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/aircraft/AircraftModel.js#L170.  @erikquinn probably knows for sure, though, if I'm wrong.

The rest of this looks great!  Awesome work on all this, I really appreciate the time and effort you've put into this so far!

erikquinn:
>@erikquinn @n8rzz just FYI, it looks like this feature makes an existing problem a bit more exposed. The app performance drops significantly as the number of spawned aircraft increases.

Is it possible this is just the projected-path issue? I haven't seen any notable performance issues at 1x with projected-paths off, even with fairly absurd quantities of traffic.

>@n8rzz OK, have a look. While digging into this, I discovered that the math related to calculating how many aircraft to prespawn and the miles in trail was behaving weird at higher spawn rates. This issue could be seen in that the prespawned traffic was visibly closer together than aircraft that are spawned afterward. I have committed an attempt to correct the math to a better estimated MIT for prespawned aircraft (using TAS instead of IAS) as well as limiting the MIT for arrival patterns. Currently I hardcoded the limit to 10 MIT, in an effort to prevent turning the sliders up too high and crashing the app, now they will have no effect once the traffic has reached 10 MIT. This value of 10 is up to you and @erikquinn though, it was just a guess on a reasonable hard limit. It does seem to prevent crashes on the few airports I have tried it on.

Yeah, the prespawn math I've never been able to get figured out. The "random" method math for the rest of the spawning should be correct. And it already does have a limit of ~5 miles in trail, throwing a warning any time a spawn pattern attempts to exceed that value. Just mentioning it in case the stuff there is of use to you in this.

Hardcoded 10 MIT is a bit high for an absolute cap-- there are situations where you'd want lower values. Let's make the minimum 6 MIT, and feel free to throw a warning below 8 if you want.

>I love the populate forward idea, I think that absolutely makes sense. I think I like option 1 better, but only because that feels like the simpler of the two possible solutions.
>
>@erikquinn do you have any strong feelings here?

Definitely option 1 of capping the acph per spawn pattern. I do feel somewhat strongly that an arbitrary total cap would not be advisable. You could throw an error when exceeding a total cap, but the behavior should not change.


>>@n8rzz @erikquinn I think, feature-wise, this is pretty much done. I am left with one issue that I'm not sure where to look to fix. For the prespawned departure(s), after resetting the traffic and recreating the prespawned departure on the ground, it will not take off on its own under System control. After issuing it a take off command manually everything else then works fine. I am missing where this check is for issuing the takeoff for departures.
>
>I'm not entirely sure, but my first guess would be here: https://github.com/openscope/openscope/blob/develop/src/assets/scripts/client/aircraft/AircraftModel.js#L170. @erikquinn probably knows for sure, though, if I'm wrong.
>
>The rest of this looks great! Awesome work on all this, I really appreciate the time and effort you've put into this so far!

Nah, not there... it's here actually:

https://github.com/openscope/openscope/blob/ead9a7115be10e954baec6ea43ba0c8c1fdbe3dd/src/assets/scripts/client/aircraft/AircraftController.js#L588

In `AircraftCommander._createAircraftWithInitializationProps()`. So it depends where you are spawning the aircraft, if perhaps through a different method than this one? I would assume so, as otherwise the conditions should pass and the auto-tower stuff would just happen without you needing to specifically ask it to.

erikquinn:
Sorry for my late response! If it's not too high maintenance, @n8rzz could you confirm that my answers align with what you approved? As in, is this ready to go, or will anything need changes based on those answers. I still haven't even had time to review it myself. Sorry for slacking guys!

CanisLupus518:
@erikquinn  I will change the hard coded MIT cap to 6 and add a warning below 8 

As far as the `isAutoTower` thing goes, it is going through that code correctly as I am calling that same method. I've walked through it a bunch of times, but I can't find the code that actually triggers the takeoff command in order to track it down further.

n8rzz:
> Sorry for my late response! If it's not too high maintenance, @n8rzz could you confirm that my answers align with what you approved? As in, is this ready to go, or will anything need changes based on those answers. I still haven't even had time to review it myself. Sorry for slacking guys!

looks good to me

erikquinn:
> @erikquinn I will change the hard coded MIT cap to 6 and add a warning below 8
> 
> As far as the `isAutoTower` thing goes, it is going through that code correctly as I am calling that same method. I've walked through it a bunch of times, but I can't find the code that actually triggers the takeoff command in order to track it down further.

Okey doke. Just rerequest a review when it's ready. I should have time to sit down and peruse tomorrow, but I'm sure I won't have any complaints. Pretty pumped for this one! 🙂

CanisLupus518:
@erikquinn the tower takeoff issue was in the RunwayModel.  The traffic reset was removing the previously spawned departure, and the newly spawned departure was waiting for that no-longer-existing aircraft to get further away. I implemented a reset on the runway model to clear the queue and the previously departed aircraft and that seems to have done the trick.

src/assets/scripts/client/AppController.js
```diff
@@ -299,6 +302,21 @@ export default class AppController {
...
+     * event callback fired from within the `TrafficRateController` when traffic should be cleared and respawned.
+     *
+     * @for AppController
+     * @method onTrafficReset
+     */
+    onTrafficReset() {
+        EventTracker.recordEvent(TRACKABLE_EVENT.AIRPORTS, 'traffic-reset', AirportController.current.icao);
+        this.aircraftController.aircraft_remove_all();
+        this.scopeModel.radarTargetCollection.reset();
+        AirportController.current.departureRunwayModel.reset();
```
erikquinn, line 315: If this is being done to clear the departure queue, you should probably do it for all runways, since users can assign any runway they choose.

erikquinn (approved):
Only one thing I saw to check. Otherwise looks great to me. Haven't had the pleasure of using it yet though. On mobile at work still.

🎉

n8rzz (approved):
Provided you address Erik's feedback, this looks great to me.  Nice work!!

erikquinn (changes requested):
After pressing "restart", everything looks fine. But on INITIAL load of the sim, the prespawned aircraft are messed up somewhow. Pictured below is KSEA (default), but you can change to other airports and observe the same thing. If you then open the dialog and reset the traffic, the prespawned traffic is as expected. Investigation needed. 🤔 

image

erikquinn:
@CanisLupus518 I pushed a change to address the runway queue resets, and in testing functionality, everything checked out except for the initial load state of the prespawned aircraft, as described above. Looking into it now....

CanisLupus518:
> @CanisLupus518 I pushed a change to address the runway queue resets, and in testing functionality, everything checked out except for the initial load state of the prespawned aircraft, as described above. Looking into it now....

I'm looking into it as well, apologies i didn't notice it

CanisLupus518:
> @CanisLupus518 I pushed a change to address the runway queue resets, and in testing functionality, everything checked out except for the initial load state of the prespawned aircraft, as described above. Looking into it now....

@erikquinn it looks like I did not handle the case where `SpawnPattern` altitude is a min\max array. I will fix it

CanisLupus518:
@erikquinn I believe I have corrected the issue

erikquinn (approved):

erikquinn:
All better!

GREAT job on this one. Messed with it for a while. Mesmerizing stuff! :)

PR #1919, opened 2022-02-20
refactor/1916 - Update KORD

### LINKED ISSUE #1916: Revive ORD - Chicago O'Hare
Reviving ORD, Chicago O'Hare.

MichaelDMcCracken:
Resolves #1916 

The purpose of this pull request is to add ORD.


Fechulo (changes requested):
Ah this file brings back memories 😍 
Github won't let me add review comments on the file itself because it's too large so:
ERNNY7: Missing restrictions at KOHLL and CHDRR
MADII6: (phonetic) name is wrong
VEECK5: VEECK is entered twice, once in the BONNT transition and then again in the body
WYNDE2: UFDUH transition is named incorrectly. Also missing restrictions at TVC and UFDUH
Misc.: Could we add the extended centerline for 28L, given that it is used as an arrival runway by default?

MichaelDMcCracken:
>  Could we add the extended centerline for 28L, given that it is used as an arrival runway by default?

How do I do that?

MichaelDMcCracken:
- [x] ERNNY7: Missing restrictions at KOHLL and CHDRR
- [x] MADII6: (phonetic) name is wrong
- [x] VEECK5: VEECK is entered twice, once in the BONNT transition and then again in the body
- [X] WYNDE2: UFDUH transition is named incorrectly.
- [X] Also missing restrictions at TVC and UFDUH
- [ ] Could we add the extended centerline for 28L, given that it is used as an arrival runway by default?

MichaelDMcCracken:
@Fechulo @erikquinn 
What does the `sepFromAdjacent` property actually do? I tried setting this to `0` for all of the runways in hopes that it would allows simultaneous parallel approaches but it didn't work.

MichaelDMcCracken:
@Fechulo anything happening with this pull?

Fechulo (approved):

PR #1957, opened 2022-06-08
enhancement/1956 - Departure altitudes

### LINKED ISSUE #1956: Departure SpawnPattern Cruise Altitude - Direction of Flight
By Default, Departure patterns are assigned a cruise altitude equal to the operating ceiling of the individual aircraft. This number seems unrealistic, even if the value is mostly cosmetic for Terminal controlling. 

The SpawnPattern already has an "altitude" property, that could be used to specify an altitude or range of altitudes to use for generating the filed cruise altitude of departures. While this doesn't necessarily automate or correct "direction of flight", an airport designer can assign the proper altitudes for the specific pattern.

CanisLupus518:
Resolves #1956

The purpose of this pull request is to give the airport designer more control over the assigned cruise altitudes of departure patterns

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Bugfixes
 
 ### Enhancements & Refactors
+- <a href="https://github.com/openscope/openscope/issues/1956" target="_blank">#1956</a> - Allow airport creators to specify requested cruise altitudes for departures
 
 
 # 6.27.0 (March 18, 2021)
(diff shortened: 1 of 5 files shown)


n8rzz (approved):
Are there any tests that could be added to verify this as well?

CanisLupus518:
> Are there any tests that could be added to verify this as well?

Yes probably, I can add that as well as fix the broken tests

PR #1808, opened 2021-04-10
devops/1793 - Use eslint-plugin-diff in lieu of lint-diff

paleite:
Resolves #1793

The purpose of this pull request is to help in trying out eslint-plugin-diff

Hi! I'm the author of eslint-plugin-diff and I saw that you wanted to give it a shot, so I thought I'd prepare a PR for you, to be able to try it out.

One thing I must point out is that in order for eslint-plugin-diff to work, ESLint version >=6 is required, and I wasn't comfortable bumping a major version without your prior consent.

I hope this helps you somewhat, and please let me know if there's anything else I can help you with.

I appreciate you wanting to try it out and your feedback is very valuable in improving the plugin.

Thanks!

PR #1961, opened 2022-07-03
Look into eslint-plugin-diff vs lint-diff #1793

### LINKED ISSUE #1793: Look into eslint-plugin-diff vs lint-diff
I just bumped into [`eslint-plugin-diff`](https://www.npmjs.com/package/eslint-plugin-diff) which has just [added CI support](https://github.com/paleite/eslint-plugin-diff/releases/tag/v1.0.6) for linting only modified code.

At a cursory glance it may confer advantages compared to `lint-diff` which we currently use:
- is a plugin for eslint, formally integrated with it; rather than wrapping eslint and hacking around the edges
- smaller unpacked size, zero dependencies
- newer, actively maintained

Might be worth further examination / evaluation.
I don't have the bandwidth to do this right now so anyone else is welcome to pick it up if they want.

erikquinn:
Resolves #1793.
Replaces #1808.

Reopening to retrigger the failed checks to see if this can work or not.


erikquinn:
Seems like the checks aren't running? 😕 

I don't know. Will just mark as orphaned for now in case anyone else wants to come back and figure it out. The current system does work though, so 🤷‍♂️

Issue #1020, opened 2018-07-05
Remove score indicator 


Marcel510:
As already said on Slack, the score indicator should be removed  from the sim since it doesn't really suit a simulator.

odiferousmint:
I agree. The stats do not make much sense, especially not with these numbers. If it does not get removed (because some people do like stats), then perhaps there should be a setting to turn it on or off and the numbers should be adjusted as you get a pretty high penalty for most if not all mistakes, but not much reward for successful departures, landings and whatnot.

erikquinn:
Agreed. Elevating the priority so we can get rid of the silliness. :)

PR #1977, opened 2022-07-28
documentation/1972 - Updated scoring file to reflect current implementation

### LINKED ISSUE #1972: Undocumented loss of points when flying in restricted areas
Browser: Any

Airport: `KDCA`

Approximate play time before issue: Any

#### Issue Description
When an airctaft is flown into a restricted area, the player loses 200 points. This is not documented in [scoring.md](/openscope/openscope/blob/develop/documentation/scoring.md).

#### Steps to reproduce
- Fly any plane into a restricted area
- Observe score go down by 200

farid-fari:
Resolves #1972

The purpose of this pull request is to correct documentation on scoring to reflect current reality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
-# 6.29.0 (September 1, 2022)
+# 6.29.0 (October 1, 2022)
 ### New Features
 
 ### Bugfixes
 
 ### Enhancements & Refactors
+- <a href="https://github.com/openscope/openscope/issues/1977" target="_blank">#1977</a> - Update scoring documentation
 
 
 # 6.28.0 (July 3, 2022)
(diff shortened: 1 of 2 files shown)


n8rzz (approved):
Great work here and thank you for the PR! If the strikethrough bits are no longer valid, we can remove them.

farid-fari:
I wasn't sure if I should remove them or if they were there because they were goals that haven't been implemented yet... Let me know if I should just delete them.

erikquinn:
I suppose this is fine, but it is worth pointing out #1020 😛 

Will merge the update in a bit anyway, since we might as well after you did the work anyway :)

farid-fari:
I don't really care much about scores either, but it's useful to see at a glance if you're doing something majorly wrong without realizing (e.g. restricted areas). Also super useful for debugging maps when you're creating them ;)

PR #1987, opened 2022-08-25
devops/make-heroku-build-again

n8rzz:
Resolves #


erikquinn:
@n8rzz does this work now, or?

n8rzz:
It is for me, yes. could you check the review app quick and verify on your machine?

Once this is merged, we'll need to update the envs with the following:

We update the stack to `heroku-22`
Screen Shot 2022-10-03 at 10 33 56 PM

We update the buildpacks to `heroku/nodejs`
Screen Shot 2022-10-03 at 10 34 06 PM

Only one buildpack needed now 🎉

erikquinn:
going to close and reopen to see if that will trigger render to build a pr preview app.

erikquinn:
With render, none of this seems to be needed. Will just close and delete I guess!

n8rzz:
Only thing we may want from this is the documentation update. But, again with `Render`, that whole doc needs to change anyway.

PR #2010, opened 2022-10-31
Enhancement/2009 - Fix departure conflicts at KJFK

### LINKED ISSUE #2009: KJFK Conflicting departure SIDS
In KJFK, the departures currently use a mix of JFK5 and SKORR4 off of 31L. During a push at the default traffic rate, this can cause unrealistic as they launch too close together because, while similar, the left turn off the runway is earlier in one than the other. Present day operations tend to favor the SKORR4 over the JFK5 for 31L, so I recommend moving all departures to that SID to make things simpler.

CanisLupus518:
Resolves #2009 

The purpose of this pull request is to reduce departure conflicts by assigning departures a uniform SID. All JFK5 departures are now using the SKORR4


Fechulo (approved):
@erikquinn Looks like an extra file got through

PR #2023, opened 2022-12-06
Bugfix/2022 - "fly present heading" while still on ground

### LINKED ISSUE #2022: Behavior of "fly present heading" command for aircraft on the ground
`fph` command can be issued to aircraft that are not yet airborne, sometimes with outcomes that don't make sense.

#### APRON and TAXI

Aircraft heading is initialized by `AircraftModel.moveToRunway` at the end of taxi when the aircraft arrives at the runway and joins the queue of waiting departures. Thus, for flight phases `APRON` and `TAXI` the result of `fph` is nonsense -- the heading is either the default value of roughly due north, or in the case of an aircraft that had previously already taxied to one runway before being told to go to a different one, it is the runway heading of the previous runway that we're no longer going to use.

`fph` should not be allowed in these phases.

#### WAITING and TAKEOFF

IRL, our expectation is that `WAITING` aircraft form a queue off the side of the runway and only actually face runway heading after being issued "line up" clearance and entering the runway proper. However, the sim internally sets aircraft to face runway heading immediately at the end of taxi. For aircraft that have finished taxiing and are `WAITING`, the effect of `fph` is effectively "fly runway heading". This is actually a valid and useful outcome; we can take advantage of this by noting it in documentation and changing the readback accordingly for clarity.

Question for the ATC folks: what to do with aircraft in `TAKEOFF`? These are also still on the ground, or have yet to reach minimum altitude, and are thus also guaranteed to be at runway heading -- turning is not allowed until in `CLIMB`. In these cases, do you use "fly present heading" or "fly runway heading" phraseology? One consideration from the sim UX perspective is that the user can chain commands, to the effect of "fly runway heading, clear for takeoff" in which case I think we'd want `N12345 fph to` and `N12345 to fph` to have the same(-ish) readbacks.

cake-pie:
Resolves #2022

- prevent nonsense outcomes to `fph` command for departing aircraft still on the ground
- treat `fph` as "fly runway heading" in appropriate context

#### todo list

- [x] reject `fph` when in phases `APRON` and `TAXI`
- [x] change readback to "fly runway heading" for `WAITING`
- [x] consider whether to change readback for `TAKEOFF`
- [x] update documentation
- [x] update tests
- [x] changelog

### DIFF
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -228,8 +228,8 @@ export default class Pilot {
 
         const readback = {};
         const runwayOrPresent = (
-                aircraftModel.flightPhase === FLIGHT_PHASE.WAITING || aircraftModel.flightPhase === FLIGHT_PHASE.TAKEOFF
-            ) ? 'runway' : 'present';
+            aircraftModel.flightPhase === FLIGHT_PHASE.WAITING || aircraftModel.flightPhase === FLIGHT_PHASE.TAKEOFF
+        ) ? 'runway' : 'present';
         readback.log = `fly ${runwayOrPresent} heading`;
         readback.say = `fly ${runwayOrPresent} heading`;
 
--- a/src/assets/scripts/client/aircraft/Pilot/Pilot.js
+++ b/src/assets/scripts/client/aircraft/Pilot/Pilot.js
@@ -227,8 +227,9 @@ export default class Pilot {
         this._mcp.setHeadingHold();
 
         const readback = {};
-        readback.log = 'fly present heading';
-        readback.say = 'fly present heading';
+        const runwayOrPresent = aircraftModel.flightPhase === FLIGHT_PHASE.WAITING ? 'runway' : 'present';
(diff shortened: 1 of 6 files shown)


src/assets/scripts/client/aircraft/Pilot/Pilot.js
```diff
@@ -227,8 +227,13 @@ export default class Pilot {
         this._mcp.setHeadingHold();
 
         const readback = {};
-        readback.log = 'fly present heading';
-        readback.say = 'fly present heading';
+        if (aircraftModel.flightPhase === FLIGHT_PHASE.WAITING) {
```
n8rzz: you could get rid of the `if`/`else` and simplify this down to:

```js
const runwayOrPresent = aircraftModel.flightPhase === FLIGHT_PHASE.WAITING ? 'runway' : 'present';
const readback = `fly ${runwayOrPresent} heading`

return [true, readback];
```

n8rzz (approved):

src/assets/scripts/client/aircraft/Pilot/Pilot.js
cake-pie: done

PR #2031, opened 2022-12-12
Enhancement/2030 - add support for system command aliases, implement 'tw' alias for 'timewarp'

### LINKED ISSUE #2030: Add support for system command aliases, implement 'tw' alias for 'timewarp'
`tw` as an alias for `timewarp` doesn't actually work because the pipeline for handling system commands doesn't check for aliases.

https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/commands/aircraftCommand/aircraftCommandMap.js#L212-L216

cake-pie:
Resolves #2030

The purpose of this pull request is to add support for system command aliases so that the alias `tw` for `timewarp` actually works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - <a href="https://github.com/openscope/openscope/issues/1991" target="_blank">#1991</a> - Update KPHL magnetic variation
 - <a href="https://github.com/openscope/openscope/issues/1787" target="_blank">#1787</a> - Standardize JSON schemas
 - <a href="https://github.com/openscope/openscope/issues/1912" target="_blank">#1912</a> - Update tutorial for auto-departure feature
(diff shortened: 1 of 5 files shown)


n8rzz (approved):
Only thing missing here are some tests, but I don't remember if we even have tests for the system commands. If we do, are there any we could add for this? If we don't, nothing to worry about for now.

The work you've done here looks wonderful, thank you for your contribution!

cake-pie:
Looked at the existing tests related to system commands, found two relevant ones. In those, only one out of the four system commands in documentation was being fed as test case, so I added the other three. Added new tests for system command aliases to ensure they produce the same outcome as the root command.

n8rzz (approved):
Hell yeah! Nice work!! :shipit:

erikquinn:
@cake-pie oh thank god for this one. So overdue 😄 

Thank you! :)

PR #2032, opened 2022-12-13
bugfix/1249 - command parser: discard empty tokens

### LINKED ISSUE #1249: Timewarp with 2 spaces breaks everything
Browser: Chrome

Airport: Any

#### Steps to reproduce
- Run `timewarp  5` (2 spaces between the command and the value)
- Observe everything break

cake-pie:
Resolves #1249

The purpose of this pull request is to make the command parser robust against typos of consecutive spaces. Previously, this would result in empty tokens when the command string is split using space as a delimiter. Some commands broke when empty strings were ingested in place of expected arguments. Such empty tokens are now discarded.

### DIFF
--- a/src/assets/scripts/client/commands/parsers/CommandParser.js
+++ b/src/assets/scripts/client/commands/parsers/CommandParser.js
@@ -128,7 +128,7 @@ export default class CommandParser {
      */
     _extractCommandsAndArgs(rawCommandWithArgs) {
         const commandOrCallsignIndex = 0;
-        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR);
+        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR).filter((t) => t);
         const callsignOrSystemCommandName = commandArgSegmentsWithCallsign[commandOrCallsignIndex];
         // effectively a slice of the array that returns everything but the first item
         const commandArgSegments = _tail(commandArgSegmentsWithCallsign);
--- a/src/assets/scripts/client/commands/parsers/CommandParser.js
+++ b/src/assets/scripts/client/commands/parsers/CommandParser.js
@@ -206,11 +206,6 @@ export default class CommandParser {
 
         for (let i = 0; i < commandArgSegments.length; i++) {
             const commandOrArg = commandArgSegments[i];
-
-            if (commandOrArg === '') {
-                continue;
-            }
-
             const commandName = findCommandNameWithAlias(commandOrArg);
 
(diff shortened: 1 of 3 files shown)


src/assets/scripts/client/commands/parsers/CommandParser.js
```diff
@@ -128,7 +128,7 @@ export default class CommandParser {
      */
     _extractCommandsAndArgs(rawCommandWithArgs) {
         const commandOrCallsignIndex = 0;
-        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR);
+        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR).filter(t => t);
```
n8rzz: What exactly are we `.filter(t => t)` here? 

Oh, I think I see it. An empty string will evaluate as falsy here and will not be returned. Is that correct? If so, that's clever.

src/assets/scripts/client/commands/parsers/CommandParser.js
```diff
@@ -128,7 +128,7 @@ export default class CommandParser {
      */
     _extractCommandsAndArgs(rawCommandWithArgs) {
         const commandOrCallsignIndex = 0;
-        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR);
+        const commandArgSegmentsWithCallsign = rawCommandWithArgs.split(COMMAND_ARGS_SEPARATOR).filter((t) => t);
```
n8rzz, line 131: Is this coercing an empty string (space) to falsey and filtering those out? I've never seen this before. If that's it, this is clever.

n8rzz (approved):
You're probably digging into this already, but I'd love to see some tests for this too. Especially since this is a bugfix. Otherwise looks good, clean and simple. Thank you for your contribution, we really appreciate it!

src/assets/scripts/client/commands/parsers/CommandParser.js
cake-pie: Yes, that's pretty much it.

cake-pie:
Added test, cleaned up commits, all set.

test/commands/parsers/CommandParser.spec.js
```diff
@@ -76,6 +77,17 @@ ava('sets #commandList with AircraftCommandModel objects when it receives transm
     });
 });
 
+ava('._extractCommandsAndArgs() discards empty tokens caused by multiple spaces', t => {
```
n8rzz, line 99: Love it!

n8rzz (commented):
Thanks for adding that test!

PR #2033, opened 2022-12-13
enhancement/2020 - Add chat log duration setting

### LINKED ISSUE #2020: Command to have aircraft repeat last transmission
Browser: All

Airport: All

Approximate play time before issue: Almost immediately

#### Enhancement Description
Due to how quickly the communication log fades out, you sometimes miss what an aircraft called you with. One solution to this is to create a new command asking the aircraft to repeat themselves.

I know that in the real world, the phraseology is usually "say again" but there is already a command for `sa` so I propose using `repeat` or `rpt`

#### Proposed Enhancement
`AAL123 repeat` or `AAL123 rpt`

rcobelli:
Resolves #2020

The purpose of this change is to allow users to adjust how long a chat message will remain visible before fading out. Currently messages are shown for 3 seconds (the new default value for the setting) but users can now change the delay to any valid number greater than 0

_Note: The link to join slack doesn't work so I can't request access to the repo. Hence why this PR is from a fork_

### DIFF
--- a/src/assets/scripts/client/constants/eventNames.js
+++ b/src/assets/scripts/client/constants/eventNames.js
@@ -28,10 +28,10 @@ export const EVENT = {
 
     /**
      * @memberof EVENT
-     * @property WIND_CHANGE
+     * @property CHAT_LOG_DURATION_CHANGE
      * @type {string}
      */
-    WIND_CHANGE: 'wind-change',
+    CHAT_LOG_DURATION_CHANGE: 'chat-log-duration-change',
 
     /**
      * A click was registered outside of a specific `StripViewModel`
@@ -55,6 +55,13 @@ export const EVENT = {
      * @type {string}
      */
     MARK_SHALLOW_RENDER: 'mark-shallow-render',
+
+    /**
+     * @memberof EVENT
+     * @property MEASURE_TOOL_STYLE_CHANGE
+     * @type {string}
+     */
+    MEASURE_TOOL_STYLE_CHANGE: 'measure-tool-style-change',
 
     /**
      * A pan event has been detected necessitating an entire redraw of each canvas
@@ -287,13 +294,6 @@ export const EVENT = {
     RANGE_RINGS_CHANGE: 'range-rings-change',
 
     /**
-     * @memberof EVENT
-     * @property MEASURE_TOOL_STYLE_CHANGE
-     * @type {string}
-     */
-    MEASURE_TOOL_STYLE_CHANGE: 'measure-tool-style-change',
-
-    /**
      * A click has been registered in the unpause button shown within the
      * screen overlay whil the app is paused
      *
@@ -304,20 +304,20 @@ export const EVENT = {
     UNPAUSE: 'unpause',
 
     /**
+     * @memberof EVENT
+     * @property WIND_CHANGE
+     * @type {string}
+     */
+    WIND_CHANGE: 'wind-change',
+
+    /**
      * The zoom level has changed necessitating an entire redraw of each canvas
      *
      * @memberof EVENT
      * @property ZOOM_VIEWPORT
      * @type {string}
      */
-    ZOOM_VIEWPORT: 'zoom-viewport',
-
-    /**
-     * @memberof EVENT
-     * @property CHAT_LOG_DURATION_CHANGE
-     * @type {string}
-     */
-    CHAT_LOG_DURATION_CHANGE: 'chat-log-duration-change'
+    ZOOM_VIEWPORT: 'zoom-viewport'
 };
 
 export const AIRCRAFT_EVENT = {
--- a/src/assets/scripts/client/constants/eventNames.js
+++ b/src/assets/scripts/client/constants/eventNames.js
@@ -310,7 +310,14 @@ export const EVENT = {
      * @property ZOOM_VIEWPORT
      * @type {string}
      */
-    ZOOM_VIEWPORT: 'zoom-viewport'
+    ZOOM_VIEWPORT: 'zoom-viewport',
+
+    /**
+     * @memberof EVENT
+     * @property CHAT_LOG_DURATION_CHANGE
+     * @type {string}
+     */
+    CHAT_LOG_DURATION_CHANGE: 'chat-log-duration-change'
 };
 
 export const AIRCRAFT_EVENT = {
--- a/src/assets/scripts/client/constants/gameOptionConstants.js
+++ b/src/assets/scripts/client/constants/gameOptionConstants.js
@@ -270,10 +270,8 @@ export const GAME_OPTION_VALUES = [
         unit: 'sec',
         validationHandler(newVal) {
             const output = parseFloat(newVal);
-            if (Number.isNaN(output) || output <= 0) {
-                return false;
-            }
-            return true;
+
+            return !Number.isNaN(output) && output > 0;
         }
     }
 ];
--- a/src/assets/scripts/client/constants/gameOptionConstants.js
+++ b/src/assets/scripts/client/constants/gameOptionConstants.js
@@ -11,6 +11,7 @@ import { MEASURE_TOOL_STYLE } from './inputConstants';
  */
 export const GAME_OPTION_NAMES = {
     CONTROL_METHOD: 'controlMethod',
+    CHAT_LOG_DURATION: 'chatLogDuration',
     DRAW_PROJECTED_PATHS: 'drawProjectedPaths',
     DRAW_ILS_DISTANCE_SEPARATOR: 'drawIlsDistanceSeparator',
     MEASURE_TOOL_PATH: 'measureToolPath',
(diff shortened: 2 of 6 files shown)


src/assets/scripts/client/constants/eventNames.js
```diff
@@ -310,7 +310,14 @@ export const EVENT = {
...
      */
-    ZOOM_VIEWPORT: 'zoom-viewport'
+    ZOOM_VIEWPORT: 'zoom-viewport',
+
+    /**
+     * @memberof EVENT
+     * @property CHAT_LOG_DURATION_CHANGE
+     * @type {string}
+     */
+    CHAT_LOG_DURATION_CHANGE: 'chat-log-duration-change'
```
n8rzz: could you please alphabetize this list?

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -98,13 +98,13 @@ export default class SettingsController {
         const descriptions = GameController.game.option.getDescriptions();
 
         _forEach(descriptions, (opt) => {
-            if (opt.type !== 'select') {
-                return;
+            if (opt.type === 'select') {
```
n8rzz, line 103: try this instead: 
```js
const $container = this._buildTextInputTemplate(opt);

if (opt.type === 'select') {
    const $container = this._buildSelectInputTemplate(opt);
}
    
this.$dialogBody.append($container);
```

src/assets/scripts/client/ui/UiController.js
```diff
@@ -156,6 +156,16 @@ class UiController {
...
 
+        /**
+         * Duration for chat log transmissions to remain visible
+         *
+         * @for UiController
+         * @property $chatLogDuration
+         * @type number
+         * @default null
+         */
+        this.$chatLogDuration = null;
```
n8rzz: you can remove the `$` since this is a numbr. The `$` is reserved only for jquery elements.

src/assets/scripts/client/ui/UiController.js
```diff
@@ -156,6 +156,16 @@ class UiController {
          */
         this.$log = null;
 
+        /**
+         * Duration for chat log transmissions to remain visible
```
n8rzz: What is the time value: Miliseconds or Seconds? We should add that in the doc block here for future us 😉

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+     * @for SettingsController
+     * @method _buildTextInputTemplate
+     * @param option {object}
+     * @return $container {jquery Element}
+     */
+    _buildTextInputTemplate(option) {
+        const $container = $(UI_OPTION_CONTAINER_TEMPLATE);
+        const $label = $(UI_OPTION_LABEL_TEMPLATE);
+        const currentValue = GameController.game.option.getOptionByName(option.name);
+        const $selector = $(`<input class="form-input" step="any"
```
n8rzz, line 206: For larger bits of HTML, I _think_ we put them in string templates at the top of the file. Check me on that, cause it's been a while since I've been in the code here. I'm ok with it here, but I'd like to keep things consistent if we can.

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+                                    type="${option.type}"
+                                    name="${option.name}"
+                                    value="${currentValue}">`);
+        const $unit = $(`<span class="form-value">${option.unit}</span>`);
+
+        $container.append($label);
+        $label.text(option.description);
```
n8rzz, line 212: can we reverse these? It just reads a little better this way.
```js
$label.text(option.description);
$container.append($label);
```

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+        $selector.change((event) => {
+            const $currentTarget = $(event.currentTarget);
+
+            if (!option.validationHandler($currentTarget.val())) {
+                // User didn't enter a valid value, revert to the old value
+                $selector.val(currentValue);
+                return;
+            }
+
+            let value = $currentTarget.val();
```
n8rzz: let's call this `nextValue` so it lines up with `currentValue`.

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+                                    value="${currentValue}">`);
+        const $unit = $(`<span class="form-value">${option.unit}</span>`);
+
+        $container.append($label);
+        $label.text(option.description);
+
+        // TODO: this should be moved to proper event handler method and only assigned here.
+        $selector.change((event) => {
+            const $currentTarget = $(event.currentTarget);
+
```
n8rzz, line 219: I'd encourage you to add another variable here fro `currentValue`:
```js
const currentValue = $currentTarget.val();
```

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+            if (!option.validationHandler($currentTarget.val())) {
```
n8rzz: with the `currentValue` variable, we can replace these `$currentTarget.val()` calls

src/assets/scripts/client/ui/SettingsController.js
```diff
@@ -177,20 +177,65 @@ export default class SettingsController {
...
+                // User didn't enter a valid value, revert to the old value
+                $selector.val(currentValue);
+                return;
```
n8rzz, line 224: could you please add an empty line above `return`?

src/assets/scripts/client/ui/UiController.js
```diff
@@ -370,6 +380,8 @@ class UiController {
         this.$toggleVideoMap = this.$element.find(SELECTORS.DOM_SELECTORS.TOGGLE_VIDEO_MAP);
         this.$tutorialDialog = this.$element.find(SELECTORS.DOM_SELECTORS.TUTORIAL);
```
n8rzz: please remove empty line

src/assets/scripts/client/ui/UiController.js
```diff
@@ -381,6 +393,7 @@ class UiController {
      */
     setupHandlers() {
         this.onAirportChangeHandler = this.onAirportChange.bind(this);
+        this.onChatLogDurationChangeHandler = this.onChatLogDurationChange.bind(this);
```
n8rzz, line 395: Well done!! This trips folks up sometimes cause it's not something folks see a lot these days.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -258,5 +259,21 @@ export const GAME_OPTION_VALUES = [
...
+    },
+    {
+        name: GAME_OPTION_NAMES.CHAT_LOG_DURATION,
+        defaultValue: '3',
+        description: 'Chat Log Duration',
+        help: 'How long chat transmissions will remain visible before fading out',
+        type: 'number',
+        onChangeEventHandler: EVENT.CHAT_LOG_DURATION_CHANGE,
+        unit: 'sec',
+        validationHandler(newVal) {
```
n8rzz, line 271: is `validationHandler` a new concept? I like it! There are some rooms for abstraction here, but I think it's probably too soon so we can leave it like this.

src/assets/scripts/client/constants/gameOptionConstants.js
```diff
@@ -258,5 +259,21 @@ export const GAME_OPTION_VALUES = [
...
+            const output = parseFloat(newVal);
```
n8rzz, line 272: how about this?
```js
const output = parseFloat(newVal);

return !Number.isNaN(output) && output > 0;
```

n8rzz (commented):
Looking pretty good! I left a few bits of feedback, though most are pretty minor/style tweaks. Please reach out and let me know if you have any questions!

src/assets/scripts/client/ui/SettingsController.js
rcobelli: For larger blocks (multiple HTML elements), you're correct. However, this is just a single element with several attributes so it looks longer than it really is.

src/assets/scripts/client/constants/gameOptionConstants.js
rcobelli: Yes, new concept

src/assets/scripts/client/ui/SettingsController.js
n8rzz: Works for me, thanks for checking!

n8rzz (approved):
Well done, thank you for your contributions! We really appreciate it.

Issue #1725, opened 2021-01-19
JSONize airport load list


cake-pie:
#### Issue Description
`airportLoadList.js` is a javascript source file that only serves to contain data.

The justification appears to be:

https://github.com/openscope/openscope/blob/67f23ecfa27e73eb698ee7fe7c008ac03fdb8d3e/assets/airports/airportLoadList.js#L2-L9

Same can be accomplished by converting all of the airport load list data into JSON format, and ingesting it at app instantiation.  
Once built, the live copy of the file at `public/assets/airports/airportLoadList.json` can be modified without re-build / re-bundle  (skip minifying this file to preserve human readability)

Other benefits:
- all javascript source files compiled into `bundle.min.js` without any need for exceptions
- the json file could be validated against a json schema as part of workflow
- airport work issues and PRs no longer modify this one javascript file in order to add/remove airport from the list
  - such issues/PRs stop including the `refactor` label, which should really only apply to actual code changes

---

Because JSON does not support comments, before I proceed with this I need to know what should be done with all the entries in the list (44 of them!) that have been commented out. I can either:
- give them a `"disabled": true` property which is basically the same thing as `wip` that was removed for #1001
- delete them outright

erikquinn:
`disabled: true` sounds logical to me!

cake-pie:
Went with `disabled` so that omission of the optional property is equivalent to false by default, which is nicer semantically.

PR #2035, opened 2022-12-14
enhancement/1821 - add aliases cm, dm, tl, tr

### LINKED ISSUE #1821: Add alias commands to support native vatsim emulator commands
Add `cm`, `dm` aliases for descend/climb, `tr`, `tl` for `t l`/`t r`.

These commands are common in VATSIM trainer software, such as TWRTrainer and ESKey for EuroScope.

cake-pie:
Resolves #1821.
Blocks #2041.

The purpose of this pull request is to add `cm`, `dm` aliases for descend/climb, `tr`, `tl` for `t l`/`t r`.

`cm`, `dm` are straightforward additions, but `tl` and `tr` required hacky special case handling because the alias system only supports an alias serving as shorthand or substitute for a command, and not command plus some argument(s). Likewise in the autocomplete config, they have to be treated as separate commands with their own entries rather than as a variation of the existing `heading`.

It's not exactly pretty, so I'll let you guys decide if you want to tolerate the hacky solution or if you'd rather forgo having `tl` and `tr`.

- [x] implement aliases
- [x] update tests
- [x] update autocomplete config
- [x] update documentation
- [x] changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1977" target="_blank">#1977</a> - Update scoring documentation
 - <a href="https://github.com/openscope/openscope/issues/1991" target="_blank">#1991</a> - Update KPHL magnetic variation
+- <a href="https://github.com/openscope/openscope/issues/1821" target="_blank">#1821</a> - Add aliases cm, dm for altitude, tl, tr for turn
 
 
 # 6.28.0 (July 3, 2022)
--- a/assets/autocomplete/commandAutocompleteConfig.json
+++ b/assets/autocomplete/commandAutocompleteConfig.json
@@ -553,6 +553,64 @@
       ]
     },
     {
+      "id": "tl",
+      "variants": [
+        {
+          "aliases": [
+            "tl"
+          ],
+          "altkeys": [
+            "turn left",
+            "left turn"
+          ],
+          "explain": "<b>t</b>urn left to <u>heading</u>, or by given <u>number of degrees</u>"
+        }
+      ],
+      "paramsets": [
+        {
+          "example": "060",
+          "explain": "turn left to fly <u>heading</u>",
+          "candidate": "^(?:|[0-3]|[0-2][0-9]|3[0-6]|00[1-9]|0[1-9][0-9]|[1-2][0-9]{2}|3[0-5][0-9]|360)$",
+          "validate": "^(00[1-9]|0[1-9][0-9]|[1-2][0-9]{2}|3[0-5][0-9]|360)$"
+        },
+        {
+          "example": "30",
+          "explain": "turn left by <u>number of degrees</u>",
+          "candidate": "^(?:[1-9][0-9]?)?$",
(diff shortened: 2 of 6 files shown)


n8rzz (commented):
Does `commandAutocompleteConfig` need to be a json file or could we translate that to js? The reason I ask, is then we can set those regexes to a names variable and then share them between commands.

cake-pie:
That's just how I happened to implement it at the time. Can't recall the exact design decisions that went into it, but off the top of my head, it has the advantage of living at `public/assets/autocomplete` so in theory people could add their own search keys for commands if they wished without having to rebuild. Don't know what happened to the parser refactoring effort, but a further possibility is that if the parser is also based on this (or something evolved from it), then we would have a path for building a means for users to bind their own aliases for commands.

It's most likely possible to add a bit more machinery to the autocomplete so that even with a json file, regexes used in more than one place can be defined once (probably inside a `commonRegexes` object or somesuch) and reused by multiple commands (by referencing the key). Regardless, that's out of scope for this PR.

n8rzz:
> That's just how I happened to implement it at the time. Can't recall the exact design decisions that went into it, but off the top of my head, it has the advantage of living at `public/assets/autocomplete` so in theory people could add their own search keys for commands if they wished without having to rebuild. Don't know what happened to the parser refactoring effort, but a further possibility is that if the parser is also based on this (or something evolved from it), then we would have a path for building a means for users to bind their own aliases for commands.
> 
> It's most likely possible to add a bit more machinery to the autocomplete so that even with a json file, regexes used in more than one place can be defined once (probably inside a `commonRegexes` object or somesuch) and reused by multiple commands (by referencing the key). Regardless, that's out of scope for this PR.

Those are good reasons! I'd be curious, though, if that's a valid use case. Do users have a need to add their own aliases on the fly?

cake-pie:
Well, `tw` was me scratching my own itch, because I use the command pretty heavily and found it unwieldy to have to type `timewarp` all the time.

Even though the code changes needed to add a simple alias are actually trivial for those who know how -- at bare minimum you only have to add an entry to the array of aliases for the command in `aircraftCommandMap.js` -- it requires knowledge of coding and entails rebuilding from source. I think being able to support a QoL feature similar to key rebinding in games is a nice aspirational goal.

n8rzz:
> Well, `tw` was me scratching my own itch, because I use the command pretty heavily and found it unwieldy to have to type `timewarp` all the time.
> 
> Even though the code changes needed to add a simple alias are actually trivial for those who know how -- at bare minimum you only have to add an entry to the array of aliases for the command in `aircraftCommandMap.js` -- it requires knowledge of coding and entails rebuilding from source. I think being able to support a QoL feature similar to key rebinding in games is a nice aspirational goal.

I completely agree with that and I love that you're thinking about those sorts of things. I'm curious again, are there other areas of the sim where we might be able to do more with a strategy such as this?

cake-pie:
> are there other areas of the sim where we might be able to do more

It's not like I've reviewed the entire codebase for stuff like this, but I've been flagging things up or refactoring stuff myself as I encounter them, such as the push to JSONize things like #1725.

PR #2039, opened 2022-12-15
bugfix/1852 - Fix airborne aircraft calling up with "ready to taxi"

### LINKED ISSUE #1852: Departures who leave and reenter airspace will say they're "ready to taxi"
Originally reported in #1844.

Airport: KPHL

#### Issue Description
Departures who are issued a climb to an altitude owned in their current airspace polygon, who pass into a different airspace polygon with a _lower_ ceiling, will still climb to the assigned altitude, resulting in them leaving the airspace (this is expected). If they continue along their route and enter a new polygon where they are again within approach control altitudes, they check back on the frequency (expected), with a very weird message of being "ready to taxi". Obviously, they should not be requesting taxi, but just checking back on the frequency with their altitude (including climb/descent, as appropriate).

#### Steps to reproduce
1. Load KPHL
2. Turn on Airspace Depiction.
2. Full-reroute any departure `WOJIK..TORNE..PTW..SARAA` (use `rr`).
3. Have aircraft level at 10,000 prior to WOJIK.
4. Over WOJIK, issue climb to 12,000.
5. Observe receipt of clearance and start of climb to 12,000.
6. Observe as aircraft leaves the airspace (vertically) just prior to TORNE, and is uncontrollable.
7. When aircraft leaves polygon 3 and enters polygon 2, they are again within our altitudes. At this time, the aircraft will become controllable and contact us to request taxi.

cake-pie:
Resolves #1852

The purpose of this pull request is to change `AircraftModel.callUp()` so that the call up message that is used is determined by whether acft is airborne or not, rather than whether it is an arrival or departure acft.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - <a href="https://github.com/openscope/openscope/issues/2026" target="_blank">#2026</a> - Improve error messages from invalid takeoff commands
 - <a href="https://github.com/openscope/openscope/issues/1249" target="_blank">#1249</a> - Fix handling of consecutive spaces in commands
 - <a href="https://github.com/openscope/openscope/issues/1795" target="_blank">#1795</a> - Fix strip bay alignment bug when opening by cliking aircraft
+- <a href="https://github.com/openscope/openscope/issues/1852" target="_blank">#1852</a> - Fix bug where airborne aircraft report "ready to taxi"
 
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1977" target="_blank">#1977</a> - Update scoring documentation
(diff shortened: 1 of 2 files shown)


n8rzz (approved):
As usual, if there is room to add a test or two to cover this i'd love to see it. I don't remember the state of tests for the `AircraftModel`, so if it's a pain in the ass don't worry about it.

cake-pie:
Not covered by any pre-existing tests, and having to set up mocks for cases where the aircraft is inside or outside airspace is more complex than I'd be comfortable tackling.

n8rzz:
> Not covered by any pre-existing tests, and having to set up mocks for cases where the aircraft is inside or outside airspace is more complex than I'd be comfortable tackling.

Totally fair, don't worry about it then. Thanks for checking!

PR #2041, opened 2022-12-15
Bugfix/2034 - command parser now chokes on unrecognized tokens

### LINKED ISSUE #2034: Unrecognized tokens silently ignored by command parser
The command parser silently ignores any tokens that it does not recognize as long as they occur between the aircraft callsign and the first valid command.

This gets rejected
rejects

But this silently discards the unrecognized tokens (equivalent to just `SWA577 dvs`)
acceptsnonsense

This can become a problem when the user makes a typo in some command. For instance, for separation reasons I might issue vectors to an aircraft that diverts it temporarily from its flight plan, and then later want to have it resume navigating along the filed waypoints. Suppose I intend to issue `JBU100 pd ssoxs sp 250` but typo as `JBU100 pf ssoxs sp 250` -- the aircraft will happily accept the 'speed' command without making any mention of the botched 'proceed direct'. If this mistake goes unnoticed, the aircraft could eventually trigger `leaving airspace without being on our route`.

By comparison `JBU100 sp 250 pf ssoxs` would actually produce `Command no understood` and retain the text in the input, making it convenient to detect and correct the mistake and reissue the command. This should be the desired behavior when any unrecognized input is present.

cake-pie:
Resolves #2034.
Depends on #2035. (rebase this once that is merged)

The purpose of this pull request is to prevent the command parser from silently ignoring any tokens that it does not recognize.

This is expected to cause a merge conflict with #2035 -- if you want, I can rebase this after that gets merged first. In the meantime, this can still be deployed and tested.

Todo
- [ ] rebase after #2035 is merged first
- [x] changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ### New Features
 
 ### Bugfixes
+- <a href="https://github.com/openscope/openscope/issues/2034" target="_blank">#2034</a> - Fix parser accepting and silently ignoring unrecognized tokens
 
 ### Enhancements & Refactors
 - <a href="https://github.com/openscope/openscope/issues/1977" target="_blank">#1977</a> - Update scoring documentation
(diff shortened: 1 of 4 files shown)


n8rzz (approved):
Just need a changelog entry and this is good to go!

PR #2044, opened 2022-12-16
bugfix/2040 - Fix spurious "no such aircraft, say again" error message

### LINKED ISSUE #2040: Spurious "no such aircraft, say again"
#### Issue Description

Transmitting aircraft callsign only (as populated automatically) without adding commands leads to spurious error readback of "no such aircraft, say again".

#### Steps to reproduce
- load an airport
- click an aircraft radar return or a strip in the strip view to populate the command input
- press enter to submit input as-is without any commands
- compare versus manual entry of the aircraft callsign with no commands but any number of trailing spaces, which will correctly give something like "UAL100, say again"

cake-pie:
Resolves #2040

The purpose of this pull request is to fix spurious error message when the command input is automatically populated with aircraft callsign by clicking an aircraft radar return or flight strip, and then transmitted as-is without any modification.


n8rzz (approved):
I love PRs like these where we can remove more code than we add. Especially when some of that code was very wrong. Thank you!

PR #2046, opened 2022-12-21
Refactor/2045 - assorted refactoring and optimization

### LINKED ISSUE #2045: Assorted refactoring and optimization
Issue number for assorted refactoring and optimization

cake-pie:
Resolves #2045

The purpose of this pull request is to perform assorted refactoring:

f5cc33e: Optimize `AirspaceModel.isPointInside()` so that cheaper tests (altitude) are used for early return before performing costlier point-in-polygon checks

b685d9c: Remove redundant test in `AircraftModel` constructor, replaced by equivalent `else`

0741e55: Potential early return in `AircraftModel.isOnGround()` before any variable assignments

a5df155:

- Address [this todo](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L1114). `AircraftModel.isTakeoff()` implemented behavior changed to match function name. This does not lead to any behavior change, because this function is only called [here](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L2489) in `increaseAircraftAltitude()` which is only [called by](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L2447) `updateAltitudePhysics()` which is only [called by](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L2359) `updatePhysics()` which has an early return at its beginning if `this.isTaxiing()`.

- Remove [this todo](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L2488) because it appears to be satisfactorily addressed by [code here](https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L2496-L2499).

a4fc00d: delete `Pilot.shouldExpediteAltitudeChange()` for which we don't even use the returned `[success, readback]`. Fold trivial functionality into `maintainAltitude()` which is the only place it is called. Updated tests accordingly.

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -345,11 +345,7 @@ export default class AircraftController {
         this.removeAircraftModelFromList(aircraftModel);
         this._removeTransponderCodeFromUse(aircraftModel);
         this.removeAllAircraftConflictsForAircraft(aircraftModel);
-
-        if (aircraftModel.isControllable) {
-            this.removeStripView(aircraftModel);
-        }
-
+        this.removeStripView(aircraftModel);
         this._scopeModel.radarTargetCollection.removeRadarTargetModelForAircraftModel(aircraftModel);
     }
 
@@ -383,10 +379,8 @@ export default class AircraftController {
             this._updateAircraftConflicts(aircraftModel, i);
             this._updateAircraftVisibility(aircraftModel);
 
-            // `#isFlightStripRemovable` will be true even when there is no corresponding
-            // `StripView` for and `aircraftModel`
-            if (aircraftModel.isFlightStripRemovable && this._stripViewController.hasStripViewModel(aircraftModel)) {
-                this._stripViewController.removeStripView(aircraftModel);
+            if (!aircraftModel.isControllable) {
+                this.removeStripView(aircraftModel);
             }
         }
     }
@@ -904,8 +898,6 @@ export default class AircraftController {
             );
 
             GameController.events_recordNew(GAME_EVENTS.ARRIVAL);
-            aircraftModel.setIsFlightStripRemovable();
-            aircraftModel.setIsRemovable();
             this.aircraft_remove(aircraftModel);
 
             return;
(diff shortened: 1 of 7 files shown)


n8rzz (approved):

cake-pie:
More changes:

4dea94a: additional changes related to flight strip removal and removability 

- both `StripViewController.hasStripViewModel()` and `StripViewController.removeStripView()` run a search in `StripViewController._collection`. Having to check for existence of a strip view before removing it thus ends up searching twice. Instead, just make it harmless to attempt removal of flight strip for an acft that doesn't have a corresponding flight strip.
- delete `AircraftModel.isFlightStripRemovable` which is completely redundant with `aircraftModel.isControllable`
- remove pointless calls to `aircraftModel.setIsRemovable()` in `AircraftController._updateAircraftVisibility()` when the acft is getting deleted anyway. (`aircraftModel.isRemovable` is actually a flag that is used to communicate to `_updateAircraftVisibility()` that the acft can be deleted once it is sufficiently far away.)

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -383,10 +379,8 @@ export default class AircraftController {
             this._updateAircraftConflicts(aircraftModel, i);
             this._updateAircraftVisibility(aircraftModel);
 
-            // `#isFlightStripRemovable` will be true even when there is no corresponding
-            // `StripView` for and `aircraftModel`
-            if (aircraftModel.isFlightStripRemovable && this._stripViewController.hasStripViewModel(aircraftModel)) {
-                this._stripViewController.removeStripView(aircraftModel);
+            if (!aircraftModel.isControllable) {
```
n8rzz, line 382: Good catch, I like this a ton! It was a poor decision to search twice like that 🤦.

n8rzz (approved):

erikquinn (approved):

PR #2051, opened 2022-12-22
Enhancement/2050 - auto tower controllability and behavior

### LINKED ISSUE #2049: Newly spawned departures fail to call up and intiate radio contact
Even with autotower off, new departures fail to call up upon spawning.

Expected behavior is to see this message:
https://github.com/openscope/openscope/blob/2860a23834ec11311cea47bac199031d0844955b/src/assets/scripts/client/aircraft/AircraftModel.js#L1278

### LINKED ISSUE #2050: Autotower issues omnibus
Various issues related to the experimental `Tower Control` (auto tower / auto departures) setting

### Controllability changes
Desired behavior with tower control set to the default `SYSTEM` setting: user is strictly TRACON only and has no control over acft that are on TWR frequency
- for departures, tower has control over aircraft that are at apron, taxi, waiting in queue and on takeoff. Aircraft climbing out will switch to departure roughly 0.5nm past rwy departure end
- arrivals switch to tower -- most convenient point for us to trigger this is when flight phase changes from `APPROACH` to `LANDING` (stabilized at final approach fix)

Expected user interface behaviors
- radar return / datablock styling to change accordingly and only display flight strips for acft we can control
- when switching frequency, arrivals should call up on the radio; departures should sign off
  - however, do not trigger radio calls when the auto tower setting is changed on the fly -- causes mass change in controllability of potentially many aircraft, but not are not actual handover/frequency change events

### Loss of functionality: reinstate scoring
Since auto tower is a setting that users can change, addition of automatic departures should not be to the detriment of users who wish to have it turned off. #1824 removed code related to scoring of takeoff. This should be reinstated and applied when the setting is `USER`

### "Decision" by acft to auto takeoff should not be baked in upon spawning
The auto tower setting can be changed while the sim is running; behavior should be adjusted on the fly to match. 

Presently any departures that were spawned while tower control is set to `SYSTEM` are predetermined to take off on their own accord due to `shouldTakeOffWhenRunwayIsClear` being set. They will do so even if the setting is changed to `USER` while they are stuck in queue. Once the setting is changed from system to user the aircraft should cease taking off by themselves.

cake-pie:
Resolves #2050

- When tower control set to system, user (TRACON) does not have control over acft on tower frequency
- Reinstate scoring of takeoffs for when tower control set to user
- Automatic departures: aircraft no longer predestined to automatically take off upon spawning, will instead make the decision when they are at front of queue and runway is clear and tower control is set to system at that point in time.

Resolves #2049

- Fix missing "ready to taxi" call up from spawning departures (when tower control set to user)

### DIFF
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -184,6 +184,7 @@ export default class AircraftController {
      */
     _setupHandlers() {
         this._onRemoveAircraftHandler = this.aircraft_remove.bind(this);
+        this._onTowerControllerChangeHandler = this._onTowerControllerChange.bind(this);
 
         return this;
     }
@@ -201,6 +202,7 @@ export default class AircraftController {
         this._eventBus.on(EVENT.SCROLL_TO_AIRCRAFT, this._onScrollToAircraft);
         this._eventBus.on(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
         this._eventBus.on(EVENT.REMOVE_AIRCRAFT_CONFLICT, this.removeConflict);
+        this._eventBus.on(EVENT.TOWER_CONTROLLER_CHANGE, this._onTowerControllerChangeHandler);
 
         return this;
     }
@@ -218,6 +220,7 @@ export default class AircraftController {
         this._eventBus.off(EVENT.SCROLL_TO_AIRCRAFT, this._onScrollToAircraft);
         this._eventBus.off(EVENT.REMOVE_AIRCRAFT, this._onRemoveAircraftHandler);
         this._eventBus.off(EVENT.REMOVE_AIRCRAFT_CONFLICT, this.removeConflict);
+        this._eventBus.off(EVENT.TOWER_CONTROLLER_CHANGE, this._onTowerControllerChangeHandler);
 
         return this;
     }
@@ -915,6 +918,19 @@ export default class AircraftController {
         // Clean up the screen from aircraft that are too far
         if (!this.isAircraftVisible(aircraftModel, 2) && !aircraftModel.isControllable && aircraftModel.isRemovable) {
             this.aircraft_remove(aircraftModel);
+        }
+    }
+
+    /**
+     * Respond to user changing the tower control (autotower) settings
+     *
+     * @for AircraftController
+     * @private
+     */
+    _onTowerControllerChange() {
+        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
+        for (let i = 0; i < this.aircraft.list.length; i++) {
+            this.aircraft.list[i].onTowerControllerChange(isAutoTower);
         }
     }
 }
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -23,7 +23,7 @@ import { speech_say } from '../speech';
 import { generateTransponderCode, isDiscreteTransponderCode, isValidTransponderCode } from '../utilities/transponderUtilities';
 import { km } from '../utilities/unitConverters';
 import { isEmptyOrNotArray } from '../utilities/validatorUtilities';
-import { FLIGHT_CATEGORY, FLIGHT_PHASE } from '../constants/aircraftConstants';
+import { FLIGHT_CATEGORY } from '../constants/aircraftConstants';
 import { EVENT, AIRCRAFT_EVENT } from '../constants/eventNames';
 import { GAME_OPTION_NAMES } from '../constants/gameOptionConstants';
 import { INVALID_INDEX } from '../constants/globalConstants';
@@ -363,11 +363,13 @@ export default class AircraftController {
             return;
         }
 
+        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
+
         // TODO: this is getting better, but still needs more simplification
         for (let i = 0; i < this.aircraft.list.length; i++) {
             const aircraftModel = this.aircraft.list[i];
 
-            aircraftModel.update();
+            aircraftModel.update(isAutoTower);
             aircraftModel.updateWarning();
 
             // TODO: conflict checking eats up a lot of resources when there are more than
@@ -580,14 +582,8 @@ export default class AircraftController {
         }
 
         if (isDeparture && isAutoTower) {
-            // create the StripView immediately for departures; arrival strips are made when controllable
-            this._stripViewController.createStripView(aircraftModel);
             aircraftModel.pilot.clearedAsFiled();
-            aircraftModel.moveToRunway(aircraftModel.fms.departureRunwayModel);
-            aircraftModel.fms.departureRunwayModel.addAircraftToQueue(aircraftModel.id);
-            aircraftModel.setFlightPhase(FLIGHT_PHASE.WAITING);
-
-            aircraftModel.shouldTakeOffWhenRunwayIsClear = true;
+            aircraftModel.taxiToRunway(aircraftModel.fms.departureRunwayModel);
 
             this._runCommandOnPreSpawnAircraft(aircraftModel, runwayCommands, aircraftModel.fms.departureRunwayModel.name);
         }
--- a/src/assets/scripts/client/aircraft/AircraftController.js
+++ b/src/assets/scripts/client/aircraft/AircraftController.js
@@ -366,7 +366,7 @@ export default class AircraftController {
             return;
         }
 
-        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
+        const isAutoTower = GameController.isAutoTower();
 
         // TODO: this is getting better, but still needs more simplification
         for (let i = 0; i < this.aircraft.list.length; i++) {
@@ -573,7 +573,6 @@ export default class AircraftController {
         const aircraftModel = new AircraftModel(initializationProps);
         const isDeparture = initializationProps.category === 'departure';
         const isArrival = initializationProps.category === 'arrival';
-        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
         const runwayCommands = initializationProps.commands;
 
         // triggering event bus rather than calling locally because multiple classes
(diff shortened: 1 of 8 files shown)


src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -363,11 +366,13 @@ export default class AircraftController {
             return;
         }
 
+        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
```
n8rzz: How would you feel about moving this whole thing to the `GameController`? So usage here would be a few lines down `GameController.isAutoTower()`. Then we'd be able to use this anywhere 😄

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -921,4 +920,17 @@ export default class AircraftController {
...
     }
+
+    /**
+     * Respond to user changing the tower control (autotower) settings
+     *
+     * @for AircraftController
+     * @private
+     */
+    _onTowerControllerChange() {
+        const isAutoTower = GameController.getGameOption(GAME_OPTION_NAMES.TOWER_CONTROLLER) === 'SYSTEM';
```
n8rzz: with my suggestion above, this line goes away completely

src/assets/scripts/client/aircraft/AircraftController.js
```diff
@@ -921,4 +920,17 @@ export default class AircraftController {
...
+        for (let i = 0; i < this.aircraft.list.length; i++) {
+            this.aircraft.list[i].onTowerControllerChange(isAutoTower);
```
n8rzz, line 937: furthermore, we wouldn't even need to pass `isAutoTower` here. the `AircraftModel` itself could handle it all internally

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -88,6 +89,14 @@ const FLIGHT_RULES = {
     IFR: 'ifr'
 };
 
+const AUTOTOWER_FLIGHT_PHASES = [
```
n8rzz, line 102: Could you add a docblock here. specifically, why do we need this and how is it used? (I know these answers, but others coming through here may not.

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1312,14 +1320,17 @@ export default class AircraftModel {
...
      */
-    takeoff(runway) {
+    takeoff(runway, byAutoTower = false) {
         const cruiseSpeed = this.model.speed.cruise;
         const initialAltitude = this.fms.getInitialClimbClearance();
 
         this._prepareMcpForTakeoff(initialAltitude, runway.angle, cruiseSpeed);
         this.setFlightPhase(FLIGHT_PHASE.TAKEOFF);
-        EventBus.trigger(AIRCRAFT_EVENT.TAKEOFF, this, runway);
+        if (!byAutoTower) {
```
n8rzz, line 1342: missing empty line above `if`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1506,10 +1518,9 @@ export default class AircraftModel {
             }
 
             case FLIGHT_PHASE.WAITING:
-                const iAmTheNextDeparture = this.fms.departureRunwayModel.isAircraftNextInQueue(this.id);
-
-                if (this.shouldTakeOffWhenRunwayIsClear && iAmTheNextDeparture) {
-                    const lastDeparture = this.fms.departureRunwayModel.lastDepartedAircraftModel;
+                const runway = this.fms.departureRunwayModel;
+                if (isAutoTower && runway.isAircraftNextInQueue(this.id)) {
```
n8rzz, line 1534: missing empty line above `if`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -1506,10 +1518,9 @@ export default class AircraftModel {
...
+                    const lastDeparture = runway.lastDepartedAircraftModel;
                     const iAmTheFirstEverDeparture = lastDeparture === null;
```
n8rzz: I know this was here before, but could we rename `iAmTheFirstEverDeparture` to `isFirstEverDeparture`?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2735,42 +2761,68 @@ export default class AircraftModel {
...
      * @private
      */
-    _contactAircraftAfterControllabilityChange() {
-        // Crossing into the center
+    _updateAircraftControllability(isAutoTower) {
+        if (this.projected) {
+            return;
+        }
+
+        const newControllability = (isAutoTower && this._isUnderTowerControl()) ?
```
n8rzz: I find this name confusing. Taking a stab at it, does this mean an aircraft that was not controllable, now is (and visa versa)?

Could `isNewlyControllable` work here instead? Or am I completely misunderstanding what this is doing?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2735,42 +2761,68 @@ export default class AircraftModel {
...
+
         if (this.isControllable) {
             this.callUp();
 
             return;
         }
 
         this.setIsRemovable();
-        EventBus.trigger(AIRCRAFT_EVENT.AIRSPACE_EXIT, this);
+        if (this.flightPhase !== FLIGHT_PHASE.LANDING) {
```
n8rzz, line 2804: missing empty line above `if`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2735,42 +2761,68 @@ export default class AircraftModel {
...
      * @private
      */
-    _updateAircraftControllability() {
-        if (this.projected) {
-            return;
+    _isUnderTowerControl() {
+        if (_includes(AUTOTOWER_FLIGHT_PHASES, this.flightPhase)) {
+            return true;
+        }
+        if (this.category !== FLIGHT_CATEGORY.DEPARTURE || this.flightPhase !== FLIGHT_PHASE.CLIMB) {
```
n8rzz, line 2819: missing empty line above `if`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2735,42 +2761,68 @@ export default class AircraftModel {
...
+            return false;
         }
 
-        const isInsideAirspace = this.isInsideAirspace(AirportController.airport_get());
+        // FAA JO 7110.65, Para. 3-9-3-b (civil acft)
+        const runway = this.fms.departureRunwayModel;
+        return this.positionModel.distanceToPosition(runway.positionModel) < nm(runway.length) + 0.5;
```
n8rzz, line 2826: missing empty line above `return`

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -2735,42 +2761,68 @@ export default class AircraftModel {
...
+    /**
+     * Respond to user changing the tower control (autotower) settings
+     * by adjusting controllability as needed *without* making radio calls
+     *
+     * @for AircraftModel
+     * @method onTowerControllerChange
+     * @param isAutoTower {boolean}
+     */
+    onTowerControllerChange(isAutoTower) {
+        if (this._isUnderTowerControl()) {
```
n8rzz: This is a nit pick, but we/I have tried really hard to enforce exclusive if statements. Meaning, if a method or function doesn't have what we need to do the work, `return`. A function/method should end like the name says.

So in this case, we would change the guts of this method to:
```js
if (!this._isUnderTowerControl()) {
  return;
}

this.isControllable = !isAutoTower;
```

This gives us clear visibility into both logical branches and also makes the method read like the name

n8rzz (changes requested):
Just a few style changes and things. Thank you so much for including a like to only the changed files, that made this review way easier.

src/assets/scripts/client/aircraft/AircraftController.js
cake-pie: Implemented `GameController.isAutoTower()`. However, from an optimization standpoint it still seems preferable to fetch that value once here and subsequently pass it to each aircraft instance as we loop through them below, than have multiple calls individually.

src/assets/scripts/client/aircraft/AircraftController.js
cake-pie: Again, fetch once and reuse...

src/assets/scripts/client/aircraft/AircraftController.js
cake-pie: ... we don't want each aircraft instance making the same redundant `isAutoTower()` call

src/assets/scripts/client/aircraft/AircraftModel.js
cake-pie: done

src/assets/scripts/client/aircraft/AircraftModel.js
cake-pie: We want to detect a change from controllable -> not controllable or vice versa.

`this.isControllable` is the "old" cached value from before. `newControllability` is our freshly calculated value as at the current simulation tick. This variable used to be called `isInsideAirspace` (see [previous version of the whole function](https://github.com/openscope/openscope/blob/94b8cee2bca8470186adf4d5a5ec25438c6e9c10/src/assets/scripts/client/aircraft/AircraftModel.js#L2756-L2774)) but that name is no longer accurate as the test has become a bit more involved. A couple of lines down we compare the two and if needed, update `this.isControllable` to the new value and react to the change by  doing other stuff.

`isNewlyControllable` seems to mean something different, namely
```
// was not controllable before; has just become controllable
const isNewlyControllable = !this.isControllable && newControllability;

// compare with
const isNewlyUncontrollable = this.isControllable && !newControllability;
```

src/assets/scripts/client/aircraft/AircraftModel.js
cake-pie: done

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: 🤦 Holy crap you're right! Ignore this entire line of feedback from me, not sure what I was thinking before. What you did originally is way better.

src/assets/scripts/client/aircraft/AircraftController.js
n8rzz: See below, you're absolutely correct. Disregard what I said (sorry for the back and forth), your original implementation is way better.

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: I see what you're saying and that helps me understand this much better. Yeah, I thought `newControllability` was a flag of some sort, but instead it's the next value for `isControllable`. I think the word `new` through me off here. Would the name `nextControllability` work here? That name helps suggest that this value is going to be used to replace something later on.

cake-pie: `nextControllability` makes sense, done

n8rzz (approved):

PR #2076, opened 2023-01-27
bugfix/2075 - Fix DAB southern arrivals and update airport guide

svalencia014:
Resolves #2075.

- Update KDAB
- improve southern spawning and correct airport guide

The purpose of this pull request is to fix southern arrivals which previously spawned at 11,000, making the controller provide vectors for descent.
This changes those aircraft to spawn at 6,000 for the jets, and 5,000 for the props (for separation), per the vZJX DAB SOP ("F11 TRACON will assign 
arrival aircraft to cross the TRACON border at no higher than 6,000").
higher than 6,000’."

PR #2086, opened 2023-03-01
Bugfix/2075

### LINKED ISSUE #2075: Fix South Arrivals at KDAB
Airport: KDAB

Per the [vZJX SOP](https://zjxartcc.org/media/doc/ZJX-5.D_Jacksonville_ARTCC_DAB_ATCT_TRACON_SOP.pdf), aircraft arriving from the south (`SMYRA` and `MIKBE`) should be assigned no higher than 6,000 by F11 TRACON to allow for the South Arrival flow described on page 25. This issue aims to fix that to be IAW SOP to make flows work better.

svalencia014:
Resolves #2075 

The purpose of this pull request is to fix southern arrivals which previously spawned at 11,000, making the controller provide vectors for the descent.
This changes those aircraft to spawn at 6,000 for the jets, and 5,000 for the props (for separation), per the vZJX DAB SOP ("F11 TRACON will assign
arrival aircraft to cross the TRACON border at no higher than 6,000").
higher than 6,000’."

(closed #2076 due to branch being renamed to follow proper naming)


documentation/airport-guides/kdab.md
```diff
@@ -1,13 +1,30 @@
...
+Daytona Beach International Airport (IATA: DAB, ICAO: KDAB, FAA LID: DAB) is a county-owned airport located three miles (5 km) southwest of Daytona Beach, next to Daytona International Speedway, in Volusia County, Florida, United States. The airport has 3 runways, a six-gate domestic terminal, and an international terminal. Daytona Beach is the headquarters of Embry-Riddle Aeronautical University. ([Wikipedia](https://en.wikipedia.org/wiki/Daytona_Beach_International_Airport))
 
 ## Airspace
-The DAB airspace covers multiple smaller airports in the region from the surface to 11,000'. It is bordered to the north and west by Jacksonville Center and to the south by the Central Florida TRACON (F11) and Miami Center.
+The DAB airspace covers multiple smaller airports in the region from the surface to 11,000'. It is bordered to the north by Jacksonville TRACON (JAX), north and west by Jacksonville Center, and to the south by the Central Florida TRACON (F11) and Miami Center.
```
Fechulo, line 7: > It is bordered to the north by Jacksonville TRACON (JAX), north and west by Jacksonville Center,

Doesn't sound quite right

documentation/airport-guides/kdab.md
```diff
@@ -1,13 +1,30 @@
...
 
 ## Procedures
 DAB has one conventional SID and one RNAV STAR. Additionally, the airspace has a significant number of Victor airways and RNAV T-routes. This airport features a high number of overflight aircraft. These aircraft have a yellow flight strip and must be guided to their destination along their planned route of flight.
 
+### Arrivals
+
+vZJX Recommends a flow for aircraft on the `TTHOR3`, crossing `TTHOR` at 12,000, `BIKEZ` at 8,000. All aircraft arriving from the north get direct `OMN` at 6,000. From there, they should get heading 260, and a descent down to 4,000. Then, turn them South Bound (I like to wait until they pass `TIGAE`) which can be anything from a `150` to a `180` heading, and a descent down to 3,000. The final turn is a `090` heading and clearing them for the ILS (`i 7L`).
```
Fechulo: This is all very useful information, but the phrasing is a little too casual. Ideally, the first person pronoun should be avoided. 

For example:
> There is no "correct" way to handle departures, so I like do to it this way.

Could be replaced with something along the lines of:
> Departure traffic can be handled in many different ways. The following is one of the suggested workflows, which may need to be adapted in certain traffic situations.

There is also a _lot_ of information here which is hard to parse as a wall of text. Perhaps a diagram of some sort might be better to convey the suggested flows?

Sample commands should be removed from here as well. If commands were to be changed, only the command reference should need to be updated.

assets/airports/kdab.json
```diff
@@ -561,7 +561,7 @@
             "destination": "kdab",
             "category": "arrival",
             "route": "FOBIN.Q93.MALET..OMN",
-            "altitude": 23000,
+            "altitude": 6000,
```
Fechulo: I think it'd be better to make use of the `commands` key to ensure aircraft enter the TRACON at 6000ft, rather than forcing them to spawn at 6000ft. (See https://github.com/openscope/openscope/blob/develop/documentation/spawnPatternReadme.md)

It may still be necessary to decrease the spawn altitude, but aircraft should ideally spawn at a higher level and then descend to the desired level

Fechulo (changes requested):

documentation/airport-guides/kdab.md
svalencia014: What would you recommend I say instead? The Satellite, East, and West Sectors own past the northern border, with a slight bit of west and mainly the Vitts sector owning past the western border 
Source: [vZJX JAX SOP](https://www.idszjxartcc.org/index.php?title=JAX_SOP#TRACON_Information) | [vZJX FAP](https://zjxartcc.org/media/doc/ZJX-1000_FAP.docx_3.pdf#page=8)

assets/airports/kdab.json
svalencia014: Testing my change, I still have a plane spawning at 11,000 when I set altitude back to 23,000 and use the `cross MALET A060` command in the file.

Fechulo: Try setting the spawn altitude to something like 18,000. If that still doesn't work then I'm happy setting the spawn height to 6k.

documentation/airport-guides/kdab.md
Fechulo: Ah, I misunderstood the original sentence. Ignore!

Fechulo (commented):
This reads much better now, good work!

assets/airports/kdab.json
svalencia014: The first plane that spawns on the `FOBIN.Q93.MALET..OMN` route still spawns at 11,000. All planes past the initial one on that route spawn normally and descend as instructed.

Fechulo: Alright, must be something to do with the first spawn logic. We can keep it at 6000 then

svalencia014:
Closing this PR as it's been open for two years without a merge

PR #2114, opened 2023-10-15
Sanitize airport definition files

inferiorhumanorgans:
As part of another project I wrote a parser/linter for the airport definition files and ran it on everything in the repository.  The resulting changes are listed in the commit message, roughly:

* Fixing invalid routes (e.g. speed restriction of `S21O` instead of `S210`)
* Filling in some missing but easy to find data (e.g. field elevation)
* Ensuring numeric values are numbers and not strings (unless units need to be specified)
* Cleaning up the coordinates, specifically if minutes are specified with a decimal component, don't allow the seconds to be specified (`dNN.NNm0` was used haphazardly) – this makes parsing a bit simpler and more predictable
* Removed the 'empty' maps where the shape wouldn't have parsed in the first place
* Reshaped old maps into the format the current code uses (array vs dict)
* Rewrote the elevation parser and enforced a space between value and units
* Removed name_offset as it's not referenced in code anywhere
* Updated the documentation

Known issues:

* Didn't update as I couldn't find info: SBGR and OSDI
* Map line segments are invalid: EGNM

One potentially intrusive change I've made is to change missing values to `null` instead of empty strings for arrival and departure routes.  This is accompanied by checking for null in a few places with the `?` operator.  This makes parsing easier and semantically seems to make more sense to me.

All the tests appear to pass.


src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -663,8 +663,8 @@ export default class AircraftModel {
         this.heading = data.heading;
         this.altitude = data.altitude;
         this.speed = data.speed;
-        this.origin = _get(data, 'origin', this.origin);
-        this.destination = _get(data, 'destination', this.destination);
+        this.origin = data.origin?.toUpperCase(),
```
n8rzz:
```suggestion
        this.origin = data.origin?.toUpperCase();
```

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -663,8 +663,8 @@ export default class AircraftModel {
...
+        this.destination = data.destination?.toUpperCase(),
```
n8rzz:
```suggestion
        this.destination = data.destination?.toUpperCase();
```

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: Remind me, do we need a plugin or anything for babel to make conditional chaining work or is this just native now?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -707,8 +707,8 @@ export default class AircraftModel {
             icaoWithWeightClass: this.model.icaoWithWeightClass,
             assignedAltitude,
             flightPlanAltitude,
-            arrivalAirportId: this.destination.toUpperCase(),
-            departureAirportId: this.origin.toUpperCase(),
+            arrivalAirportId: this.destination?.toUpperCase(),
```
n8rzz, line 710: How do things behave when `destination` is `null`?

src/assets/scripts/client/aircraft/AircraftModel.js
```diff
@@ -707,8 +707,8 @@ export default class AircraftModel {
...
+            departureAirportId: this.origin?.toUpperCase(),
```
n8rzz, line 711: How do things behave when `origin` is `null`?

src/assets/scripts/client/aircraft/FlightManagementSystem/Fms.js
```diff
@@ -308,8 +308,8 @@ export default class Fms {
      * @private
      */
     _initializeAirportsAndRunways(origin, destination) {
-        const originLowerCase = origin.toLowerCase();
-        const destinationLowerCase = destination.toLowerCase();
+        const originLowerCase = origin?.toLowerCase();
```
n8rzz, line 311: same question for these two: how does this behave when the are `null`?

n8rzz (approved):
I have questions about how the app will behave with some of these properties moving to `null` instead of an empty string. I _think_ it will be okay, but we tend to lean on strict equals a lot and `null` is not `''`. 

Any chance you could add a few tests to support some of the spots I called out, or just verify those spots are already covered by tests, then I think we'll be good to go!

Thanks for the effort and your contribution on this! I'd be interested to see the external tool you mention. At one time, I attempted an [airport validator](https://github.com/openscope/validator) using a js library called `tcomb`. It kinda worked but I wasn't very happy with it. I'd love to see what someone else put together!

src/assets/scripts/client/aircraft/AircraftModel.js
inferiorhumanorgans: The `?` operator only calls the following function if the value is not null or undefined, so the code is equivalent to e.g.:

`this.origin ? this.origin.toUpperCase() : null`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

inferiorhumanorgans:
https://git.sr.ht/~az1/openscope-data

So it's mostly intended as a parser (and it's still under development), but adding proper validation would be pretty easy.

src/assets/scripts/client/aircraft/AircraftModel.js
n8rzz: Yes, I know. I'm asking because many places in the app use strict `===` equals and this property was formerly a string. An _"invalid"_ or _"empty"_ state before would have been `''`. Now it's null. Using strict equals, what was once true is no longer. 

All I'm asking is: when `origin` or `destination` are [now] `null`, does everything behave the same way? And, if so, can we verify that with a test or two?

n8rzz:
> https://git.sr.ht/~az1/openscope-data
> 
> So it's mostly intended as a parser (and it's still under development), but adding proper validation would be pretty easy.

That's pretty freaking cool! Nice work!!

src/assets/scripts/client/aircraft/AircraftModel.js
inferiorhumanorgans: Sure, I'll take a look over the weekend.

src/assets/scripts/client/aircraft/AircraftModel.js
inferiorhumanorgans: The oldest browser I've laying around is Safari 14 and the `?` operator seems to work as expected (but I can't do much more than that as Openscope itself doesn't work in Safari 14).  Seems to work in Firefox LTS and some ancient version of Chrome as well.

src/assets/scripts/client/aircraft/AircraftModel.js
inferiorhumanorgans: I've changed the only place that was comparing the origin and/or destination fields to an empty string in b4fb948, and corrected the defaults to `null`.  I'll take a look at the test coverage later.

inferiorhumanorgans:
> > https://git.sr.ht/~az1/openscope-data
> > So it's mostly intended as a parser (and it's still under development), but adding proper validation would be pretty easy.
> 
> That's pretty freaking cool! Nice work!!

Welp it's intended first and foremost as a parser so as a linter it's not great.  Namely it bails on the first error.  Certainly it could be massaged to look like a proper linter, but that's not something I'm terribly focused on right now.

Meanwhile I've snuck in a few more changes:

* nuked the 'empty' maps where the shape wouldn't have parsed in the first place
* reshaped old maps into the format the current code uses
* rewrote the elevation parser and enforced a space between value and units
* removed name_offset as it's not referenced in code anywhere (I've not looked at the label drawing code to see if it should be smarter than it is)
* updated the documentation

src/assets/scripts/client/aircraft/AircraftModel.js
inferiorhumanorgans: In terms of test coverage creating objects with default (empty string, now `null`) origin and/or destination is covered both by some of the mocks and some explicit checks for behavior e.g. "_initializeAirportsAndRunways() does not make calls to initialize departure airport+runway when origin ICAO is null".  I updated this in 23dc216 and at first blush I think that should manage expectations within the code appropriately.

n8rzz: Awesome! Thanks for looking into that, I appreciate it.

inferiorhumanorgans: I actually caught one more instance of empty strings in the spawn patterns that I missed.  I'll push that tomorrow.

PR #2121, opened 2024-02-11
Bugfix/2108

### LINKED ISSUE #2108: Incorrect fix location on SID at RJOA
Airport: RJOA

The fix LEMON (MARCO SID) is incorrectly implemented to an identically named fix in South Korea.
Correct coordinates: N34.224683 E132.463589

image

niin54:
Resolves #2108 

The purpose of this pull request is to correct the waypoint for LEMON near RJOA
May be changed in future if more accurate data for the waypoint is received (see original issue)


niin54:
After testing, it looks like it works just fine

Fechulo (approved):

a6rst:
True coordinates are at **N34-13-28.86 E132-27-48.92** ([AIP Japan](https://aisjapan.mlit.go.jp/), see ENR 4.3)

image

